PR #6, opened 2021-05-31
Copy of probabilisticsamplerprocessor with Sumo extensions

pmm-sumo:
This PR brings probabilisticsamplerprocessor with custom Sumo extension which injects the sampling rate into `sampling.probability` attribute


pmatyjasek-sumo (approved):

pmm-sumo:
TIL that builder won't allow overriding existing components (at least not in this way)

pmalek-sumo:
> TIL that builder won't allow overriding existing components (at least not in this way)

😢  I believe the problem is not in the component already existing in the conf but rather it stems from the fact that the `probabilisticsamplerprocessor` is not a module itself hence you cannot `replace` it.

PR #12, opened 2021-06-11
feat(ci): add dev container builds to Dev Builds job on CI

pmalek-sumo:
This PR adds pushing to [Open Source dev ECR](https://gallery.ecr.aws/a4t4y2n3/sumologic-otel-collector-dev) on commits in main and release branches.

Example what it looks like after the push:

image

Example of successful pipepile ( back then launched on a PR branch not on main):

https://github.com/SumoLogic/sumologic-otel-collector/actions/runs/929478964

### DIFF
--- a/Dockerfile_dev
+++ b/Dockerfile_dev
@@ -1,9 +1,3 @@
-FROM golang:1.16.4 as builder
-ADD . /src
-WORKDIR /src/otelcolbuilder/
-RUN make install
-RUN make build
-
 FROM alpine:3.13 as certs
 RUN apk --update add ca-certificates
 
@@ -14,7 +8,7 @@ ARG USER_UID=10001
 USER ${USER_UID}
 
 COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
-COPY --from=builder /src/otelcolbuilder/cmd/otelcol-sumo /otelcol-sumo
+COPY otelcol-sumo /otelcol-sumo
 EXPOSE 55680 55679
 ENTRYPOINT ["/otelcol-sumo"]
 CMD ["--config", "/etc/otel/config.yaml"]
--- a/Dockerfile_dev
+++ b/Dockerfile_dev
@@ -1,12 +1,13 @@
 FROM golang:1.16.4 as builder
 ADD . /src
 WORKDIR /src/otelcolbuilder/
-RUN make install && make build
+RUN make install
+RUN make build
 
 FROM alpine:3.13 as certs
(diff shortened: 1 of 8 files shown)


ci/push_docker_multiplatform_manifest.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
+
+set -eo pipefail
+
+if ! docker buildx ls | grep -q arm ; then
+    echo "Your Buildx seems to lack ARM architecture support"
+    echo
+    docker buildx ls
+    exit 1
+fi
```
sumo-drosiek, line 10: does it mean I cannot build other architecture unless I have arm support?

Dockerfile_dev
```diff
@@ -0,0 +1,14 @@
...
+RUN apk --update add ca-certificates
+
+FROM golang:1.16.4
+ARG BUILD_TAG=latest
+ENV TAG $BUILD_TAG
+ARG USER_UID=10001
+USER ${USER_UID}
+
+COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY otelcol-sumo /otelcol-sumo
```
astencel-sumo, line 11: Sorry if this is a stupid question but where did this `otelcol-sumo` file come from?

astencel-sumo: Oh I think I get it, it's being built in CI (GH Actions) in the `build` job and then copied in the `build-container-images` job.

pmalek-sumo: Yes. The reason for this is that I tried to do it the same way as for fluentd (run all builds and wait for build pids) but the CI would stall for an hour without completing: https://github.com/SumoLogic/sumologic-otel-collector/actions/runs/928636321

Then I tried to split this and use the github provided builders for arm (using strategy matrix for that) but then amd 64 build would build in handful of minues (👍 ) but arm builds would stall and not complete even within 30-40m (sometimes they would complete but we don't want to wait 30m for a build 😬 ) https://github.com/SumoLogic/sumologic-otel-collector/actions/runs/929008398

ci/push_docker_multiplatform_manifest.sh
pmalek-sumo: I could improve this checking the params to this scripts against output of `docker buildx ls` 👍

astencel-sumo (approved):

Dockerfile_dev
astencel-sumo: Ohhh this sucks. Thanks for explaining.

PR #14, opened 2021-06-14
feat(sumologicexporter): Translate metadata

astencel-sumo:


pkg/exporter/sumologicexporter/transform_metadata_test.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.opentelemetry.io/collector/consumer/pdata"
+)
+
+func TestTransformMetadata(t *testing.T) {
+	metadata := pdata.NewAttributeMap()
+	metadata.InsertString("host.name", "testing-host")
```
pmalek-sumo, line 12: more tests :)

Try it with more attributes in the attr map

pkg/exporter/sumologicexporter/README.md
```diff
@@ -8,6 +8,10 @@ The following configuration options are supported:
 - `compress_encoding` (optional): Compression encoding format, either empty string (`""`), `gzip` or `deflate` (default `gzip`).
 Empty string means no compression
 - `max_request_body_size` (optional): Max HTTP request body size in bytes before compression (if applied). By default `1_048_576` (1MB) is used.
+- `transform_metadata` (optional): Specifies whether metadata attributes should be transformed
```
astencel-sumo, line 11: Now that I read it again, should I add a `Metadata Transformation` subsection to the README that will list *all* the metadata that is being transformed? Effectively copying the `metadataTransformation` variable from [transform_metadata.go:5](https://github.com/SumoLogic/sumologic-otel-collector/pull/14/files#diff-09f7a032d741100d7ab9a0d2ed0a5e226b7c36dc411bb01de71cac48ba1d1eefR5)

pkg/exporter/sumologicexporter/factory_test.go
```diff
@@ -48,6 +48,7 @@ func TestCreateDefaultConfig(t *testing.T) {
 		SourceHost:         "",
 		Client:             "otelcol",
 		GraphiteTemplate:   "%{_metric_}",
+		TransformMetadata:  true,
```
pmatyjasek-sumo, line 51: why not `DefaultTransformMetadata` instead of `true`? ;)

pkg/exporter/sumologicexporter/config.go
```diff
@@ -135,4 +140,6 @@ const (
 	DefaultClient string = "otelcol"
 	// DefaultGraphiteTemplate defines default template for Graphite
 	DefaultGraphiteTemplate string = "%{_metric_}"
+	// DefaultTransformMetadata defines default TransformMetadata
```
pmatyjasek-sumo, line 143:
```suggestion
	// By default TransformMetadata is enabled
```

pkg/exporter/sumologicexporter/transform_metadata.go
```diff
@@ -0,0 +1,37 @@
+package sumologicexporter
```
pmatyjasek-sumo, line 1: missing header

pkg/exporter/sumologicexporter/transform_metadata_test.go
```diff
@@ -0,0 +1,89 @@
+package sumologicexporter
```
pmatyjasek-sumo, line 1: missing header

pkg/exporter/sumologicexporter/README.md
pmatyjasek-sumo: I think it would be good to add a table with those transformations in README

pkg/exporter/sumologicexporter/config.go
astencel-sumo: The comment I added matches all the other comments in this struct if you look at the other fields like `Default Compress`, `DefaultLogFormat` etc.

pkg/exporter/sumologicexporter/factory_test.go
astencel-sumo: The other settings are also set not using the `Default` fields, like `CompressEncoding`, `LogFormat`, `MetricFormat` etc.

I believe the purpose of this test is to verify that the default config sets the values we expect. If the default value is changed, this test should be updated.

pkg/exporter/sumologicexporter/transform_metadata.go
astencel-sumo: That's correct. I want to ask here:
1) Is the part `// Copyright 2021, OpenTelemetry Authors` correct? Are we _OpenTelemetry Authors_ here in this repo?
2) Do we need all those headers in the first place? I haven't seen headers in other repositories of ours.

astencel-sumo: From the discussion in the sync:
1. We should be using `Sumo Logic, Inc.` as the author for code originally written in Sumo Logic and specific to Sumo Logic projects (as opposed to upstream projects like [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector), [OpenTelemetry Collector Contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/) etc.)
2. Yes we need the headers, we should probably stick to what the License says in the appendix: https://www.apache.org/licenses/LICENSE-2.0.html#apply

astencel-sumo: Added header.

pkg/exporter/sumologicexporter/transform_metadata_test.go
astencel-sumo: Added header.

pkg/exporter/sumologicexporter/README.md
astencel-sumo: Added section on metadata transformation to README, with a list of all transformed attribute keys.

pkg/exporter/sumologicexporter/translate_metadata.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// translateMetadata renames metadata keys according to metadataTranslations.
+func translateMetadata(attributes pdata.AttributeMap) {
+	attributes.Range(func(otKey string, value pdata.AttributeValue) bool {
+		sumoKey := metadataTranslations[otKey]
+		if sumoKey != "" {
+			attributes.Upsert(sumoKey, value)
+			attributes.Delete(otKey)
+		}
```
pmatyjasek-sumo, line 54: ``
```suggestion
		if sumoKey, ok := metadataTranslations[otKey]; ok {
			attributes.Upsert(sumoKey, value)
			attributes.Delete(otKey)
		}
```

astencel-sumo: Thanks, done

pmatyjasek-sumo (approved):

PR #15, opened 2021-06-14
refactor(ci): check all prs

pmalek-sumo:


sumo-drosiek (approved):

.github/workflows/pull_requests.yml
```diff
@@ -3,8 +3,7 @@ name: PRs checks
 on:
   pull_request:
     branches:
-      - main
-      - 'release-v[0-9]+.[0-9]+'
+      - '**'
```
sumo-drosiek, line 6: why two asterisks?

pmalek-sumo: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

> `*`: Matches zero or more characters, but does not match the / character. For example, `Octo*` matches `Octocat`.
> `**`: Matches zero or more of any character.

PR #16, opened 2021-06-14
feat(sumologcexporter): add otlp logs support

sumo-drosiek:
Add support for OTLP format for logs. This implementation doesn't respect HTTP restriction like message size. This is done such way due to simplicity and urgency of the solution

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp-request
https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto

### DIFF
--- a/pkg/exporter/sumologicexporter/sender.go
+++ b/pkg/exporter/sumologicexporter/sender.go
@@ -73,6 +73,7 @@ const (
 	contentTypePrometheus string = "application/vnd.sumologic.prometheus"
 	contentTypeCarbon2    string = "application/vnd.sumologic.carbon2"
 	contentTypeGraphite   string = "application/vnd.sumologic.graphite"
+	contentTypeOTLP       string = "application/x-protobuf"
 
 	contentEncodingGzip    string = "gzip"
 	contentEncodingDeflate string = "deflate"
@@ -179,6 +180,12 @@ func (s *sender) logToJSON(record pdata.LogRecord) (string, error) {
 // to configured LogFormat and as the result of execution
 // returns array of records which has not been sent correctly and error
 func (s *sender) sendLogs(ctx context.Context, flds fields) ([]pdata.LogRecord, error) {
+
+	// Follow different execution path for OTLP format
+	if s.config.LogFormat == OTLPLogFormat {
+		return s.sendOTLPLogs(ctx, flds)
+	}
+
 	var (
 		body           strings.Builder
 		errs           []error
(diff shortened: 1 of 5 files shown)


pkg/exporter/sumologicexporter/sender.go
```diff
@@ -241,6 +247,34 @@ func (s *sender) sendLogs(ctx context.Context, flds fields) ([]pdata.LogRecord,
 	return droppedRecords, nil
 }
 
+// sendLogs sends log records from the logBuffer in OTLP format
+// and as the result of execution
+// returns array of records which has not been sent correctly and error
```
pmalek-sumo, line 252:
```suggestion
// sendLogs sends log records from the logBuffer in OTLP format and as a result 
// it returns an array of records which has not been sent correctly and an error.
```

pkg/exporter/sumologicexporter/sender.go
```diff
@@ -241,6 +247,34 @@ func (s *sender) sendLogs(ctx context.Context, flds fields) ([]pdata.LogRecord,
...
+// TODO: add support for HTTP limits
+func (s *sender) sendOTLPLogs(ctx context.Context, flds fields) ([]pdata.LogRecord, error) {
+	ld := pdata.NewLogs()
+	rls := ld.ResourceLogs()
+	rls.Resize(1)
+	rl := rls.At(0)
+	ills := rl.InstrumentationLibraryLogs()
+	ills.Resize(1)
+	ill := ills.At(0)
+	logs := ill.Logs()
```
pmalek-sumo, line 262: preallocate?

pkg/exporter/sumologicexporter/sender.go
```diff
@@ -241,6 +247,34 @@ func (s *sender) sendLogs(ctx context.Context, flds fields) ([]pdata.LogRecord,
...
+
+	body, err := ld.ToOtlpProtoBytes()
+	if err != nil {
+		return s.logBuffer, err
+	}
+
+	if err := s.send(ctx, LogsPipeline, bytes.NewReader(body), flds); err != nil {
+		return s.logBuffer, err
+	}
+	return nil, err
```
pmalek-sumo, line 275:
```suggestion
	if err := s.send(ctx, LogsPipeline, bytes.NewReader(body), flds); err != nil {
		return s.logBuffer, err
	}
	return nil, nil
```

pkg/exporter/sumologicexporter/sender.go
sumo-drosiek: Added. Also I used `CopyTo` method as its recommended to not reuse LogRecord in multiple structures

pkg/exporter/sumologicexporter/sender.go
```diff
@@ -186,6 +187,11 @@ func (s *sender) sendLogs(ctx context.Context, flds fields) ([]pdata.LogRecord,
 		currentRecords []pdata.LogRecord
 	)
 
+	// Follow different execution path for OTLP format
+	if s.config.LogFormat == OTLPLogFormat {
+		return s.sendOTLPLogs(ctx, flds)
+	}
```
pmalek-sumo, line 193: nit: this can be moved above the vars declaration block as it doesn't it them

pmalek-sumo (approved):

kkujawa-sumo (approved):

PR #18, opened 2021-06-15
test(sumologicexporter): Add tests for `metadata_attributes`

astencel-sumo:
These tests specify how the `metadata_attributes` configuration works for different formats of logs and metrics.

### DIFF
--- a/pkg/exporter/sumologicexporter/sender_test.go
+++ b/pkg/exporter/sumologicexporter/sender_test.go
@@ -26,7 +26,6 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"go.opentelemetry.io/collector/config/confighttp"
 	"go.opentelemetry.io/collector/consumer/pdata"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/exporter/sumologicexporter/sender_test.go
```diff
@@ -60,6 +60,7 @@ func prepareSenderTest(t *testing.T, cb []func(w http.ResponseWriter, req *http.
 		MetricFormat:       "carbon2",
 		Client:             "otelcol",
 		MaxRequestBodySize: 20_971_520,
+		GraphiteTemplate:   DefaultGraphiteTemplate,
```
pmalek-sumo, line 63: Why is this needed here? Maybe we should use the default config via `createDefaultConfig` and then change it here?

sumo-drosiek:
Is adding new test file with unit tests which tests another file a good approach? I mean, now we will have two unit test files for exporter: `exporter_metadata_test.go` and `exporter_test.go`

astencel-sumo:
> Is adding new test file with unit tests which tests another file a good approach? I mean, now we will have two unit test files for exporter: `exporter_metadata_test.go` and `exporter_test.go`

Right, thanks for bringing this to attention. I had a feeling it's hard to find anything in such a single big file with unit tests and thought maybe splitting into a couple separate files could help.

pkg/exporter/sumologicexporter/sender_test.go
astencel-sumo: I agree it would be more readable to use the default config and override anything needed. I'll refactor that.

astencel-sumo: Done @pmalek-sumo please take a look

astencel-sumo:
> > Is adding new test file with unit tests which tests another file a good approach? I mean, now we will have two unit test files for exporter: `exporter_metadata_test.go` and `exporter_test.go`
> 
> Right, thanks for bringing this to attention. I had a feeling it's hard to find anything in such a single big file with unit tests and thought maybe splitting into a couple separate files could help.

@sumo-drosiek I have moved the new tests into `exporter_test.go` to keep all tests in single file. Plese take a look.

pmalek-sumo (approved):

PR #27, opened 2021-06-17
refactor: sumologicextension

pmalek-sumo:


pkg/extension/sumologicextension/extension.go
```diff
@@ -186,82 +180,33 @@ func (se *SumologicExtension) storeCollectorCredentials() error {
...
-	nonceSize := gcm.NonceSize()
-	if nonceSize > len(data) {
-		return nil, fmt.Errorf("unable to decrypt credentials")
-	}
-	nonce, ciphertext := data[:nonceSize], data[nonceSize:]
-	plaintext, err := gcm.Open(nil, nonce, ciphertext, nil)
-	if err != nil {
-		return nil, err
+func (se *SumologicExtension) register(ctx context.Context) error {
+	type OpenRegisterRequestPayload struct {
```
pmatyjasek-sumo, line 201: Why this is moved here? I would prefer to keep structs outside functions

pkg/extension/sumologicextension/extension.go
```diff
@@ -137,43 +128,46 @@ func (se *SumologicExtension) checkCollectorCredentials() bool {
 }
 
 // getCollectorCredentials retrieves and decrypts collector credentials using
-// collector name as passphrase.
-func (se *SumologicExtension) getCollectorCredentials() error {
-	filenameHash, err := createHash(se.conf.CollectorName)
+// hashed collector name as passphrase.
+func (se *SumologicExtension) getCollectorCredentials() (OpenRegisterResponsePayload, error) {
```
pmatyjasek-sumo, line 132: What is the value added of returning `OpenRegisterResponsePayload` and assign it to `se` struct outside the function instead of make it in place in this method?

pmalek-sumo: This IMHO makes it obvious what the function does: it's called `get...` so it gets it hence the return value. Setting this field in the function is a non obvious side effect.

pkg/extension/sumologicextension/extension.go
pmalek-sumo: Moved it inside the registration in order to limit the scope because it's only used inside the `register` func.

I can leave it where it was if that's more readable 🤷

pmatyjasek-sumo (approved):

PR #28, opened 2021-06-18
feat(ci): add dependabot for docker

pmalek-sumo:


.github/dependabot.yml
```diff
@@ -13,3 +13,10 @@ updates:
       interval: "daily"
       time: "06:27"
       timezone: "Europe/Warsaw"
+
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
```
astencel-sumo, line 19: In fact, due to a recent [change](https://github.blog/changelog/2021-06-16-dependabot-now-schedules-version-updates-uniformly/) in how Dependabot schedules updates, I believe it's best to delete the `schedule` section altogether. What do you think?

.github/dependabot.yml
```diff
@@ -13,3 +13,10 @@ updates:
...
+      interval: "daily"
+      time: "06:26"
+      timezone: "Europe/Warsaw"
```
astencel-sumo, line 22: In fact, due to a recent [change](https://github.blog/changelog/2021-06-16-dependabot-now-schedules-version-updates-uniformly/) in how Dependabot schedules updates, I believe it's best to delete the `time` and `timezone` entries. What do you think?

astencel-sumo (approved):

.github/dependabot.yml
pmalek-sumo: Thanks! will update the PR

PR #35, opened 2021-06-22
docs(sumologicextension): add sumologicextension README

pmalek-sumo:


pkg/extension/sumologicextension/README.md
```diff
@@ -2,14 +2,57 @@
...
+[httpclientauthenticator]: https://github.com/open-telemetry/opentelemetry-collector/blob/2e84285efc665798d76773b9901727e8836e9d8f/config/configauth/clientauth.go#L34-L39
+[configauth_authentication]: https://github.com/open-telemetry/opentelemetry-collector/blob/3f5c7180c51ed67a6f54158ede5e523822e9659e/config/configauth/configauth.go#L29-L33
+
+## Configuration
+
+* `access_id`: (required) access ID for Sumo Logic service, see
+  [help][credentials_help] for more details
+* `access_key`: (required) access key for Sumo Logic service, see
+  [help][credentials_help] for more details
+* `collector_name`: (required) name that will be used for locally stored
```
perk-sumo, line 35: Why is that required?

astencel-sumo (approved):

perk-sumo (approved):
One general question but not necessarily for this PR :)

pkg/extension/sumologicextension/README.md
pmalek-sumo: It's required by the API as of now. We can tweak how it's being done by assigning a default collector name but I suggest we merge this README as is and only then change it when we change the behavior

PR #36, opened 2021-06-22
docs: add configuration to main README

pmalek-sumo:
Some links refer to what was introduced in https://github.com/SumoLogic/sumologic-otel-collector/pull/35 so there's a small dependency between those 2 PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,52 @@ SumoLogic Distro of [OpenTelemetry Collector][otc_link] built with
 [otc_link]: https://github.com/open-telemetry/opentelemetry-collector
 [otc_builder_link]: https://github.com/open-telemetry/opentelemetry-collector-builder
 
+## Configuration
+
+In order to send data to [Sumo Logic][sumologic_webpage] one needs to configure
+the [sumologicextension][sumologicextension] with credentials and define it (the extension) in the same service
+as [sumologicexporter][sumologicexporter] is defined so that it's used as auth
+extension.
+
+```yaml
+extensions:
+  sumologic:
+    access_id: <my_access_id>
+    access_key: <my_access_key>
+    collector_name: <my_collector_name>
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,6 +8,52 @@ SumoLogic Distro of [OpenTelemetry Collector][otc_link] built with
...
+    collection_interval: 30s
+    scrapers:
+      load:
+
+# Define your processors...
+processors:
+
+exporters:
+  sumologic:
+    authkj:
```
astencel-sumo, line 37: Vim artifact warning ⚠️

README.md
```diff
@@ -8,6 +8,52 @@ SumoLogic Distro of [OpenTelemetry Collector][otc_link] built with
...
+      authenticator: sumologic
+
+service:
+  extensions: [sumologic]
```
perk-sumo, line 41: Can we have that predefined by default so that noone has to add this section?

README.md
```diff
@@ -8,6 +8,52 @@ SumoLogic Distro of [OpenTelemetry Collector][otc_link] built with
...
+receivers:
+  hostmetrics:
+    collection_interval: 30s
+    scrapers:
+      load:
+
+# Define your processors...
+processors:
+
+exporters:
```
perk-sumo, line 35: Can we have that predefined by default so that noone has to add this section?

perk-sumo (approved):

README.md
pmalek-sumo: I guess not, if it is possible it would be a hack. Even if it is possible let's discuss it separately and only then change this README

README.md
pmalek-sumo: I guess not, if it is possible it would be a hack. Even if it is possible let's discuss it separately and only then change this README

PR #42, opened 2021-06-23
feat(sumologicextension): add collector fields option to registration

pmalek-sumo:


pmatyjasek-sumo (approved):

pkg/extension/sumologicextension/README.md
```diff
@@ -36,6 +36,9 @@ and can be used as an authenticator for the
   registration info from previous runs or for registration in case those are not found
 * `collector_description`: collector description that will be used for registration
 * `collector_category`: collector category that will be used for registration
+* `collector_fields`: a map of key value pairs that will be used as collector
```
pmm-sumo, line 39: I think we might want to make it clear why user needs setting anything here. Like: _if this is not set, you will not see your attributes in logs_ (not literally)

Also, maybe we should start with some reasonable defaults that might be coming from file receiver?

pmalek-sumo: Sure thing. The behavio you mentioned is just for logs right? Metrics get all the attributes automatically right (+when sending via non otlp you can customize it in the sumologicextension, correct?)

sumo-drosiek: shouldnt't we group all `collector_*` configs as `collector` configuration subsection?

pmalek-sumo: We can but let's do that in a separate PR

pmalek-sumo: Given that collector fields that can be set with this property are [configurable but static](https://help.sumologic.com/Manage/Fields#collector-and-source-fields) (i.e. it's a key value map) I don't see much sense in adding "sensible defaults" unless there's something specific, static that we could add to designated it's OT sending data (but I reckon this might be deduced from a different property/tag already, doesn't it?)

sumo-drosiek (approved):

PR #43, opened 2021-06-24
refactor: Change mechanism of building encryption hash

pmatyjasek-sumo:
Change mechanism of building encryption hash. Use combination of collector name, access ID and access key.

Signed-off-by: Patryk Matyjasek <pmatyjasek@sumologic.com>

### DIFF
--- a/pkg/extension/sumologicextension/README.md
+++ b/pkg/extension/sumologicextension/README.md
@@ -89,10 +89,16 @@ service:
 
 ## Storing credentials
 
-When collector is starting first time Sumo Logic extension is using `access_key` and `access_id` to register collector with API.
-After registration extension gets collector credentials which are used to authenticate instance of collector.
-Credentials are stored on filesystem in `collector_credentials_path` which is by default set to `$HOME/.sumologic-otel-collector`.
-Name of file that contains credentials is created by hashing combination of `collector_name`, `access_id` and `access_key`.
-This mechanism allows to keep a state of collector (whether is registered or not).
-When collector is restarting it checks if state file exists in `collector_credentials_path`.
-If user want to register collector on the same machine and use another state file, need add the collector name to otc config.
+When collector is starting for the first time, Sumo Logic extension is using `access_key` and `access_id`
+to register collector with API.
+Upon registration, the extension gets collector credentials which are used to authenticate the collector
+when sending request to API (heartbeats, sending data etc).
+Credentials are stored on local filesystem to be reused when collector gets restarted (to prevent re-registration).a
+The path that's used to store the credentials files is configured via `collector_credentials_path` which is by default
+set to `$HOME/.sumologic-otel-collector`.
+Name of that file that contains the credentials is created by hashing a string being a combination of `collector_name`, `access_id` and `access_key`.
+This mechanism allows to keep the state of the collector (whether it is registered or not).
+When collector is restarting it checks if the state file exists in `collector_credentials_path`.
+If one would like to register another collector on the same machine then `collector_name` configuration property
+has to be specified in order to register the collector under that specific name which will be used to create
+a separate state file.
--- a/pkg/extension/sumologicextension/README.md
+++ b/pkg/extension/sumologicextension/README.md
@@ -90,13 +90,13 @@ service:
 ## Storing credentials
 
 When collector is starting for the first time, Sumo Logic extension is using `access_key` and `access_id`
-to register collector with API.
+to register the collector with API.
 Upon registration, the extension gets collector credentials which are used to authenticate the collector
 when sending request to API (heartbeats, sending data etc).
-Credentials are stored on local filesystem to be reused when collector gets restarted (to prevent re-registration).a
+Credentials are stored on local filesystem to be reused when collector gets restarted (to prevent re-registration).
 The path that's used to store the credentials files is configured via `collector_credentials_path` which is by default
 set to `$HOME/.sumologic-otel-collector`.
-Name of that file that contains the credentials is created by hashing a string being a combination of `collector_name`, `access_id` and `access_key`.
+Name of that file that contains the credentials is created by hashing a string which is a combination of `collector_name`, `access_id` and `access_key`.
 This mechanism allows to keep the state of the collector (whether it is registered or not).
 When collector is restarting it checks if the state file exists in `collector_credentials_path`.
 If one would like to register another collector on the same machine then `collector_name` configuration property
--- a/pkg/extension/sumologicextension/README.md
+++ b/pkg/extension/sumologicextension/README.md
@@ -86,3 +86,13 @@ service:
       processors: []
       exporters: [sumologic]
 ```
+
+## Storing credentials
+
+When collector is starting first time Sumo Logic extension is using `access_key` and `access_id` to register collector with API.
+After registration extension gets collector credentials which are used to authenticate instance of collector.
+Credentials are stored on filesystem in `collector_credentials_path` which is by default set to `$HOME/.sumologic-otel-collector`.
+Name of file that contains credentials is created by hashing combination of `collector_name`, `access_id` and `access_key`.
+This mechanism allows to keep a state of collector (whether is registered or not).
+When collector is restarting it checks if state file exists in `collector_credentials_path`.
+If user want to register collector on the same machine and use another state file, need add the collector name to otc config.
--- a/pkg/extension/sumologicextension/credentials.go
+++ b/pkg/extension/sumologicextension/credentials.go
@@ -31,11 +31,16 @@ import (
 	"go.uber.org/zap"
 )
 
+type CollectorCredentials struct {
+	CollectorName string                          `json:"collectorName"`
+	Credentials   api.OpenRegisterResponsePayload `json:"collectorCredentials"`
+}
+
 // CredsGetter is an interface to get collector authentication data
 type CredsGetter interface {
 	CheckCollectorCredentials(string) bool
-	GetStoredCredentials(string) (api.OpenRegisterResponsePayload, error)
-	RegisterCollector(ctx context.Context, collectorName string) (api.OpenRegisterResponsePayload, error)
+	GetStoredCredentials(string) (CollectorCredentials, error)
+	RegisterCollector(ctx context.Context, collectorName string) (CollectorCredentials, error)
 }
 
 // credsGetter is a common structure for both types of getting collector credentials
@@ -61,43 +66,43 @@ func (cr credsGetter) CheckCollectorCredentials(key string) bool {
 
 // GetStoredCredentials retrieves collector credentials stored in local file system and then decrypts it
 // using hashed collector name as passphrase.
-func (cr credsGetter) GetStoredCredentials(key string) (api.OpenRegisterResponsePayload, error) {
+func (cr credsGetter) GetStoredCredentials(key string) (CollectorCredentials, error) {
 	filenameHash, err := hash(key)
 	if err != nil {
-		return api.OpenRegisterResponsePayload{}, err
+		return CollectorCredentials{}, err
 	}
 
 	path := path.Join(cr.conf.CollectorCredentialsPath, filenameHash)
 	creds, err := os.Open(path)
 	if err != nil {
-		return api.OpenRegisterResponsePayload{}, err
+		return CollectorCredentials{}, err
 	}
 
 	defer creds.Close()
 	encryptedCreds, err := ioutil.ReadAll(creds)
 	if err != nil {
-		return api.OpenRegisterResponsePayload{}, err
+		return CollectorCredentials{}, err
 	}
 
 	collectorCreds, err := decrypt(encryptedCreds, key)
 	if err != nil {
-		return api.OpenRegisterResponsePayload{}, err
+		return CollectorCredentials{}, err
 	}
 
-	var credentialsInfo api.OpenRegisterResponsePayload
+	var credentialsInfo CollectorCredentials
 	if err = json.Unmarshal(collectorCreds, &credentialsInfo); err != nil {
-		return api.OpenRegisterResponsePayload{}, err
+		return CollectorCredentials{}, err
 	}
 
 	return credentialsInfo, nil
 }
 
 // RegisterCollector registers new collector using registration API and returns collector credentials.
-func (cr credsGetter) RegisterCollector(ctx context.Context, collectorName string) (api.OpenRegisterResponsePayload, error) {
+func (cr credsGetter) RegisterCollector(ctx context.Context, collectorName string) (CollectorCredentials, error) {
 	baseUrl := strings.TrimSuffix(cr.conf.ApiBaseUrl, "/")
 	u, err := url.Parse(baseUrl)
 	if err != nil {
-		return api.OpenRegisterResponsePayload{}, err
+		return CollectorCredentials{}, err
 	}
 	u.Path = registerUrl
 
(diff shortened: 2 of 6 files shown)


pkg/extension/sumologicextension/README.md
```diff
@@ -82,3 +82,13 @@ service:
...
+
+## Storing credentials
+
+When collector is starting first time Sumo Logic extension is using `access_key` and `access_id` to register collector with API.
+After registration extension gets collector credentials which are used to authenticate instance of collector.
+Credentials are stored on filesystem in `collector_credentials_path` which is by default set to `$HOME/.sumologic-otel-collector`.
+Name of file that contains credentials is created by hashing combination of `collector_name`, `access_id` and `access_key`.
+This mechanism allows to keep a state of collector (whether is registered or not).
+When collector is restarting it checks if state file exists in `collector_credentials_path`.
+If user want to register collector on the same machine and use another state file, need add the collector name to otc config.
```
pmalek-sumo, line 94:
```suggestion
When collector is starting for the first time, Sumo Logic extension is using `access_key` and `access_id`
to register collector with API.
Upon registration, the extension gets collector credentials which are used to authenticate the collector
when sending request to API (heartbeats, sending data etc).
Credentials are stored on local filesystem to be reused when collector gets restarted (to prevent re-registration).a
The path that's used to store the credentials files is configured via `collector_credentials_path` which is by default
set to `$HOME/.sumologic-otel-collector`.
Name of that file that contains the credentials is created by hashing a string being a combination of `collector_name`, `access_id` and `access_key`.
This mechanism allows to keep the state of the collector (whether it is registered or not).
When collector is restarting it checks if the state file exists in `collector_credentials_path`.
If one would like to register another collector on the same machine then `collector_name` configuration property
has to be specified in order to register the collector under that specific name which will be used to create
a separate state file.
```

pmalek-sumo (commented):
There's one more thing we need to add in this PR which is to add a default prefix to the collector name when an empty name (default) is being specified.

Something like `sumologic-otel-collector` (and then the following registrations with the same name would create collectors with this name+timestamp suffix.

pkg/extension/sumologicextension/credentials.go
```diff
@@ -48,7 +48,8 @@ type credsGetter struct {
 // CheckCollectorCredentials checks if collector credentials can be found in path
 // configured in the config.
 func (cr credsGetter) CheckCollectorCredentials() bool {
-	filenameHash, err := hash(cr.conf.CollectorName)
+	key := cr.conf.CollectorName + cr.conf.Credentials.AccessID + cr.conf.Credentials.AccessKey
+	filenameHash, err := hash(key)
 	if err != nil {
```
astencel-sumo, line 53: Sorry, this is not strictly related to this PR, but shouldn't we be logging these errors?

pkg/extension/sumologicextension/extension.go
```diff
@@ -124,7 +124,8 @@ func (se *SumologicExtension) storeCollectorCredentials() error {
 	if err := ensureStoreCredentialsDir(se.conf.CollectorCredentialsPath); err != nil {
 		return err
 	}
-	filenameHash, err := hash(se.conf.CollectorName)
+	key := se.conf.CollectorName + se.conf.Credentials.AccessID + se.conf.Credentials.AccessKey
```
astencel-sumo, line 127: Shouldn't the `storeCollectorCredentials` and `ensureStoreCredentialsDir` functions be part of the `credsGetter`?

This would also make it easier to encapsulate the `CollectorName + AccessID + AccessKey` hash creation, which is repeated here and in `credsGetter.GetStoredCredentials` function.

astencel-sumo: Probably a better name than `credsGetter` would then be `credsStore` because it would serve for both retrieving and storing the credentials.

pmalek-sumo: I like `credentialsStore` and I agree that it might be more obvious what it does granted we also have a method to store the credentials on it (not only to get them). 👍 

As for repeated hashing definition: I'd even go as far as make a function that would take the config and return the hashed string. This way if we ever decide to change the implementation this will be the only place to change. 👍 

@pmatyjasek-sumo WDYT?

As for other reorganization ideas I'm in for it but I'd suggest to skip the discussion on those in this PR

astencel-sumo: > make a function that would take the config and return the hashed string
💯

pmatyjasek-sumo: > Probably a better name than `credsGetter` would then be `credsStore` because it would serve for both retrieving and storing the credentials.

Thats why those `storeCollectorCredentials` and `ensureStoreCredentialsDir` are not parts of `credsGetter`. It grabs the authentication data for collector in two ways. If collector wasn't previously registered (file does not exists on filesystem) it register collector through the API and "gets" credentials. If file exists it also "gets" credentials but in this scenario it is taken from file on filesystem. Also both methods requires the same data as `AccessID`, `AccessKey` and `CollectorName` which are stored in config. `storeCollectorCredentials` needs some additional data like `collector credentials`

pkg/extension/sumologicextension/extension.go
```diff
@@ -57,20 +60,30 @@ const (
...
 func newSumologicExtension(conf *Config, logger *zap.Logger) (*SumologicExtension, error) {
+	collectorName := ""
+	uuid := uuid.New()
+	hostname, err := os.Hostname()
+	if err != nil {
+		return nil, err
+	}
 	if conf.CollectorName == "" {
-		return nil, errors.New("collector name is unset")
+		conf.CollectorName = fmt.Sprintf("%s-%s", hostname, uuid.String())
```
sumo-drosiek, line 70: If I restart collector, is a new one created, or the previous name is being used?

pmalek-sumo: @pmatyjasek-sumo is working on this. The idea is that when you don't specify the collector name (i.e. `""`) you'll get one specific name (based on hostname as far as I recall correctly) and yes it should be the same on each subsequent run with the same empty collector name.

sumo-drosiek:
One last concern:

If you don't set `clobber` and restart collector multiple times, you will get:
`collector-name-<timestamp1>-<timestamp2>-<timestamp3>-<timestamp4>`

Am I right?

pkg/extension/sumologicextension/extension.go
```diff
@@ -57,40 +61,60 @@ const (
...
+	if err != nil {
+		return nil, err
+	}
+	credentialsGetter := credsGetter{
+		conf:   conf,
+		logger: logger,
+	}
 	if conf.CollectorName == "" {
-		return nil, errors.New("collector name is unset")
+		key := fmt.Sprintf("%s%s%s", conf.CollectorName, conf.Credentials.AccessID, conf.Credentials.AccessKey)
```
pmalek-sumo, line 75: extract this "algorithm" of creating a key to a function (receiving a config?)

pkg/extension/sumologicextension/extension.go
```diff
@@ -57,40 +61,60 @@ const (
...
+		// If collector name is not set by the user check if the collector was restarted
+		if !credentialsGetter.CheckCollectorCredentials(key) {
+			// If credentials file is not stored on filesystem generate collector name
+			collectorName = fmt.Sprintf("%s-%s", hostname, uuid.String())
+		}
+	} else {
+		collectorName = conf.CollectorName
 	}
 	if conf.Credentials.AccessID == "" || conf.Credentials.AccessKey == "" {
```
pmalek-sumo, line 84: This can be checked as first thing in the function, fast return on empty keys

pkg/extension/sumologicextension/extension.go
```diff
@@ -30,17 +30,21 @@ import (
...
 
+	"github.com/google/uuid"
+
 	"github.com/open-telemetry/opentelemetry-collector-contrib/extension/sumologicextension/api"
 	"go.opentelemetry.io/collector/component"
 	"go.uber.org/zap"
 )
 
 type SumologicExtension struct {
+	collectorName     string
```
pmalek-sumo, line 41: As suggested in the `pkg/extension/sumologicextension/api/register.go`: let's not add this here as it's not obvious what it's being used for: we already have the config? Why add another `collectorName` in the extension struct itself? (those might be the questions that someone reading this for the first time might be asking)

pkg/extension/sumologicextension/extension.go
```diff
@@ -57,40 +61,60 @@ const (
...
+	} else {
+		collectorName = conf.CollectorName
 	}
 	if conf.Credentials.AccessID == "" || conf.Credentials.AccessKey == "" {
 		return nil, errors.New("access_key and/or access_id not provided")
 	}
 	if conf.HeartBeatInterval <= 0 {
 		conf.HeartBeatInterval = DefaultHeartbeatInterval
 	}
+	hashKey := fmt.Sprintf("%s%s%s", conf.CollectorName, conf.Credentials.AccessID, conf.Credentials.AccessKey)
```
pmalek-sumo, line 90: as commented elsewhere: this should be a func

pkg/extension/sumologicextension/extension.go
```diff
@@ -57,40 +61,60 @@ const (
...
+		logger: logger,
+	}
 	if conf.CollectorName == "" {
-		return nil, errors.New("collector name is unset")
+		key := fmt.Sprintf("%s%s%s", conf.CollectorName, conf.Credentials.AccessID, conf.Credentials.AccessKey)
+		// If collector name is not set by the user check if the collector was restarted
+		if !credentialsGetter.CheckCollectorCredentials(key) {
+			// If credentials file is not stored on filesystem generate collector name
+			collectorName = fmt.Sprintf("%s-%s", hostname, uuid.String())
+		}
```
pmalek-sumo, line 80: I believe we need to account for more facts here: if user specifies `clobber` as `true` then do we really want to append uuid? Doesn't make sense to me, the user specified by the clobber flag: 

> Please override that collector name for me, whatever it is. (if it's empty, override the default name that corresponds to empty collector name in the config)

pkg/extension/sumologicextension/extension.go
```diff
@@ -148,7 +172,7 @@ func (se *SumologicExtension) storeCollectorCredentials(credentials api.OpenRegi
 		return err
 	}
 
-	encrypedCreds, err := encrypt(collectorCreds, se.conf.CollectorName)
+	encrypedCreds, err := encrypt(collectorCreds, se.hashKey)
```
pmalek-sumo, line 175:
```suggestion
	encryptedCreds, err := encrypt(collectorCreds, se.hashKey)
```

pkg/extension/sumologicextension/extension.go
```diff
@@ -57,40 +61,60 @@ const (
 )
 
 func newSumologicExtension(conf *Config, logger *zap.Logger) (*SumologicExtension, error) {
+	collectorName := ""
+	uuid := uuid.New()
```
pmalek-sumo, line 65: This is just used once, just use it in place

pkg/extension/sumologicextension/extension.go
```diff
@@ -30,17 +30,21 @@ import (
 	"sync"
 	"time"
 
+	"github.com/google/uuid"
+
```
pmalek-sumo, line 34:
```suggestion
	"github.com/google/uuid"
```

pmalek-sumo (changes requested):

pkg/extension/sumologicextension/extension.go
pmatyjasek-sumo: This is added to avoid overriding user configuration.

pmalek-sumo:
Sure, but this is not obvious what it will be used for (especially since
it's the first field in the struct now)

pkg/extension/sumologicextension/extension.go
```diff
@@ -57,50 +61,85 @@ const (
...
+		conf:   conf,
+		logger: logger,
+	}
 	if conf.CollectorName == "" {
-		return nil, errors.New("collector name is unset")
+		key := createHashKey(conf)
+		// If collector name is not set by the user check if the collector was restarted
+		if !credentialsGetter.CheckCollectorCredentials(key) {
+			// If credentials file is not stored on filesystem generate collector name
+			collectorName = fmt.Sprintf("%s-%s", hostname, uuid.New().String())
```
pmalek-sumo, line 78:
```suggestion
			collectorName = fmt.Sprintf("%s-%s", hostname, uuid.New())
```

If a struct implements [`Stringer`](https://golang.org/pkg/fmt/#Stringer) then no need to add `.String()` call in format string for `%s`

pkg/extension/sumologicextension/extension.go
```diff
@@ -57,50 +61,85 @@ const (
...
-		colCreds, err = se.credentialsGetter.GetStoredCredentials()
+	if se.credentialsGetter.CheckCollectorCredentials(se.hashKey) {
+		colCreds, err = se.credentialsGetter.GetStoredCredentials(se.hashKey)
 		if err != nil {
 			return err
 		}
-		se.logger.Info("Found stored credentials")
+		colName := colCreds.CollectorName
+		if !se.conf.Clobber {
+			if se.collectorName == "" {
```
pmalek-sumo, line 116: Can we add a comment when this happens?

pkg/extension/sumologicextension/extension.go
```diff
@@ -57,50 +61,85 @@ const (
...
+				se.collectorName = colName
+			}
+			se.logger.Info("Found stored credentials")
```
pmalek-sumo, line 119:
```suggestion
			se.logger.Info("Found stored credentials, skipping registration",
				zap.String("collector_name", se.collectorName),
			)
```

pmalek-sumo (commented):
Can we please add more tests for those added use cases (especially with empty collector name and `clobber` flag)?

pmalek-sumo (commented):
What about the tests 😬  ? :)

pkg/extension/sumologicextension/extension_test.go
```diff
@@ -223,3 +226,88 @@ func TestStoreCredentials(t *testing.T) {
...
+	cfg.CollectorCredentialsPath = dir
+
+	se, err := newSumologicExtension(cfg, zap.NewNop())
+	require.NoError(t, err)
+	require.NoError(t, se.Start(context.Background(), componenttest.NewNopHost()))
+	assert.NotEmpty(t, se.collectorName)
+	t.Cleanup(func() {
+		srv.Close()
+		os.RemoveAll(dir)
+	})
```
pmalek-sumo, line 265: Move higher up

pkg/extension/sumologicextension/extension_test.go
```diff
@@ -223,3 +226,88 @@ func TestStoreCredentials(t *testing.T) {
...
+	cfg.ExtensionSettings = config.ExtensionSettings{}
+	cfg.ApiBaseUrl = srv.URL
+	cfg.Credentials.AccessID = "dummy_access_id"
+	cfg.Credentials.AccessKey = "dummy_access_key"
+	cfg.CollectorCredentialsPath = dir
+
+	se, err := newSumologicExtension(cfg, zap.NewNop())
+	require.NoError(t, err)
+	require.NoError(t, se.Start(context.Background(), componenttest.NewNopHost()))
+	assert.NotEmpty(t, se.collectorName)
```
pmalek-sumo, line 261: add TODO to narrow down the checks

pkg/extension/sumologicextension/extension_test.go
```diff
@@ -223,3 +226,88 @@ func TestStoreCredentials(t *testing.T) {
 		require.NoError(t, se.Start(context.Background(), componenttest.NewNopHost()))
 	})
 }
+
+func TestEmptyCollectorName(t *testing.T) {
```
pmalek-sumo, line 230:
```suggestion
func TestRegister_EmptyCollectorName(t *testing.T) {
```

pmalek-sumo (approved):

PR #54, opened 2021-06-29
sumologicexporter: do not overwrite attributes using metadata translation

sumo-drosiek:
User should be able to write it owns attributes (especially _source ones) and we shouldn't overwrite it

### DIFF
--- a/pkg/exporter/sumologicexporter/translate_metadata.go
+++ b/pkg/exporter/sumologicexporter/translate_metadata.go
@@ -42,6 +42,7 @@ var metadataTranslations = map[string]string{
 	"k8s.replicaset.name":     "replicaset",
 	"k8s.statefulset.name":    "statefulset",
 	"service.name":            "service",
+	"file.path.resolved":      "_sourceName",
 }
 
 // translateMetadata renames metadata keys according to metadataTranslations.
--- a/pkg/exporter/sumologicexporter/translate_metadata.go
+++ b/pkg/exporter/sumologicexporter/translate_metadata.go
@@ -49,7 +49,11 @@ var metadataTranslations = map[string]string{
 func translateMetadata(attributes pdata.AttributeMap) {
 	attributes.Range(func(otKey string, value pdata.AttributeValue) bool {
 		if sumoKey, ok := metadataTranslations[otKey]; ok {
-			attributes.Upsert(sumoKey, value)
+			// do not rename attribute if target name already exists
+			if _, ok := attributes.Get(sumoKey); ok {
+				return true
+			}
+			attributes.Insert(sumoKey, value)
 			attributes.Delete(otKey)
 		}
 		return true
--- a/pkg/exporter/sumologicexporter/translate_metadata_test.go
+++ b/pkg/exporter/sumologicexporter/translate_metadata_test.go
@@ -62,7 +62,7 @@ func TestTranslateMetadataLeavesOtherAttributesUnchanged(t *testing.T) {
 	assertAttribute(t, metadata, "three", "three1")
 }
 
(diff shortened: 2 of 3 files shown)


pkg/exporter/sumologicexporter/translate_metadata.go
```diff
@@ -42,12 +42,16 @@ var metadataTranslations = map[string]string{
...
+	"file.path.resolved":      "_sourceName",
 }
 
 // translateMetadata renames metadata keys according to metadataTranslations.
 func translateMetadata(attributes pdata.AttributeMap) {
 	attributes.Range(func(otKey string, value pdata.AttributeValue) bool {
 		if sumoKey, ok := metadataTranslations[otKey]; ok {
+			if _, ok := attributes.Get(sumoKey); ok {
+				return true
+			}
```
pmalek-sumo, line 54: Can we add some comment here? It's not obvious to me what this does.

For all attributes, for all all metadata translated attributes check if particular sumo metadata key is in the metric attributes and if it is then skip it (don't remove it?)?

sumo-drosiek: I added comment

sumo-drosiek: `// do not rename attribute if target name already exists`

pmalek-sumo: Sorry but I don't understand: `do not rename attribute if target name already exists` -> this checks if `sumoKey` is in `attributes`. Without this check we simply overwrite the value under this key (and remove the corresponding OT key).

Maybe this should be using a different wording? If it exists do not change the value of already existing Sumo key and do not remove a corresponding OT key ?

sumo-drosiek: we can use different wording, I understand it that way.

```
			attributes.Insert(sumoKey, value)
			attributes.Delete(otKey)
```
renames (move) attribute from `otKey` to `sumoKey`

if attribute with name `sumoKey` (target) already exists do not overwrite it

pmalek-sumo (approved):

pkg/exporter/sumologicexporter/translate_metadata_test.go
```diff
@@ -62,19 +62,19 @@ func TestTranslateMetadataLeavesOtherAttributesUnchanged(t *testing.T) {
 	assertAttribute(t, metadata, "three", "three1")
 }
 
-func TestTranslateMetadataOverwritesExistingAttribute(t *testing.T) {
+func TestTranslateMetadataDoNotOverwritesExistingAttribute(t *testing.T) {
```
astencel-sumo, line 65:
```suggestion
func TestTranslateMetadataDoesNotOverwriteExistingAttribute(t *testing.T) {
```

pkg/exporter/sumologicexporter/translate_metadata_test.go
```diff
@@ -62,19 +62,19 @@ func TestTranslateMetadataLeavesOtherAttributesUnchanged(t *testing.T) {
...
 	translateMetadata(metadata)
 
-	assert.Equal(t, 1, metadata.Len())
-	assertAttribute(t, metadata, "host", "hostname1")
+	assert.Equal(t, 2, metadata.Len())
+	assertAttribute(t, metadata, "host", "host1")
 }
 
-func TestTranslateMetadataOverwritesMultipleExistingAttributes(t *testing.T) {
+func TestTranslateMetadataDoNotOverwritesMultipleExistingAttributes(t *testing.T) {
```
astencel-sumo, line 77:
```suggestion
func TestTranslateMetadataDoesNotOverwriteMultipleExistingAttributes(t *testing.T) {
```

PR #59, opened 2021-07-01
refactor(sumologicextension): rename collector credentials path to directory

pmalek-sumo:
After first round of feedback it seems this setting is not obvious hence renaming it from `collector_credentials_path` to `collector_credentials_directory`


perk-sumo (approved):
It's a directory so It totally makes sense to me 👍

PR #61, opened 2021-07-02
refactor!: change translate_metadata to translate_attributes

sumo-drosiek:
change `translate_metadata` to `translate_attributes` and translate all attributes (not only those living in `metadata_attributes`)

PR #72, opened 2021-07-05
test(sumologicexporter): Re-add attribute translation tests for logs text format

astencel-sumo:
Sending logs in text format is different to JSON format with regards to metadata treatment, so I propose to keep the attribute translation tests for both JSON and text format (related to #61).


sumo-drosiek (approved):
I didn't feel that those tests improve coverage in any way, but LGTM

pmalek-sumo (approved):

PR #90, opened 2021-07-08
refactor(sumologicexporter): join attribute translation tests into one with subtests

pmalek-sumo:


sumo-drosiek (approved):

astencel-sumo (approved):
This makes it easier and much cleaner/more readable to add separate test cases 👍

PR #98, opened 2021-07-13
refactor(sumologicextension): extract CredentialsStore interface and localFsCredentialsStore implementation

pmalek-sumo:
Allow custom CredentialsStore implementation by using an interface in the extension.


pkg/extension/sumologicextension/credentialsstore_localfs.go
```diff
@@ -0,0 +1,143 @@
...
+	fi, err := os.Stat(path)
+	if err != nil {
+		if err := os.Mkdir(path, 0700); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	// If the directory doesn't have the execution bit then
+	// set it so that we can 'exec' into it.
```
sumo-drosiek, line 135:
```suggestion
	// set it so that we can access files in it.
```

sumo-drosiek (approved):

pkg/extension/sumologicextension/credentialsstore_localfs.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return CollectorCredentials{}, err
+	}
+
+	path := path.Join(cr.collectorCredentialsDirectory, filenameHash)
+	creds, err := os.Open(path)
+	if err != nil {
+		return CollectorCredentials{}, err
+	}
+
```
kkujawa-sumo, line 63: It is only suggestion but I don't like this empty line, I prefer when defer is as close as possible to `os.Open` and here is added to another block of code.

kkujawa-sumo (approved):

pkg/extension/sumologicextension/credentialsstore_localfs.go
pmalek-sumo: I agree and I also prefer it this way, thanks!

pmalek-sumo: Changed

PR #106, opened 2021-07-15
fix(sumologicexporter): do not use sumologic extension if endpoint is set

sumo-drosiek:
Setting `auth: null`, or `auth.authenticator: null` in configiuration doesn't clear the authenticator configuration. In order to that we have to do it in code

### DIFF
--- a/pkg/exporter/sumologicexporter/exporter.go
+++ b/pkg/exporter/sumologicexporter/exporter.go
@@ -478,6 +478,12 @@ func (se *sumologicexporter) start(ctx context.Context, host component.Host) err
 		se.dataUrlLogs = httpSettings.Endpoint
 		se.dataUrlMetrics = httpSettings.Endpoint
 		se.dataUrlTraces = httpSettings.Endpoint
+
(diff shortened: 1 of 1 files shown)


pkg/exporter/sumologicexporter/exporter.go
```diff
@@ -467,10 +467,15 @@ func (se *sumologicexporter) start(ctx context.Context, host component.Host) err
...
 		se.dataUrlTraces = httpSettings.Endpoint
+
+		if strings.HasPrefix(httpSettings.Auth.AuthenticatorName, "sumologic") {
+			httpSettings.Auth = nil
+		}
 	} else {
 		return fmt.Errorf("no auth extension and no endpoint specified")
 	}
 
+	fmt.Printf("%v", host.GetExtensions())
```
pmalek-sumo, line 478: 😬

sumo-drosiek: don't worry, fixed together with failing tests

pmalek-sumo (approved):

PR #115, opened 2021-07-21
feat: add udpdemux tool

pmalek-sumo:
This tool should serve a temporary purpose of fanning out UDP traffic to multiple collectors so that we can verify we are processing the data in them in the similar/the same way.

### DIFF
--- a/pkg/tools/udpdemux/Makefile
+++ b/pkg/tools/udpdemux/Makefile
@@ -1,0 +1,10 @@
+.PHONY: build
+build:
+	go build -v -o udpdemux .
+
+.PHONY: test
+test: build
+
+.PHONY: lint
+lint:
+	golangci-lint run
--- a/pkg/tools/udpdemux/README.md
+++ b/pkg/tools/udpdemux/README.md
@@ -1,0 +1,19 @@
+# udpdemux
+
+`udpdemux` listens on a specified UDP port and forwards all the packets it
+receives onto a specified localhost ports.
+
+## Usage
+
+```
+Listen for UDP traffic and forward to any number of localhost ports
+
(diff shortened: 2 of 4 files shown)


pkg/tools/udpdemux/Makefile
```diff
@@ -0,0 +1,10 @@
+.PHONY: build
+build:
+	go build -v -o udpdemux .
+
+.PHONY: test
+test: build
+
+.PHONY: lint
+lint:
+	golangci-lint run
```
sumo-drosiek, line 10: missing new line at EOL

pkg/tools/udpdemux/README.md
```diff
@@ -0,0 +1,19 @@
+# udpdemux
+
+`udpdemux` listens on a specified UDP port and forwards all the packets it
+receives onto a specified localhost ports.
```
sumo-drosiek, line 4: `all the received packets` ?

pkg/tools/udpdemux/main.go
```diff
@@ -0,0 +1,192 @@
...
+						panic(err)
+					}
+
+					n, err := c.Write(msg.data)
+					if errors.Is(err, net.ErrClosed) {
+						fmt.Printf("Connection to %v already closed: %v\n",
+							c.RemoteAddr(), err,
+						)
+						continue
+					}
```
sumo-drosiek, line 145: Is it possible for UDP?

pmalek-sumo: Shouldn't really be possible but just in case I've put this check here.

sumo-drosiek (approved):

PR #119, opened 2021-07-29
feat(sumologicextension): check locally stored credentials with api before starting the collector

pmalek-sumo:


pkg/extension/sumologicextension/extension.go
```diff
@@ -101,39 +101,17 @@ func createHashKey(conf *Config) string {
...
-		se.logger.Info("Locally stored credentials not found, registering the collector")
-		if colCreds, err = se.registerCollector(ctx, se.collectorName); err != nil {
-			return err
-		}
-		if err := se.credentialsStore.Store(se.hashKey, colCreds); err != nil {
-			se.logger.Error("Unable to store collector credentials", zap.Error(err))
-		}
+	colCreds, registrationDone, err := se.getCredentials(ctx)
+	if err != nil {
+		return err
```
sumo-drosiek, line 114: I think we should add possibility to re-register in case of invalid cached credentials

pkg/extension/sumologicextension/extension.go
```diff
@@ -148,11 +126,18 @@ func (se *SumologicExtension) Start(ctx context.Context, host component.Host) er
...
+	se.httpClient.Transport, err = se.RoundTripper(se.httpClient.Transport)
 	if err != nil {
 		return fmt.Errorf("couldn't create HTTP client transport: %w", err)
 	}
-	se.httpClient.Transport = rt
+
+	if !registrationDone {
+		se.logger.Info("Checking if locally retrieved credentials are still valid...")
+		if err := se.validateCredenials(ctx, colCreds.Credentials); err != nil {
+			return fmt.Errorf("locally stored credentials invalid: %w", err)
```
sumo-drosiek, line 137:
```suggestion
			return fmt.Errorf("locally stored credentials are invalid: %w", err)
```

pkg/extension/sumologicextension/extension.go
pmalek-sumo: Created #120 to track this. Let's keep this PR as is (in terms of behavior) and introduce this in a separate PR when #120 is agreed on how it should behave.

Thanks!

pkg/extension/sumologicextension/extension.go
```diff
@@ -148,11 +126,18 @@ func (se *SumologicExtension) Start(ctx context.Context, host component.Host) er
...
 	// Set the transport so that all requests from se.httpClient will contain
 	// the collector credentials.
-	rt, err := se.RoundTripper(se.httpClient.Transport)
+	se.httpClient.Transport, err = se.RoundTripper(se.httpClient.Transport)
 	if err != nil {
 		return fmt.Errorf("couldn't create HTTP client transport: %w", err)
 	}
-	se.httpClient.Transport = rt
+
+	if !registrationDone {
```
sumo-drosiek, line 134: isn't `registrationDone` already `true` if we store credentials?

pkg/extension/sumologicextension/extension.go
```diff
@@ -148,11 +126,18 @@ func (se *SumologicExtension) Start(ctx context.Context, host component.Host) er
...
+		se.logger.Info("Checking if locally retrieved credentials are still valid...")
+		if err := se.validateCredenials(ctx, colCreds.Credentials); err != nil {
+			return fmt.Errorf("locally stored credentials are invalid: %w", err)
+		}
+		se.logger.Info("Local collector credentials all good, starting up the collector")
+	}
```
sumo-drosiek, line 140: Could we move it to `getCredentials` or it will too much logic in that function

pmalek-sumo: The problem with that is that we "check" the credentials by issuing a heartbeat which requires collector credentials injected into http calls which is done in lines 129:132 with `se.RoundTripper(...)` and this can only be done when we obtain the credentials. Hence validating the credentials can only be done with a configured http client and transport.

sumo-drosiek: ok

sumo-drosiek (approved):

PR #126, opened 2021-08-05
chore(telegraf): use newer telegraf

sumo-drosiek:
Update telegraf with version which supports multiple http targets


sumo-drosiek:
already updated

PR #136, opened 2021-08-12
docs(extensions): Create separate section for sumologicextension

astencel-sumo:
This is to bring order to the configuration docs: Every extension, receiver, processor, exporter should have its own separate configuration section.


perk-sumo (approved):
👍

PR #145, opened 2021-08-13
docs(installation): Create separate section for each installation platform

astencel-sumo:


docs/Installation.md
```diff
@@ -4,28 +4,84 @@ The Sumo Logic OT Distro can be run using either the binary file available in [G
...
 ## Standalone
 
 The Sumo Logic OT Distro is a static Go binary.
 To run it as a standalone process you only need to run the binary file downloaded from
 [Github releases][github_releases] with an appropriate configuration.
 
-1. Set the release version variable:
+Follow the steps for your platform below.
+
+### Linux on Intel
```
kkujawa-sumo, line 21: Maybe it's better to write "Linux on amd64"

astencel-sumo: Yeah wasn't sure how to phrase it...

perk-sumo: I'd go with [X86-64 like Wikipedia redirects](https://en.wikipedia.org/wiki/X86-64).
Also to be fair I'd change the build names to that as well.

perk-sumo: Also it's the same that is shown in the Amazon ECR: 
image

astencel-sumo: After discussing this in team, we decided to go with "amd64" and "arm64" for now as these are the names used by upstream and in binary names.

perk-sumo (approved):

frankreno (commented):
LGTM! Will we be able to automate something to keep the links pointing to the latest release every release?

sumo-drosiek:
We should be: https://docs.github.com/en/github/administering-a-repository/releasing-projects-on-github/linking-to-releases

PR #148, opened 2021-08-16
feat(sumologicextension): fail for 4xx errors except 429

sumo-drosiek:

### DIFF
--- a/pkg/extension/sumologicextension/extension.go
+++ b/pkg/extension/sumologicextension/extension.go
@@ -286,10 +286,19 @@ func (se *SumologicExtension) registerCollector(ctx context.Context, collectorNa
 			zap.Int("status_code", res.StatusCode),
 			zap.String("response", buff.String()),
 		)
-		return CollectorCredentials{}, fmt.Errorf(
-			"failed to register the collector, got HTTP status code: %d",
-			res.StatusCode,
-		)
+
+		// Return unrecoverable error for 4xx status codes except 429
+		if res.StatusCode >= 400 && res.StatusCode < 500 && res.StatusCode != 429 {
+			return CollectorCredentials{}, backoff.Permanent(fmt.Errorf(
+				"failed to register the collector, got HTTP status code: %d",
+				res.StatusCode,
+			))
+		} else {
+			return CollectorCredentials{}, fmt.Errorf(
+				"failed to register the collector, got HTTP status code: %d",
+				res.StatusCode,
+			)
+		}
 	}
 
 	var resp api.OpenRegisterResponsePayload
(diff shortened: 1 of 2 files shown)


pkg/extension/sumologicextension/extension.go
```diff
@@ -286,10 +288,20 @@ func (se *SumologicExtension) registerCollector(ctx context.Context, collectorNa
 			zap.Int("status_code", res.StatusCode),
 			zap.String("response", buff.String()),
 		)
-		return CollectorCredentials{}, fmt.Errorf(
-			"failed to register the collector, got HTTP status code: %d",
-			res.StatusCode,
-		)
+
+		// Return unrecoverable error for 4xx status codes except 429
+		if res.StatusCode >= 400 && res.StatusCode < 500 && res.StatusCode != 429 {
```
pmm-sumo, line 293: We might consider reusing this abstraction: https://github.com/open-telemetry/opentelemetry-collector/blob/1fada742275967c44c7c01445f097793b9d0caae/consumer/consumererror/permanent.go

sumo-drosiek: Could you check it now? as abstraction is not exported I use backoff.Pernament

pkg/extension/sumologicextension/extension.go
```diff
@@ -286,10 +286,19 @@ func (se *SumologicExtension) registerCollector(ctx context.Context, collectorNa
...
+			return CollectorCredentials{}, backoff.Permanent(fmt.Errorf(
+				"failed to register the collector, got HTTP status code: %d",
```
astencel-sumo, line 293: Not strictly related to this very PR, but aren't we losing the original error's message and only passing the response code here? Wouldn't the original error's message potentially contain additional info for the user, helping them fix the problem?

pkg/extension/sumologicextension/extension.go
```diff
@@ -322,8 +331,9 @@ func (se *SumologicExtension) registerCollectorWithBackoff(ctx context.Context,
 		se.logger.Warn("Collector registration failed: ", zap.Error(err))
 
 		nbo := se.backOff.NextBackOff()
-		if nbo == se.backOff.Stop {
-			return CollectorCredentials{}, fmt.Errorf("collector registration failed")
+		// Return error if backoff reaches the limit or uncoverable error is spotted
+		if _, ok := err.(*backoff.PermanentError); nbo == se.backOff.Stop || ok {
```
astencel-sumo, line 335: This `ok` variable name gives absolutely no information on what it means and makes this line incomprehensible.

A name like `errorIsPermanent` would make this line much easier to read and understand.

astencel-sumo (approved):
Added some comments, but not blockers to going ahead with it 👍

pkg/extension/sumologicextension/extension.go
sumo-drosiek: `ok` is used for type casting in golang, so I would live it

pkg/extension/sumologicextension/extension.go
sumo-drosiek: Response is logged above
```
		se.logger.Debug("Collector registration failed",
			zap.Int("status_code", res.StatusCode),
			zap.String("response", buff.String()),
		)
```

pmm-sumo (approved):

PR #151, opened 2021-08-17
Add playbook to install sumologic-otel-collector

kkujawa-sumo:


sumo-drosiek:
I am not sure if its good idea to keep `ansible` in root repository. I would create some common directory for `ansible`, `puppet` etc

sumo-drosiek:
LGTM, but lets discuss directory structure

PR #158, opened 2021-08-19
Add playbook to install sumologic-otel-collector

kkujawa-sumo:


examples/ansible/vars/default.yaml
```diff
@@ -0,0 +1,6 @@
+sumologic_otel_collector:
+  version: 0.0.18
+  src_config_path: config.yml
```
astencel-sumo, line 3: Very minor nit: Can we change this name from `config.yml` to `config.yaml`? My first run failed because I misread this and assumed it was `.yaml` and not `.yml`. Note that all other YAML files in this PR (including this one) have `.yaml` extensions and not `.yml`. :)

astencel-sumo (approved):
Works like a charm 🧙

examples/ansible/vars/default.yaml
kkujawa-sumo: I used steps from: https://github.com/SumoLogic/sumologic-otel-collector/blob/main/docs/Installation.md#systemd-service where `.yml` extension is in use but I see that in other places `.yaml` extension is used. I vote for using `.yaml` ;)

astencel-sumo (approved):
👌

PR #181, opened 2021-08-30
feat(puppet): add manifest and module to install sumologic-otel-collector

kkujawa-sumo:


vagrant/provision.sh
```diff
@@ -24,3 +24,44 @@ curl -LJ "https://golang.org/dl/go${GO_VERSION}.linux-amd64.tar.gz" -o go.linux-
...
+192.168.79.13 puppetserver.example.com puppet
+192.168.79.13 agent.example.com
+END
+
+# Install puppet server & puppet agent
+wget https://apt.puppetlabs.com/puppet6-release-focal.deb
+dpkg -i puppet6-release-focal.deb
+apt-get update -y
+apt-get install puppetserver -y
+apt-get install puppet-agent -y
```
pmalek-sumo, line 39:
```suggestion
apt-get install puppetserver puppet-agent -y
```

Why not together? Should be a touch faster

vagrant/provision.sh
```diff
@@ -24,3 +24,44 @@ curl -LJ "https://golang.org/dl/go${GO_VERSION}.linux-amd64.tar.gz" -o go.linux-
 
 # Install ansible
 pip3 install ansible
+
+# Add puppet hosts
+tee -a /etc/hosts << END
+192.168.79.13 puppetserver.example.com puppet
+192.168.79.13 agent.example.com
```
pmalek-sumo, line 31: What's the purpose of those?

sumo-drosiek: if it's essential, could we use `127.0.0.1`?

kkujawa-sumo: Puppet agent and server communicate using hosts names, certificates for agent and server needs to be generated for different different host names to have the same behaviour as in communication with agent on different machine (in vagrant there is server and agent on the single host). 
I changed IP address to `127.0.0.1`.

sumo-drosiek: I suppose using `localhost` and `sumologic-otel-collector` hostnames doesn't work?

kkujawa-sumo: I changed puppet server to sumologic-otel-collector

sumo-drosiek (approved):

PR #224, opened 2021-09-15
feat(sourceprocessor): parse out stream, time and log from docker log

sumo-drosiek:
Due to fluent-bit configuration for sumologic kubernetes collection,
logs from kubernetes with docker log driver are send as json with
`log`, `stream` and `time` keys.
We would like to extract `time` and `stream` to record level attributes
and treat `log` as body of the log

Related issue: https://github.com/SumoLogic/sumologic-kubernetes-collection/issues/1758

### DIFF
--- a/pkg/processor/sourceprocessor/source_processor.go
+++ b/pkg/processor/sourceprocessor/source_processor.go
@@ -16,6 +16,7 @@ package sourceprocessor
 
 import (
 	"context"
+	"encoding/json"
 	"log"
 	"regexp"
 	"strings"
@@ -54,6 +55,13 @@ func (stk sourceKeys) convertKey(key string) string {
 	default:
 		return key
 	}
+}
+
+// dockerLog represents log from k8s using docker log driver send by FluentBit
+type dockerLog struct {
+	Stream string
+	Time   string
+	Log    string
 }
 
 type sourceProcessor struct {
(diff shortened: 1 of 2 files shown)


pkg/processor/sourceprocessor/source_processor.go
```diff
@@ -239,6 +249,29 @@ func (sp *sourceProcessor) ProcessLogs(ctx context.Context, md pdata.Logs) (pdat
...
+		for j := 0; j < ills.Len(); j++ {
+			ill := ills.At(j)
+			logs := ill.Logs()
+			for k := 0; k < logs.Len(); k++ {
+				log := logs.At(k)
+				if log.Body().Type() == pdata.AttributeValueTypeString {
+					err := json.Unmarshal([]byte(log.Body().StringVal()), &dockerLog)
+
+					// Skip if error occures or any of the expected fields is empty
+					if err != nil || dockerLog.Stream == "" || dockerLog.Time == "" || dockerLog.Log == "" {
```
astencel-sumo, line 264: Do we not want to log any errors, warnings?

sumo-drosiek: As customer can have CRI-O logs I don't think so
eventually, we can log everything and add configuration option to enable/disable docker logs parsing

astencel-sumo: Oh I see, so it's not an erroneous situation when an error is returned from `json.Unmarshal`, it just means the log does not come from Docker runtime.

astencel-sumo (approved):

PR #227, opened 2021-09-16
fix(sourceprocessor): don't prevent using attributes from Sumo schema in source templates

pmalek-sumo:
This will help in cases where some want to use `_sourceName` in source template which until now rendered `undefined`.

### DIFF
--- a/pkg/exporter/sumologicexporter/README.md
+++ b/pkg/exporter/sumologicexporter/README.md
@@ -159,6 +159,36 @@ Below is a list of all attribute keys that are being translated.
 
 ## Source Templates
 
+> **IMPORTANT NOTE**:
+>
+> When using non-`OTLP` based format (e.g. `JSON` for logs) metadata attributes
+> used in source templates have to have a regex defined in
+> `metadata_attributes` that would match them.
+>
+> Otherwise the attributes will not be available during source templates rendering.
+> Hence this is correct:
+>
+> ```
+> source_name: "%{k8s.namespace.name}.%{k8s.pod.name}.%{k8s.container.name}"
+> source_category: "%{k8s.namespace.name}/%{k8s.pod.pod_name}"
+> source_host: '%{k8s.pod.hostname}'
+> metadata_attributes:
+>   - k8s.*
+>   - some_other_metadata_regex.*
+> ```
+>
+> While is **not**:
+>
+> ```
+> source_name: "%{k8s.namespace.name}.%{k8s.pod.name}.%{k8s.container.name}"
+> source_category: "%{k8s.namespace.name}/%{k8s.pod.pod_name}"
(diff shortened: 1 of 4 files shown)


pkg/exporter/sumologicexporter/README.md
```diff
@@ -159,6 +159,36 @@ Below is a list of all attribute keys that are being translated.
 
 ## Source Templates
 
+> **IMPORTANT NOTE**:
+>
+> When using non-`OTLP` based format (e.g. `JSON` for logs) metadata attributes
```
sumo-drosiek, line 164:
```suggestion
> When using `HHTP` based format (e.g. `JSON` for logs) metadata attributes
```

pmalek-sumo: I wanted to use this wording but we're still using HTTP based OTLP right? (serialized OTLP payload, in HTTP request)

sumo-drosiek: :exploding_head:

astencel-sumo: This IMPORTANT NOTE applies only to logs, doesn't it? Not to metrics or traces? I think this should be very explicitly stated here.

Great to see the examples! I'd go even further with them - add a sentence describing why the incorrect configuration will not work.

pkg/exporter/sumologicexporter/translate_attributes.go
```diff
@@ -62,10 +62,11 @@ func translateAttributes(attributes pdata.AttributeMap) {
 }
 
 // translateConfigValue renames attribute keys in config values according to attributeTranslations.
+// Example:
```
astencel-sumo, line 65: :clap: Great to see these examples

pkg/exporter/sumologicexporter/translate_attributes_test.go
```diff
@@ -127,7 +127,31 @@ func assertAttribute(t *testing.T, metadata pdata.AttributeMap, attributeName st
 }
 
 func TestTranslateConfigValue(t *testing.T) {
-	translatedValue := translateConfigValue("%{k8s.pod.name}-%{host.name}/%{pod}-%{host}")
+	testcases := []struct {
```
astencel-sumo, line 130: :+1: Awesome tests

pkg/exporter/sumologicexporter/README.md
sumo-drosiek: This is for all of the pillars in case of using source headers

Only in logs they are send as separate header (`X-Sumo-FIelds`)

astencel-sumo (approved):

sumo-drosiek (approved):

PR #228, opened 2021-09-16
fix(k8sprocessor): namespace does not need to be in extracted metadata

kkujawa-sumo:
Fix: namespace does not need to be in extracted metadata when there is following configuration

```
pod_association:
- from: build_hostname
```
without this fix it is required to this section to configuration:

```
extract:
  metadata:
   - namespace
```
related to: https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/1762

### DIFF
--- a/pkg/processor/k8sprocessor/kube/client.go
+++ b/pkg/processor/k8sprocessor/kube/client.go
@@ -373,6 +373,7 @@ func (c *WatchClient) extractField(v string, r FieldExtractionRule) string {
 func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
 	newPod := &Pod{
 		Name:      pod.Name,
+		Namespace: pod.Namespace,
 		Address:   pod.Status.PodIP,
 		PodUID:    string(pod.UID),
 		StartTime: pod.Status.StartTime,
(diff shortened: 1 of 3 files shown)


sumo-drosiek:
@pmm-sumo WDYT?

Not sure if we should go to unify everything into `attributes` or extend the struct?

In addition: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/4310

kkujawa-sumo:
> @pmm-sumo WDYT?
> 
> Not sure if we should unify everything into `attributes` or extend the struct 🤷
> 
> In addition: [open-telemetry/opentelemetry-collector-contrib#4310](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/4310)

The struct is already extended in upstream https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/ac26625c2eea420d6703a15e9bd81d686fb5b8af/processor/k8sprocessor/kube/kube.go#L72

pmalek-sumo:
What's the plan for this PR then? Are we waiting for https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/4310#issuecomment-910365660 ?

pmm-sumo:
I am not sure if I understand the concern here. Namespace is already in the struct in upstream. That will at least make as closer to the upstream :) No?

astencel-sumo (approved):
:ok_hand:

sumo-drosiek:
@pmalek-sumo 
> What's the plan for this PR then? Are we waiting for [open-telemetry/opentelemetry-collector-contrib#4310 (comment)](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/4310#issuecomment-910365660) ?

This is independent because of upstream inertia

---
@pmm-sumo 
> I like the Attrtibutes idea, but I think one step at a time

This is out change to move everything as attributes and we are going there and back :P

sumo-drosiek (approved):

PR #231, opened 2021-09-16
fix(sourceprocessor): fix source_host_key feature

pmalek-sumo:
It seems that using `source_host_key` in source processor didn't work at all.

This PR adds the missing functionality and some tests for that.

### DIFF
--- a/pkg/processor/sourceprocessor/attribute_filler.go
+++ b/pkg/processor/sourceprocessor/attribute_filler.go
@@ -42,12 +42,12 @@ type attributeFiller struct {
 	labels          []string
 }
 
-func createSourceHostFiller() attributeFiller {
+func createSourceHostFiller(sourceHostAttrName string) attributeFiller {
 	return attributeFiller{
 		name:            sourceHostKey,
-		compiledFormat:  "",
+		compiledFormat:  "%s",
 		dashReplacement: "",
-		labels:          make([]string, 0),
+		labels:          []string{sourceHostAttrName},
 		prefix:          "",
 	}
 }
--- a/pkg/processor/sourceprocessor/source_processor_test.go
+++ b/pkg/processor/sourceprocessor/source_processor_test.go
@@ -96,6 +96,24 @@ var (
 	}
 )
 
+func newLogsDataWithLogs(resourceAttrs map[string]string, logAttrs map[string]string) pdata.Logs {
+	ld := pdata.NewLogs()
+	rs := ld.ResourceLogs().AppendEmpty()
+	attrs := rs.Resource().Attributes()
+	for k, v := range resourceAttrs {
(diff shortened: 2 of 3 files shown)


pkg/processor/sourceprocessor/source_processor_test.go
```diff
@@ -149,6 +167,97 @@ func assertSpansEqual(t *testing.T, t1 pdata.Traces, t2 pdata.Traces) {
...
+		})
+
+		require.Equal(t, out.ResourceLogs().Len(), 1)
+		resAttrs := out.ResourceLogs().At(0).Resource().Attributes()
+
+		{
+			_, ok := resAttrs.Get("_sourceName")
+			require.False(t, ok)
+		}
+
```
sumo-drosiek, line 253: Why it doesn't set at all?
shouldn't it be `will-it-work-`/`will-it-work-undefined`?

pmalek-sumo:
That might be a different bug :) or maybe just a different behavior in
rendering templates

pkg/processor/sourceprocessor/source_processor_test.go
pmalek-sumo: That might be a different bug :) or maybe just a different behavior in rendering templates

pmalek-sumo: The reason for this is: https://github.com/SumoLogic/sumologic-otel-collector/blob/1c3ef97a999fa3344b3a66287847337b4348104f/pkg/processor/sourceprocessor/attribute_filler.go#L118

Let's park this and sort this out in a separate PR/issue.

pkg/processor/sourceprocessor/attribute_filler.go
```diff
@@ -42,12 +42,12 @@ type attributeFiller struct {
 	labels          []string
 }
 
-func createSourceHostFiller() attributeFiller {
+func createSourceHostFiller(sourceHostAttrName string) attributeFiller {
```
pmm-sumo, line 45: Just thinking out loud:
* perhaps we could add sourceHost as config option? (that would make it more streamline with other options)?
* should we use extractFormat() here instead of filling the struct inline?

pmalek-sumo: * you mean to provide an exact host name to be set as `_sourceHost`? As an alternative to get it from an attribute allowing both to be configured?
* I don't see the reason for that since filling this inline here works well enough. And using `extractFormat` might make it more complicated since we'll need to "bend" the logic to fit all source related templates 🤷

pmm-sumo: Yeah, the only argument I have is consistency, I see your point though. That works

pmm-sumo (approved):

PR #235, opened 2021-09-17
feat(filterprocessor): enable log filtering

pmalek-sumo:
This PR enables `strict` and `regexp` filtering of logs based on resource attributes using filterprocessor.

ref: https://github.com/pmalek-sumo/opentelemetry-collector-contrib/commit/6643765789daf85c51150b5de8bb64a82d9c52e0

related PR upstream that will introduce this when merged: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/5237


perk-sumo (approved):

kkujawa-sumo:
If we plan to use this processor it will be to add it to documentation https://github.com/SumoLogic/sumologic-otel-collector/blob/main/docs/Configuration.md#processors, without link to proper docs it is difficult to find how to configure this processor.

PR #236, opened 2021-09-17
fix(sumologicexporter): don't send null logs

pmalek-sumo:
To prevent sending `null` logs check if the body is not empty before attaching it like in the attached screenshot:

image

### DIFF
--- a/pkg/exporter/sumologicexporter/sender.go
+++ b/pkg/exporter/sumologicexporter/sender.go
@@ -181,7 +181,11 @@ func (s *sender) logToText(record pdata.LogRecord) string {
 // logToJSON converts LogRecord to a json line, returns it and error eventually
 func (s *sender) logToJSON(record pdata.LogRecord) (string, error) {
 	data := s.filter.filterOut(record.Attributes())
-	data.orig.Upsert(logKey, record.Body())
+
+	// Only append the body when it's not empty to prevent sending 'null' log.
+	if body := record.Body(); !isEmptyAttributeValue(body) {
+		data.orig.Upsert(logKey, body)
+	}
 
 	nextLine, err := json.Marshal(pdata.AttributeMapToMap(data.orig))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


kkujawa-sumo (commented):
Can the test case with sending null logs be added to unit tests? If yes, I recommend to add this.

pmalek-sumo:
> Can the test case with sending null logs be added to unit tests? If yes, I recommend to add this.

I wanted to do this but it seems that it's non trivial to make the exporter send the empty logs due to logic in `appendAndSend`, i.e. the provided body there is empty (which is ok) and we cannot really configure `MaxRequestBodySize` to force the flush (in order to check the asserts in tests). 

I'm open to ideas to actually add a test for that.

ref: https://github.com/SumoLogic/sumologic-otel-collector/blob/2a5baaa9d851bf87738c59859c518b6825815490/pkg/exporter/sumologicexporter/sender.go#L400-L438

pkg/exporter/sumologicexporter/sender.go
```diff
@@ -181,7 +181,16 @@ func (s *sender) logToText(record pdata.LogRecord) string {
 // logToJSON converts LogRecord to a json line, returns it and error eventually
 func (s *sender) logToJSON(record pdata.LogRecord) (string, error) {
 	data := s.filter.filterOut(record.Attributes())
-	data.orig.Upsert(logKey, record.Body())
+
+	// Only append the body when it's not empty to prevent sending 'null' log.
```
astencel-sumo, line 185: How about encapsulating this `if` statement in a` function like `isEmpty`?

```
if (!isLogBodyEmpty(body)) {
  data.orig.Upsert(longKey, body)
}
```

astencel-sumo (approved):

pkg/exporter/sumologicexporter/sender.go
pmalek-sumo: Done (somewhat like proposed :) )

sumo-drosiek (approved):

kkujawa-sumo (approved):

PR #237, opened 2021-09-17
sourceprocessor: don't try to add host for systemd related records

pmalek-sumo:
Instead of relying on this behavior, use e.g. attributes processor in the pipeline.


kkujawa-sumo (approved):
LGTM but fix typo in commit message

PR #242, opened 2021-09-21
Include upstream updates to sumo-owned processors

pmm-sumo:
Please find attached retrofit of upstream processors update

It seems that delimiter update was botched during rebase, I think I fixed that but please take a special look on that during review


sumo-drosiek:
tests are failing :sweat:

pmm-sumo:
> tests are failing 😓

Apologies 😿  - draft PR effect :) In any case, there's the thing with namespace handling which I fixed but I am not sure about, we could discuss later

pmalek-sumo (commented):
I'll continue with the review later as it needs a bit more time to digest :)

pmm-sumo:
I think we should consider deprecating it in favor of [`k8sattributesprocessor`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor) + create our own `k8sextendedprocessor` with the parts that were not accepted upstream

PR #245, opened 2021-09-21
fix(sumologicexporter): do not send attributes in both json body and headers during attribute translations

sumo-drosiek:
If metadata_attributes are configured and json attribute is being used, both names before and after translation should be added to metadata_attributes for proper behavior

adding attributes name before translation to metadata_attributes ensures that they will be send as fields
adding attributes name after translation to metadata_attributes ensures that they won't be send in json body

### DIFF
--- a/pkg/exporter/sumologicexporter/exporter.go
+++ b/pkg/exporter/sumologicexporter/exporter.go
@@ -214,7 +214,7 @@ func (se *sumologicexporter) pushLogsData(ctx context.Context, ld pdata.Logs) er
 		currentMetadata  fields = newFields(pdata.NewAttributeMap())
 		previousMetadata fields = newFields(pdata.NewAttributeMap())
 		errs             []error
-		droppedRecords   []pdata.LogRecord
+		droppedRecords   []logPair
 		err              error
 	)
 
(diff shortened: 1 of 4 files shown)


pkg/exporter/sumologicexporter/exporter.go
```diff
@@ -260,14 +260,25 @@ func (se *sumologicexporter) pushLogsData(ctx context.Context, ld pdata.Logs) er
...
+				// Extract attributes which should be send as json body
+				// and connect them with log record
+				lp := logPair{
+					log:            log,
+					bodyAttributes: sdr.filter.filterOut(attributes),
+				}
+
 				if se.config.TranslateAttributes {
-					translateAttributes(attributes)
+					// Translate fields so they can be properly send in JSON format
```
pmalek-sumo, line 271: Can we rephrase this comment? `properly`  isn't too specific :(

pkg/exporter/sumologicexporter/exporter.go
```diff
@@ -260,14 +260,25 @@ func (se *sumologicexporter) pushLogsData(ctx context.Context, ld pdata.Logs) er
...
+					translateAttributes(lp.bodyAttributes.orig)
+					// Translate metadata co it can be used along with translated source host templates
 					translateAttributes(currentMetadata.orig)
+					// Translate original log attribute in case of using OTLP format
```
pmalek-sumo, line 275: > in case of using OTLP format

... but it's done regardless of configured format? So maybe make it conditional? Or change the comment?

sumo-drosiek: I make it conditional, did the same for JSON, but honestly not sure if this is not an overengineering

sumo-drosiek:
Closing in favor of #266

PR #246, opened 2021-09-21
feat(sumologicexporter): add timestamp field when sending JSON logs

pmalek-sumo:
This is to ensure feature parity with https://github.com/SumoLogic/fluentd-output-sumologic which adds milliseconds timestamp to sent out log entries.

This PR also adds the following configuration fields to `sumologicexporter`:

```yaml
    json_logs:
      # defines whether to include a timestamp field when sending
      # JSON logs, which would contain UNIX epoch timestamp in milliseconds.
      # This option affects JSON log format only.
      # default = true.
      add_timestamp: {true, false}
      # when add_timestamp is set to true then this key defines what is the name
      # of the timestamp key.
      # default = "timestamp".
      timestamp_key: <timestamp_key>
```


sumo-drosiek:
Please add it to config and `README.md`

sumo-drosiek:
please add `timestamp_key` configuration as well
https://github.com/SumoLogic/fluentd-output-sumologic/blame/1.7.2/README.md#L32

pmalek-sumo:
> please add `timestamp_key` configuration as well
> https://github.com/SumoLogic/fluentd-output-sumologic/blame/1.7.2/README.md#L32

Added, @sumo-drosiek PTAL

sumo-drosiek (approved):

PR #247, opened 2021-09-22
doc(filestorageextension): add configuration doc

pmalek-sumo:


sumo-drosiek:
In general LGTM, but it would be nice to have a practical example

pmalek-sumo:
> In general LGTM, but it would be nice to have a practical example
> or you enable it and it works with `filelogreceiver` automagically?

Added an example, PTAL

sumo-drosiek:
> > In general LGTM, but it would be nice to have a practical example
> > or you enable it and it works with `filelogreceiver` automagically?
> 
> Added an example, PTAL

so automagically, interesting :)

sumo-drosiek (approved):

PR #248, opened 2021-09-22
fix(sourceprocessor): fix replacing dotted attributes in templates

astencel-sumo:


pkg/processor/sourceprocessor/source_processor_test.go
```diff
@@ -39,16 +39,20 @@ func createConfig() *Config {
 	return config
 }
 
-var (
-	cfg = createConfig()
-
-	k8sLabels = map[string]string{
+func createK8sLabels() map[string]string {
```
pmalek-sumo, line 42: We need (not in this PR) to change the other shared `var`s into `func`s as well to prevent tests sharing variables

👍  for this change

pmalek-sumo (approved):

PR #250, opened 2021-09-22
fix(sourceprocessor): replace missing attributes with "undefined"

astencel-sumo:


pkg/processor/sourceprocessor/attribute_filler.go
```diff
@@ -113,11 +113,11 @@ func (f *attributeFiller) fillAttributes(atts *pdata.AttributeMap) bool {
 func (f *attributeFiller) resourceLabelValues(atts *pdata.AttributeMap) []interface{} {
 	arr := make([]interface{}, 0)
 	for _, label := range f.labels {
-		value, ok := atts.Get(label)
-		if !ok {
-			return nil
+		if value, found := atts.Get(label); found {
+			arr = append(arr, value.StringVal())
+		} else {
+			arr = append(arr, "undefined")
```
pmalek-sumo, line 119: Define as `const`?

pkg/processor/sourceprocessor/source_processor_test.go
```diff
@@ -278,7 +276,7 @@ func TestTraceSourceProcessor(t *testing.T) {
 	td, err := rtp.ProcessTraces(context.Background(), test)
 	assert.NoError(t, err)
 
-	assertTracesEqual(t, td, want)
+	assertTracesEqual(t, want, td)
```
pmalek-sumo, line 279: 😈

pkg/processor/sourceprocessor/attribute_filler.go
astencel-sumo: This value is only being used once - unless you count the appearances in tests. Do you think I should replace the "undefined"-s in tests with the const as well? I'm on the fence about it .

pmalek-sumo: 🤷  Let's leave it then. I don't have a strong opinion about it but I'm worried about the number of occurrences of this magic string growing because X.

> ... I'm on the fence about it.

Then jump :)

pmalek-sumo (approved):

PR #254, opened 2021-09-24
feat(sourceprocessor)!: use plain attribute names in templates

astencel-sumo:
Instead of telling users to use `namespace`, `pod`, `container` tokens in templates and then mapping them to actual attribute names like `k8s.namespace.name` etc., just use attribute names directly in templates.


pkg/processor/sourceprocessor/README.md
```diff
@@ -2,14 +2,13 @@
 
 The `sourceprocessor` adds `_sourceName` and other tags related to Sumo Logic metadata taxonomy.
 
-It leverages data tagged by `k8sprocessor` and must be after it in the processing chain.
-It has certain expectations on the label names used by `k8sprocessor` which might be configured below.
+It uses data tagged by `k8sprocessor` and must be after it in the processing chain.
```
sumo-drosiek, line 5: IMHO we don't need it. we can mention that default values depends on them :thinking:

sumo-drosiek (approved):
one nit, otherwise LGTM

pmalek-sumo (approved):

PR #258, opened 2021-09-27
feat(sumologicexporter): make source metadata attributes always available in source templates

pmalek-sumo:


pkg/exporter/sumologicexporter/exporter.go
```diff
@@ -137,6 +138,40 @@ func initExporter(cfg *Config, createSettings component.ExporterCreateSettings)
...
+	}
+	if !sourceHost {
+		metadataAttributes = append(metadataAttributes, attributeKeySourceHost)
+	}
+	if !sourceName {
+		metadataAttributes = append(metadataAttributes, attributeKeySourceName)
+	}
+
+	return metadataAttributes
+}
```
sumo-drosiek, line 173: cannot we just add element to the list?
is it any risk related to duplication?

pmalek-sumo: I don't believe there's risk of duplicating anything in terms of headers/attributes but we do it once here and n times ( where n is the number of metadata attribute regexes) for each processed entry right? So it might be worth the hassle here

pkg/exporter/sumologicexporter/sender_test.go
```diff
@@ -765,6 +819,71 @@ func TestOverrideSourceHost(t *testing.T) {
 	})
 }
 
+func TestLogsDontSendSourceFieldsInXSumoFieldsHeaderWhenCorrespondingXSumoHeaderIsSet(t *testing.T) {
```
sumo-drosiek, line 822: not only for `CorrespondingXSumoHeaderIsSet`

pmalek-sumo: 👍

sumo-drosiek (approved):

PR #267, opened 2021-09-29
Refactor translation for fields and introduce logPair struct

sumo-drosiek:


pkg/exporter/sumologicexporter/exporter_test.go
```diff
@@ -503,7 +515,8 @@ func TestPushJSONLogsWithAttributeTranslation(t *testing.T) {
 		func(w http.ResponseWriter, req *http.Request) {
 			body := extractBody(t, req)
 			var regex string
-			regex += `{"InstanceType":"wizard","host":"harry-potter","log":"Example log","timestamp":\d{13}}`
+			// Mind that host attribute is not being send in log body
+			regex += `{"InstanceType":"wizard","log":"Example log","timestamp":\d{13}}`
```
sumo-drosiek, line 519: @pmalek-sumo why not `regex := `?

pmalek-sumo: Other places had it this way (mostly to keep the indentation at the same level, then you have it all nicely look as if it was a block of text)

pkg/exporter/sumologicexporter/sender.go
```diff
@@ -303,15 +311,17 @@ func (s *sender) sendLogs(ctx context.Context, flds fields) ([]pdata.LogRecord,
...
 	rl := ld.ResourceLogs().AppendEmpty()
 	ill := rl.InstrumentationLibraryLogs().AppendEmpty()
 	logs := ill.Logs()
 	logs.EnsureCapacity(len(s.logBuffer))
 	for _, record := range s.logBuffer {
 		log := logs.AppendEmpty()
-		record.CopyTo(log)
+		record.log.CopyTo(log)
+		log.Attributes().Clear()
+		record.attributes.CopyTo(log.Attributes())
```
pmalek-sumo, line 324: Perhaps we can use `EnsureCapacity` before copying?

pmalek-sumo (approved):

PR #270, opened 2021-09-29
feat(sourceprocessor)!: allow templating of source host field

astencel-sumo:
This change replaces `source_host_key` configuration option with `source_host` option.
This is for consistency, to allow templating of source host field in the same way that source name and source category fields can be templated.

BREAKING CHANGE: `source_host_key` option is removed in favor of `source_host` option.
To upgrade, take the value of `source_host_key`, surround it with `'%{` and `}'`, and put it in `source_host` option.
For example: `source_host_key: k8s.pod.hostname` should be changed to `source_host: '%{k8s.pod.hostname}'`

### DIFF
--- a/pkg/processor/sourceprocessor/README.md
+++ b/pkg/processor/sourceprocessor/README.md
@@ -13,6 +13,10 @@ processors:
     # default: ""
     collector: <collector>
 
+    # Template for source host, put in `_sourceHost` tag.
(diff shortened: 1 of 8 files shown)


sumo-drosiek (approved):
It is not breaking change for me

pmalek-sumo (approved):

PR #288, opened 2021-10-05
tests(integration): add config for metricfrequencyprocessor

pmalek-sumo:


sumo-drosiek:
commit message doesn't make sense for me. I would expect rather:

`tests(integration): add config for metricfrequencyprocessor`

sumo-drosiek (approved):

PR #291, opened 2021-10-06
feat(otelcol): enable persistent queue

pmalek-sumo:


pmm-sumo:
Don't we need to update OTC core version?

pmalek-sumo:
> Don't we need to update OTC core version?

To what version specifically? It's already bumped to `0.36.0` https://github.com/SumoLogic/sumologic-otel-collector/blob/391ca97473f0ddbf9e3e01db7c5dad012f001ff1/otelcolbuilder/.otelcol-builder.yaml#L9

and the test that I've added seems to pass

pmm-sumo (approved):

PR #337, opened 2021-11-12
Include additional upstream OTC components

pmm-sumo:
This adds almost all receiver and processor components from OpenTelemetry Collector Contrib.

It also lists otlpexporter and otlphttpexporter as available (since they are part of core and could be useful for pipelining collectors)

Additionally, it bumps the core version to v0.38.0 (which includes a quite important fix for K8s tracing workloads)

There's no documentation provided for the newly added components. We can discuss if it should be provided later or skipped altogether (keeping our distro docs in sync with upstream for all components might require significant amount of work and maybe we should just point to upstream docs)

The components add ~6 MiB (or 4%) extra to the macOS binary
```
-rwxr-xr-x   1 pmaciolek  staff  148868536 Nov 12 12:13 otelcol-sumo
-rwxr-xr-x   1 pmaciolek  staff  143106848 Nov 10 13:26 otelcol-sumo-v0.0.40
```

### DIFF
--- a/pkg/extension/sumologicextension/go.mod
+++ b/pkg/extension/sumologicextension/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/stretchr/testify v1.7.0
 	go.opentelemetry.io/collector v0.38.0
 	go.uber.org/zap v1.19.1
+	google.golang.org/grpc v1.41.0
 )
 
 require (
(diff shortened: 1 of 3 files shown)


pmalek-sumo:
To mitigate test failures e.g. https://github.com/SumoLogic/sumologic-otel-collector/runs/4189069712?check_suite_focus=true you can use:

```
make gomod-download-all
```

from root dir

pkg/extension/sumologicextension/go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/stretchr/testify v1.7.0
 	go.opentelemetry.io/collector v0.38.0
 	go.uber.org/zap v1.19.1
+	google.golang.org/grpc v1.41.0
```
pmm-sumo, line 12: BTW, this is an effect of `make gomod-download-all`

pmalek-sumo: That's correct: because of your patch in https://github.com/SumoLogic/sumologic-otel-collector/pull/348 grpc is now a direct (instead of indirect) dependency 👍

pmalek-sumo (approved):
🚀

PR #352, opened 2021-11-25
chore: bump ot core to 0.40.0

pmalek-sumo:
Using

* https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.40.0
* https://github.com/open-telemetry/opentelemetry-collector/releases/tag/cmd%2Fbuilder%2Fv0.40.0


otelcolbuilder/.otelcol-builder.yaml
```diff
@@ -115,8 +115,6 @@ replaces:
   # TODO: remove this when:
   # - regexp log filtering is released upstream:
   #   PR: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/5680
-  # - record level log filtering is available upstream:
-  #   commit: https://github.com/pmalek-sumo/opentelemetry-collector-contrib/commit/11de2c03d3dbbde2e73b8a816318a2515c843985
```
pmalek-sumo, line 119: This is already merged https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/5418 and available upstream https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/cb2a35d282351a071707c76448632fd7ffebc963 since `0.37.0`

pkg/processor/k8sprocessor/pod_association.go
```diff
@@ -32,9 +32,16 @@ import (
 func extractPodID(ctx context.Context, attrs pdata.AttributeMap, associations []kube.Association) (podIdentifierKey string, podIdentifierValue kube.PodIdentifier) {
 	hostname := stringAttributeFromMap(attrs, conventions.AttributeHostName)
 	var connectionIP kube.PodIdentifier
-	if c, ok := client.FromContext(ctx); ok {
-		connectionIP = kube.PodIdentifier(c.IP)
+
+	if c := client.FromContext(ctx); c.Addr != nil {
```
pmm-sumo, line 36: Interesting change, I believe it's not due to update of the core?

otelcolbuilder/.otelcol-builder.yaml
```diff
@@ -115,17 +115,15 @@ replaces:
...
-  - github.com/open-telemetry/opentelemetry-collector-contrib/internal/coreinternal => github.com/SumoLogic/opentelemetry-collector-contrib/internal/coreinternal 4e57c94574d31e2ef1df385167421e6c54d60d75
-  - github.com/open-telemetry/opentelemetry-collector-contrib/processor/attributesprocessor => github.com/SumoLogic/opentelemetry-collector-contrib/processor/attributesprocessor 4e57c94574d31e2ef1df385167421e6c54d60d75
-  - github.com/open-telemetry/opentelemetry-collector-contrib/processor/resourceprocessor => github.com/SumoLogic/opentelemetry-collector-contrib/processor/resourceprocessor 4e57c94574d31e2ef1df385167421e6c54d60d75
+  # commit: https://github.com/SumoLogic/opentelemetry-collector-contrib/commit/6077b2f2274ecaaf220f9d4f11c461b2cbd8cd68
+  # branch: https://github.com/SumoLogic/opentelemetry-collector-contrib/tree/v0.40.0-filterprocessor
+  - github.com/open-telemetry/opentelemetry-collector-contrib/processor/filterprocessor => github.com/SumoLogic/opentelemetry-collector-contrib/processor/filterprocessor 6077b2f2274ecaaf220f9d4f11c461b2cbd8cd68
```
pmm-sumo, line 126: 🦸🏻

pkg/processor/k8sprocessor/pod_association.go
```diff
@@ -32,9 +32,16 @@ import (
...
+		ip, _, err := net.SplitHostPort(c.Addr.String())
```
pmm-sumo, line 37: In what circumstances port might be available in the context?

pkg/processor/k8sprocessor/pod_association.go
```diff
@@ -32,9 +32,16 @@ import (
...
+		if err == nil {
+			connectionIP = kube.PodIdentifier(ip)
+		} else if ip := net.ParseIP(c.Addr.String()); ip != nil {
```
pmm-sumo, line 40: My understanding is this is the most common path. How much extra processing time does it add? It will executed for each resourceX package

pkg/processor/k8sprocessor/pod_association.go
pmm-sumo: I don't see it present in the upstream code. Should we put it there? https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/6fff7610fdf982db67904ee91006e63faf91a254/processor/k8sattributesprocessor/pod_association.go#L89-L96

pkg/processor/k8sprocessor/pod_association.go
pmalek-sumo: I'm not too familiar with how tracing workflow works 😕 The UTs pass, if we can add an additional set of tests that would make us more comfortable I'm all in on that.

pkg/processor/k8sprocessor/pod_association.go
pmalek-sumo: I believe this is the change

pmm-sumo: It seems it will always be just the IP address though: https://github.com/open-telemetry/opentelemetry-collector/blob/ac9619e17e22da915b25dc75f0de61274c1760ce/client/client.go#L144-L162

pmm-sumo: And there's some splitting already in-place: https://github.com/open-telemetry/opentelemetry-collector/blob/ac9619e17e22da915b25dc75f0de61274c1760ce/client/client.go#L144-L162

pkg/processor/k8sprocessor/pod_association.go
pmm-sumo: I think this is largely unnecessary at this point (see the comment above). The `client.Info` should already take care of extracting the address. I believe we might skip the extra processing and do `c.Addr.String()` without additional checks (though they do not seem to do harm, except for extra processing time, probably miniscule)

pkg/processor/k8sprocessor/pod_association.go
pmalek-sumo: @pmm-sumo Can you please check now.

pmm-sumo (approved):

PR #354, opened 2021-12-02
feat(k8sprocessor): add ignored pod names as config parameter

pmalek-sumo:
This PR retrofits upstream PR https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/3520

### DIFF
--- a/pkg/processor/k8sprocessor/options.go
+++ b/pkg/processor/k8sprocessor/options.go
@@ -365,3 +365,22 @@ func WithDelimiter(delimiter string) Option {
 		return nil
 	}
 }
+
+// WithExcludes allows specifying pods to exclude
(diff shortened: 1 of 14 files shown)


pkg/processor/k8sprocessor/options.go
```diff
@@ -365,3 +365,34 @@ func WithDelimiter(delimiter string) Option {
 		return nil
 	}
 }
+
+var defaultExcludes = ExcludeConfig{
+	Pods: []ExcludePodConfig{
```
pmm-sumo, line 370: As discussed, we could keep the list empty and populate it in Helm chart only for tracing pipeline

pmm-sumo (approved):

PR #362, opened 2021-12-14
fix(k8sprocessor): change info to debug log when logging about pod ownership lookup

pmalek-sumo:


pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -271,9 +271,11 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
 	}
 
 	if c.Rules.OwnerLookupEnabled {
-		c.logger.Info("pod owner lookup", zap.Any("pod", pod.Name), zap.Any("owner refs", pod.OwnerReferences))
+		c.logger.Debug("pod owner lookup",
+			zap.String("pod.Name", pod.Name),
+			zap.Any("pod.OwnerReferences", pod.OwnerReferences),
+		)
 		owners := c.op.GetOwners(pod)
-		c.logger.Info("pod owner lookup #2", zap.Any("owners", owners))
```
swiatekm-sumo, line 276: There isn't a second lookup in this function, right?

swiatekm-sumo (approved):

pkg/processor/k8sprocessor/kube/client.go
pmalek-sumo: That's why this is getting deleted in this PR :)

PR #374, opened 2021-12-28
fix: add match glob pattern to git describe

pmalek-sumo:
Due to [recent changes in tags that are being pushed to this repo](https://github.com/SumoLogic/sumologic-otel-collector/pull/370) we need to be more specific w.r.t to the tags that we can match when calling `git describe` to get the version string that we embed in the binary.

This is to prevent getting a version string like:

```
pkg/processor/metricfrequencyprocessor/v0.0.44-beta.0-1-geb02b42d1d
```


kkujawa-sumo:
@pmalek-sumo just for future reference, how does desired form of version string look like?

kkujawa-sumo (approved):

pmalek-sumo:
> @pmalek-sumo just for future reference, how does desired form of version string look like?

`v0.0.44-beta.0-1-geb02b42d1d` or `v0.0.44-beta.0` which depends on the fact if we're on a tag or not (as it always is with `git describe ...`)

PR #381, opened 2022-01-12
 chore(vagrant): add Markdown lint to vagrant environment

kkujawa-sumo:


vagrant/provision.sh
```diff
@@ -78,3 +79,7 @@ add-apt-repository \
    stable"
 apt-get install -y docker-ce docker-ce-cli containerd.io
 usermod -aG docker vagrant
+
+# Install Markdown lint
+# see: https://github.com/markdownlint/markdownlint
+gem install mdl
```
pmalek-sumo, line 85: I believe we should migrate to `markdownlint-cli` just as we did in https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2011/files because of issues with disabling the linter inline.

PR #384, opened 2022-01-14
WIP: fix(k8sprocessor): add service attributes immediately after endpoints are created

pmalek-sumo:
Do not merge - treat this PR as RFC :)

This PR highlights problems in k8sprocessor but doesn't solve all of them.

* It's not trivial to resolve the problem of not immediately updating pod ownership (assigning replicaset/statefulset/service etc owners to pods). The problem lies in the separate update mechanism for pods and for the rest of k8s objects. When we receive an update for pod, we fill its attributes and only when subsequently its owner object update comes in then we fill it in the cache and the subsequent update (or resync on which we rely - which happens [every 5 min](https://github.com/SumoLogic/sumologic-otel-collector/blob/ba870d20af6a79cda98c1a888ce18e77af60468b/pkg/processor/k8sprocessor/kube/informer.go#L51)) will make the pod contain the owner info.
The attempt to solve it in this PR does a somewhat good job of feeding back service information to the client but it doesn't do it in a thread safe manner. It would require even more locking/channels etc to make `Attributes` access thread safe which might not be worth the hussle.

* There's a data race introduced in the service feedback notification loop test (and not only) - where accessing `Pod`'s `Attributes` is done from multiple goroutines without proper locking/etc.

  ```
  ==================
  WARNING: DATA RACE
  Write at 0x00c000138450 by goroutine 193:
    runtime.mapassign_faststr()
        /Users/pmalek/.gvm/gos/go1.17.6/src/runtime/map_faststr.go:202 +0x0
    github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube.(*WatchClient).OwnerFeedbackLoop()
        /Users/pmalek/code/opentelemetry/sumologic-otel-collector/pkg/processor/k8sprocessor/kube/client.go:150 +0x371
    github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube.New·dwrap·2()
        /Users/pmalek/code/opentelemetry/sumologic-otel-collector/pkg/processor/k8sprocessor/kube/client.go:106 +0x47
  
  Previous read at 0x00c000138450 by goroutine 164:
    reflect.maplen()
        /Users/pmalek/.gvm/gos/go1.17.6/src/runtime/map.go:1360 +0x0
    reflect.Value.Len()
        /Users/pmalek/.gvm/gos/go1.17.6/src/reflect/value.go:1509 +0x268
    reflect.deepValueEqual()
        /Users/pmalek/.gvm/gos/go1.17.6/src/reflect/deepequal.go:132 +0xe26
    reflect.DeepEqual()
        /Users/pmalek/.gvm/gos/go1.17.6/src/reflect/deepequal.go:218 +0x404
    github.com/stretchr/testify/assert.ObjectsAreEqual()
        /Users/pmalek/.gvm/pkgsets/go1.17.6/global/pkg/mod/github.com/stretchr/testify@v1.7.0/assert/assertions.go:65 +0x184
    github.com/stretchr/testify/assert.ObjectsAreEqualValues()
        /Users/pmalek/.gvm/pkgsets/go1.17.6/global/pkg/mod/github.com/stretchr/testify@v1.7.0/assert/assertions.go:81 +0x79
    github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube.Test_PodIsAnnotatedWithOwnershipInfoImmediatelyAfterOwnersAreCreated.func3()
        /Users/pmalek/code/opentelemetry/sumologic-otel-collector/pkg/processor/k8sprocessor/kube/client_test.go:1182 +0x24c
    github.com/stretchr/testify/assert.Eventually.func1()
        /Users/pmalek/.gvm/pkgsets/go1.17.6/global/pkg/mod/github.com/stretchr/testify@v1.7.0/assert/assertions.go:1655 +0x39
  
  Goroutine 193 (running) created at:
    github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube.New()
        /Users/pmalek/code/opentelemetry/sumologic-otel-collector/pkg/processor/k8sprocessor/kube/client.go:106 +0x877
    github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube.Test_PodIsAnnotatedWithOwnershipInfoImmediatelyAfterOwnersAreCreated()
        /Users/pmalek/code/opentelemetry/sumologic-otel-collector/pkg/processor/k8sprocessor/kube/client_test.go:983 +0x404
    testing.tRunner()
        /Users/pmalek/.gvm/gos/go1.17.6/src/testing/testing.go:1259 +0x22f
    testing.(*T).Run·dwrap·21()
        /Users/pmalek/.gvm/gos/go1.17.6/src/testing/testing.go:1306 +0x47
  
  Goroutine 164 (finished) created at:
    github.com/stretchr/testify/assert.Eventually()
        /Users/pmalek/.gvm/pkgsets/go1.17.6/global/pkg/mod/github.com/stretchr/testify@v1.7.0/assert/assertions.go:1655 +0x3b7
    github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube.Test_PodIsAnnotatedWithOwnershipInfoImmediatelyAfterOwnersAreCreated()
        /Users/pmalek/code/opentelemetry/sumologic-otel-collector/pkg/processor/k8sprocessor/kube/client_test.go:1172 +0x104c
    testing.tRunner()
        /Users/pmalek/.gvm/gos/go1.17.6/src/testing/testing.go:1259 +0x22f
    testing.(*T).Run·dwrap·21()
        /Users/pmalek/.gvm/gos/go1.17.6/src/testing/testing.go:1306 +0x47
  ==================
      testing.go:1152: race detected during execution of test
  --- FAIL: Test_PodIsAnnotatedWithOwnershipInfoImmediatelyAfterOwnersAreCreated (0.02s)
  ```

### DIFF
--- a/pkg/processor/k8sprocessor/kube/client.go
+++ b/pkg/processor/k8sprocessor/kube/client.go
@@ -66,13 +66,14 @@ func New(
 	newOwnerProviderFunc OwnerProvider,
 	delimiter string,
 ) (Client, error) {
+	stopCh := make(chan struct{})
 	c := &WatchClient{
 		logger:       logger,
 		Rules:        rules,
 		Filters:      filters,
 		Associations: associations,
 		Exclude:      exclude,
-		stopCh:       make(chan struct{}),
+		stopCh:       stopCh,
 		delimiter:    delimiter,
 		Pods:         map[PodIdentifier]*Pod{},
 	}
@@ -102,6 +103,7 @@ func New(
 		if err != nil {
 			return nil, err
 		}
+		go c.OwnerFeedbackLoop(stopCh)
 	}
 
 	logger.Info(
@@ -115,6 +117,51 @@ func New(
 
 	c.informer = newInformer(c.kc, c.Filters.Namespace, labelSelector, fieldSelector)
 	return c, err
+}
+
+// OwnerFeedbackLoop starts a loop that will receive notifications about updates
+// to endpoints objects which are then used to update corresponding pods with
+// ownership info.
+// For info it only updates the services attribute. See
+// Test_PodIsAnnotatedWithOwnershipInfoImmediatelyAfterOwnersAreCreated test in
+// client_test.go for details.
+func (c *WatchClient) OwnerFeedbackLoop(stopCh <-chan struct{}) {
+	ch := c.op.GetPodUpdateChan()
+	for {
+		select {
+		case <-stopCh:
+			return
+
+		case pod, ok := <-ch:
+			if !ok {
+				return
+			}
+
+			p, ok := c.GetPod(generatePodID(pod))
+			if !ok {
+				continue
+			}
+
+			if c.Rules.OwnerLookupEnabled {
+				if c.Rules.ServiceName {
+					services := c.op.GetServices(pod.Name)
+					if len(services) > 0 {
+						c.m.Lock()
+						// TODO: this cannot be done this way because Attributes are
+						// being accessed elsewhere e.g.
+						// * in kubernetesprocessor.getAttributesForPod()
+						// * in tests when assertions on attributes are being made.
+						//
+						// which create data race situations.
+						p.Attributes[c.Rules.Tags.ServiceName] =
+							strings.Join(services, c.delimiter)
+						c.Pods[PodIdentifier(pod.Name)] = p
+						c.m.Unlock()
+					}
+				}
+			}
+		}
+	}
 }
 
 // Start registers pod event handlers and starts watching the kubernetes cluster for pod changes.
@@ -319,7 +366,10 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
 		}
 
 		if c.Rules.ServiceName {
-			tags[c.Rules.Tags.ServiceName] = strings.Join(c.op.GetServices(pod), c.delimiter)
+			services := c.op.GetServices(pod.Name)
+			if len(services) > 0 {
+				tags[c.Rules.Tags.ServiceName] = strings.Join(services, c.delimiter)
+			}
 		}
 
 	}
@@ -427,8 +477,12 @@ func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
 	}
 	// Use pod_name.namespace_name identifier
 	if newPod.Name != "" && newPod.Namespace != "" {
-		c.Pods[PodIdentifier(fmt.Sprintf("%s.%s", newPod.Name, newPod.Namespace))] = newPod
+		c.Pods[generatePodID(newPod)] = newPod
 	}
+}
+
+func generatePodID(pod *Pod) PodIdentifier {
+	return PodIdentifier(fmt.Sprintf("%s.%s", pod.Name, pod.Namespace))
 }
 
 func (c *WatchClient) forgetPod(pod *api_v1.Pod) {
--- a/pkg/processor/k8sprocessor/kube/client_test.go
+++ b/pkg/processor/k8sprocessor/kube/client_test.go
@@ -15,6 +15,7 @@
 package kube
 
 import (
+	"context"
 	"fmt"
 	"reflect"
 	"regexp"
@@ -28,6 +29,7 @@ import (
 	"go.uber.org/zap/zaptest/observer"
 	api_v1 "k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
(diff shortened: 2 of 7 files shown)


pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -117,6 +119,51 @@ func New(
...
+			return
+
+		case pod, ok := <-ch:
+			if !ok {
+				return
+			}
+
+			p, ok := c.GetPod(generatePodID(pod))
+			if !ok {
+				continue
```
swiatekm-sumo, line 142: We should log something for both of these faults. Either an info or a warning, at a glance.

pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -117,6 +119,51 @@ func New(
...
+			}
+
+			if c.Rules.OwnerLookupEnabled {
```
swiatekm-sumo, line 145: Is this check necessary? We're only creating the owner cache and starting this loop if this is enabled.

pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -117,6 +119,51 @@ func New(
...
+				if c.Rules.ServiceName {
+					services := c.op.GetServices(pod.Name)
+					if len(services) > 0 {
+						c.m.Lock()
+						// TODO: this cannot be done this way because Attributes are
+						// being accessed elsewhere e.g.
+						// * in kubernetesprocessor.getAttributesForPod()
+						// * in tests when assertions on attributes are being made.
+						//
+						// which create data race situations.
```
swiatekm-sumo, line 155: Is this a problem because other contexts hold references to the Pod object? If so, can't we make a copy of it, replace the Attributes, and store the copy? That's what `addOrUpdatePod` does, in effect, although it creates a new object from the data it gets, as opposed to copying the existing one.

pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -117,6 +119,51 @@ func New(
 	return c, err
 }
 
+// OwnerFeedbackLoop starts a loop that will receive notifications about updates
+// to endpoints objects which are then used to update corresponding pods with
+// ownership info.
+// For info it only updates the services attribute. See
+// Test_PodIsAnnotatedWithOwnershipInfoImmediatelyAfterOwnersAreCreated test in
+// client_test.go for details.
+func (c *WatchClient) OwnerFeedbackLoop(stopCh <-chan struct{}) {
```
swiatekm-sumo, line 128: I wonder if we can piggyback on the infrastructure informers have for handling updates instead of reinventing it from scratch. Our owner cache is, after all, a kind of aggregate informer for multiple resource kinds.

Also, of note, is that while you don't need this for Services, if you want to do this for other owners, you need to track which pod is owned by which owner somewhere.

pmalek-sumo: I think I know what you mean but can you elaborate?

> you need to track which pod is owned by which owner somewhere.

That I know and this is what hold me back from implementing it (because of potential increased memory usage)

pkg/processor/k8sprocessor/kube/client.go
pmalek-sumo: Yes but as of now there's no problem with that because `Attributes` is only accessed in

* https://github.com/SumoLogic/sumologic-otel-collector/blob/c7750165c933ebbacd3079f0693ce4b2d7ab5951/pkg/processor/k8sprocessor/kube/client.go#L407
* and https://github.com/SumoLogic/sumologic-otel-collector/blob/2dcbc25d9907bdb219e03a18f6a584b62a1c9f32/pkg/processor/k8sprocessor/processor.go#L129

where in the former it's created and then saved in `c.Pod` under a lock and in the latter it's just retrieved to copy into the processed record.

pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -117,6 +119,51 @@ func New(
...
+						// TODO: this cannot be done this way because Attributes are
+						// being accessed elsewhere e.g.
+						// * in kubernetesprocessor.getAttributesForPod()
+						// * in tests when assertions on attributes are being made.
+						//
+						// which create data race situations.
+						p.Attributes[c.Rules.Tags.ServiceName] =
+							strings.Join(services, c.delimiter)
+						c.Pods[PodIdentifier(pod.Name)] = p
+						c.m.Unlock()
```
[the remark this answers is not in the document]
pmalek-sumo: Why? BTW: `defer` is invoked on an end of a function scope not any scope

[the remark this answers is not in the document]
swiatekm-sumo: > Why? BTW: `defer` is invoked on an end of a function scope not any scope

Goes to show, reading the language docs helps sometimes. I guess this is fine, if it panics then we don't care, and if it doesn't, then we'll aways unlock. Force of habit of wanting to wrap everything in finalizers in languages with exceptions.

pkg/processor/k8sprocessor/kube/client.go
swiatekm-sumo: So it's not actually a problem? I'm not sure what the point of this TODO is, in that case. Why not fix it right now if it's as simple as copying the Pod?

PR #386, opened 2022-01-14
fix(k8sprocessor): do not include empty service label

pmalek-sumo:
When there's no services found for a pod, do not include an empty service label.


swiatekm-sumo (approved):
We have a similar problem with ContainerId, for the record. It can be empty, but we don't currently check for it.

PR #400, opened 2022-01-31
feat(sumologicextension): use sha256 for collector credentials storage hashing

pmalek-sumo:
This idea is to still support credential files created up until today (using md5 for hashing) but use sha256 for newer installations.


sumo-drosiek:
What is the plan for credentials which are md5? We are going to migrate them or deprecate and remove support after some time?

pmalek-sumo:
> What is the plan for md5 credentials? Are we going to migrate them or deprecate and remove support after some time?

No particular plan to migrate them. My initial plan was to start using something else than md5 as default and then at some point drop support for md5 based files.

We can think of a migration in code and create an issue for that.

sumo-drosiek:
just wanted some clarification, I believe we could add some deprecation notes in this release

sumo-drosiek (approved):

PR #402, opened 2022-02-01
feat(sumologicextension): use different key for encryption and filename

pmalek-sumo:


sumo-drosiek (approved):
> What is difference between key and passphrase in hashing functions?

Nothing 😅 I was also hesitant on the naming. I can make them both `key`s if you feel that would be better

pmalek-sumo:
> What is difference between key and passphrase in hashing functions?

Nothing 😅 I was also hesitant on the naming. I can make them both `key`s if you feel that would be better

sumo-drosiek:
I would make it consistent :)
`key` is fine

pmalek-sumo:
> I would make it consistent :) `key` is fine

done

sumo-drosiek (approved):

PR #404, opened 2022-02-07
feat(sumologicextension): re-register when locally stored credentials are invalid

pmalek-sumo:


pkg/extension/sumologicextension/extension.go
```diff
@@ -148,43 +148,82 @@ func (se *SumologicExtension) validateCredenials(
...
 		zap.String(collectorNameField, colCreds.Credentials.CollectorName),
 		zap.String(collectorIdField, colCreds.Credentials.CollectorId),
 	)
 
+	go se.heartbeatLoop()
+
+	return nil
+}
+
+func (se *SumologicExtension) _r(ctx context.Context, host component.Host) (credentials.CollectorCredentials, bool, error) {
```
sumo-drosiek, line 200: what does mean `_r`?

sumo-drosiek: I guess `register`, but would change name to more informative

pmalek-sumo: Forgot to change it 😅 . PTAL now

sumo-drosiek (approved):

pkg/extension/sumologicextension/extension.go
```diff
@@ -148,43 +148,86 @@ func (se *SumologicExtension) validateCredenials(
...
+				zap.String(collectorNameField, colCreds.Credentials.CollectorName),
+				zap.String(collectorIdField, colCreds.Credentials.CollectorId),
+			)
+		} else {
+			se.logger.Info("Local collector credentials all good, starting up the collector",
+				zap.String(collectorNameField, colCreds.Credentials.CollectorName),
+				zap.String(collectorIdField, colCreds.Credentials.CollectorId),
+			)
+		}
+	}
```
swiatekm-sumo, line 186: This flow of logic is a bit confusing to me. As far as I understand it, what we want is:
1. Check if we have local credentials and if they're valid.
2. If not, register with the remote.
This code seems to call `handleCredentials` a second time if the first time fails, and it's not clear why the result should be different.

In general, I think we'd benefit from clearly separating acting on the remote from doing things locally. I'd probably split `getCredentials` into two separate functions and be explicit about the current state in this logic.

swiatekm-sumo (commented):
The code looks correct to me, but I found the structure unnecessarily confusing.

pkg/extension/sumologicextension/extension.go
pmalek-sumo: Perhaps you're right with the fact that it's a bit twisted 🤔 

The result would be different because we've deleted the credentials stored locally (line 165). Do you have an idea how you would see this?

swiatekm-sumo: At a high level, I think I'd like this to look something like:
```go
if se.conf.Clobber {
  colCreds, err := se.getRemoteCredentials()
  return colCreds, err
}
colCreds, exists := se.getLocalCredentials()
if exists && se.validateCredentials(colCreds) {
  return colCreds
}

colCreds, err := se.getRemoteCredentials()
return colCreds, err
```
just for getting the credentials. Then they can be stored locally higher up the call chain.

pmalek-sumo: Alrighty then but that looks like refactoring which would ideally land in a separate PR :)

swiatekm-sumo: Fair enough, can you add a TODO here then, if you're not going to do the refactor immediately?

pmalek-sumo: Done

swiatekm-sumo (approved):

PR #410, opened 2022-02-11
Update Comparison.md

drduke1:
Added content from "Sumo OT Distro SE considerations doc" https://docs.google.com/document/d/1F5kZjZ84EZVS-PUtXwPH7Vjdjj9QGtRlrzwQJLWRpLw/edit 

@frankreno I'm not sure how we want to handle the "Delivery Phase" section.


docs/Comparison.md
```diff
@@ -1,11 +1,93 @@
-# Comparison between the Installed Collector and the OpenTelemetry Collector
+# Comparison between the Installed Collector and OpenTelemetry Collector
+
+## Delivery Phase 1: “Initial Release”
+
+_Expected March 2022_
```
frankreno, line 5: Let's remove the phases/dates.  The main reason is this is internal. All other items in the doc are valid.  As we enhance OT and add support for these we can update this.

frankreno (approved):
One minor update LGTM to otherwise

PR #417, opened 2022-02-21
refactor(sumologicextension): refactor registration and credentials storage calls

pmalek-sumo:
Related to discussion in: https://github.com/SumoLogic/sumologic-otel-collector/pull/404#discussion_r801525739

### DIFF
--- a/pkg/extension/sumologicextension/extension.go
+++ b/pkg/extension/sumologicextension/extension.go
@@ -34,6 +34,7 @@ import (
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config"
 	"go.opentelemetry.io/collector/config/configauth"
+	"go.opentelemetry.io/collector/config/confighttp"
 	"go.uber.org/zap"
 	grpccredentials "google.golang.org/grpc/credentials"
 
@@ -89,7 +90,6 @@ func newSumologicExtension(conf *Config, logger *zap.Logger) (*SumologicExtensio
 		return nil, err
 	}
 
-	var collectorName string
 	credentialsStore, err := credentials.NewLocalFsStore(
 		credentials.WithCredentialsDirectory(conf.CollectorCredentialsDirectory),
 		credentials.WithLogger(logger),
@@ -98,16 +98,23 @@ func newSumologicExtension(conf *Config, logger *zap.Logger) (*SumologicExtensio
 		return nil, fmt.Errorf("failed to initialize credentials store: %w", err)
 	}
 
+	var (
+		collectorName string
+		hashKey       = createHashKey(conf)
+	)
 	if conf.CollectorName == "" {
-		key := createHashKey(conf)
-		// If collector name is not set by the user check if the collector was restarted
-		if !credentialsStore.Check(key) {
+		// If collector name is not set by the user, check if the collector was restarted
+		// and that we can reuse collector name save in credentials store.
+		if creds, err := credentialsStore.Get(hashKey); err != nil {
 			// If credentials file is not stored on filesystem generate collector name
 			collectorName = fmt.Sprintf("%s-%s", hostname, uuid.New())
+		} else {
+			collectorName = creds.CollectorName
 		}
 	} else {
 		collectorName = conf.CollectorName
 	}
+
 	if conf.HeartBeatInterval <= 0 {
 		conf.HeartBeatInterval = DefaultHeartbeatInterval
 	}
@@ -123,7 +130,7 @@ func newSumologicExtension(conf *Config, logger *zap.Logger) (*SumologicExtensio
 		baseUrl:          strings.TrimSuffix(conf.ApiBaseUrl, "/"),
 		conf:             conf,
 		logger:           logger,
-		hashKey:          createHashKey(conf),
+		hashKey:          hashKey,
 		credentialsStore: credentialsStore,
 		closeChan:        make(chan struct{}),
 		backOff:          backOff,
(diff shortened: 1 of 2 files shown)


pkg/extension/sumologicextension/extension.go
```diff
@@ -200,104 +168,146 @@ func (se *SumologicExtension) Start(ctx context.Context, host component.Host) er
...
+// getLocalCredentials returns the credentials either retrieved from local credentials
+// storage in cases they were available there or it registers the collector and
+// returns the credentials obtained from the API.
+// It also returns a boolean flag indicating if the successful registration
+// with the API took place and an error.
+func (se *SumologicExtension) getLocalCredentials(ctx context.Context) (credentials.CollectorCredentials, error) {
+	if !se.credentialsStore.Check(se.hashKey) {
+		return credentials.CollectorCredentials{},
+			fmt.Errorf("collector credentials not found (hash key: %s)", se.hashKey)
+	}
```
swiatekm-sumo, line 298: Is there anything this check gives us that the `.Get` below doesn't?

pkg/receiver/telegrafreceiver/converter.go
```diff
@@ -42,6 +42,10 @@ func newConverter(separateField bool, logger *zap.Logger) MetricConverter {
 	}
 }
 
+var _filter = map[string]struct{}{
+	"http_listener_v2_path": {},
+}
+
 // Convert converts telegraf.Metric to pdata.Metrics.
```
swiatekm-sumo, line 49: Why are these changes here?

swiatekm-sumo (changes requested):

pkg/extension/sumologicextension/extension.go
```diff
@@ -200,104 +168,146 @@ func (se *SumologicExtension) Start(ctx context.Context, host component.Host) er
...
 
+func (se *SumologicExtension) validateCredentials(
+	ctx context.Context,
+) error {
+	se.logger.Info("Validating collector credentials...",
+		zap.String(collectorIdField, se.registrationInfo.CollectorCredentialId),
+		zap.String(collectorNameField, se.registrationInfo.CollectorName),
+	)
+
+	return se.sendHeartbeat(ctx)
```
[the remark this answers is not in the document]
pmalek-sumo: Well we could but this really checks how the credentials were injected into the http client's transport 🤔 So however we do this it will be implicit anyway (so some extent).

I'm open to suggestions.

pkg/extension/sumologicextension/extension.go
```diff
@@ -200,104 +168,146 @@ func (se *SumologicExtension) Start(ctx context.Context, host component.Host) er
...
-	if se.credentialsStore.Check(se.hashKey) {
-		colCreds, err = se.credentialsStore.Get(se.hashKey)
-		if err != nil {
-			return credentials.CollectorCredentials{}, false, err
-		}
-		se.collectorName = colCreds.CollectorName
-		if !se.conf.Clobber {
-			if colCreds.ApiBaseUrl != "" {
-				se.baseUrl = colCreds.ApiBaseUrl
+	if !se.conf.Clobber {
```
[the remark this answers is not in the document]
pmalek-sumo: this is going away anyway as we'll be adding a separate flag for re-registration

pkg/extension/sumologicextension/extension.go
```diff
@@ -200,104 +168,146 @@ func (se *SumologicExtension) Start(ctx context.Context, host component.Host) er
...
-	return colCreds, registrationDone, err
+	colCreds, err = se.getCredentialsByRegistering(ctx)
+	if err != nil {
+		return credentials.CollectorCredentials{}, err
+	}
+
+	return colCreds, nil
+}
+
+func (se *SumologicExtension) getCredentialsByRegistering(ctx context.Context) (credentials.CollectorCredentials, error) {
```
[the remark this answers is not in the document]
pmalek-sumo: Seems consistent to `getLocalCredentials` though 🤔

pkg/receiver/telegrafreceiver/converter.go
pmalek-sumo: Derp, will fix this

pkg/extension/sumologicextension/extension.go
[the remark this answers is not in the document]
swiatekm-sumo: I guess it's fine, it captures the meaning well enough.

pkg/extension/sumologicextension/extension.go
pmalek-sumo: An explicit check only. Changed to `.Get`

pkg/extension/sumologicextension/extension.go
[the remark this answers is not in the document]
swiatekm-sumo: I think what it *should* check is whether the generated transport lets us do a valid request. Something like:
```go
httpClient := se.getHttpClient(creds)
se.sendHeartbeatWithHttpClient(httpClient)
```
and use these functions for normal heartbeats as well.

pkg/extension/sumologicextension/extension.go
```diff
@@ -200,104 +169,166 @@ func (se *SumologicExtension) Start(ctx context.Context, host component.Host) er
...
-		se.collectorName = colCreds.CollectorName
-		if !se.conf.Clobber {
-			if colCreds.ApiBaseUrl != "" {
-				se.baseUrl = colCreds.ApiBaseUrl
+	if !se.conf.Clobber {
+		colCreds, err = se.getLocalCredentials(ctx)
+		if err == nil {
+			if errI := se.injectCredentials(colCreds, host); errI != nil {
+				return credentials.CollectorCredentials{}, errI
+			}
```
swiatekm-sumo, line 258: With the changes to `.validateCredentials`, can we now move this call to after the credentials were validated?

pmalek-sumo: What we could do is we could remove `injectCredentials` call and move it to `validateCredentials` as we'll need to set the `.registrationInfo` field anyway.

pmalek-sumo: I've moved stuff around PTAL @swiatekm-sumo

pkg/extension/sumologicextension/extension.go
```diff
@@ -200,104 +169,161 @@ func (se *SumologicExtension) Start(ctx context.Context, host component.Host) er
...
+	host component.Host,
+	colCreds credentials.CollectorCredentials,
+) error {
+	se.logger.Info("Validating collector credentials...",
+		zap.String(collectorIdField, se.registrationInfo.CollectorCredentialId),
+		zap.String(collectorNameField, se.registrationInfo.CollectorName),
+	)
+
+	if err := se.injectCredentials(colCreds, host); err != nil {
+		return err
```
swiatekm-sumo, line 194: Do we actually need this here? It doesn't look like it does anything other than setting the client, which we're explicitly using for the heartbeat anyway.

pkg/extension/sumologicextension/extension.go
```diff
@@ -200,104 +169,161 @@ func (se *SumologicExtension) Start(ctx context.Context, host component.Host) er
...
+}
+
+func (se *SumologicExtension) validateCredentials(
+	ctx context.Context,
+	host component.Host,
+	colCreds credentials.CollectorCredentials,
+) error {
+	se.logger.Info("Validating collector credentials...",
+		zap.String(collectorIdField, se.registrationInfo.CollectorCredentialId),
+		zap.String(collectorNameField, se.registrationInfo.CollectorName),
```
swiatekm-sumo, line 190: I think these should use `colCreds` instead.

pkg/extension/sumologicextension/extension.go
pmalek-sumo: It is needed as it sets the credentials in `registrationInfo` and the client (as you mentioned).

pkg/extension/sumologicextension/extension.go
pmalek-sumo: changed

swiatekm-sumo (approved):

PR #420, opened 2022-02-22
feat(sumologicextension): re-register collector on HTTP Unauthorized

pmalek-sumo:
This PR addresses the problem of Sumo API returning 401 during collector's runtime in which case data stops being sent due to collector not being able to get authorized.

With this PR the collector will detect that (via 401 returned on heartbeat request), trigger re-registration and then `sumologicexporter` will reconfigure as well if it detects 401 in response to sending data.

Exemplar log from re-registration process:

```
2022-02-22T14:42:25.378+0100    debug   sumologicexporter@v0.44.0/sender.go:156 Sending data    {"kind": "exporter", "name": "sumologic/1", "pipeline": "metrics", "headers": {"Content-Encoding":["gzip"],"Content-Type":["application/x-protobuf"],"X-Sumo-Client":["otelcol"],"X-Sumo-Host":["pmalek_source_host"],"X-Sumo-Name":["pmalek_source_name"]}}
2022-02-22T14:42:25.565+0100    debug   sumologicextension@v0.44.0/extension.go:540     Heartbeat sent  {"kind": "extension", "name": "sumologic", "collector_name": "pmalek_my_custom_collector_reg_030", "collector_id": "000000000123123C"}
2022-02-22T14:42:30.371+0100    debug   sumologicexporter@v0.44.0/sender.go:156 Sending data    {"kind": "exporter", "name": "sumologic/1", "pipeline": "metrics", "headers": {"Content-Encoding":["gzip"],"Content-Type":["application/x-protobuf"],"X-Sumo-Client":["otelcol"],"X-Sumo-Host":["pmalek_source_host"],"X-Sumo-Name":["pmalek_source_name"]}}
2022-02-22T14:42:30.569+0100    warn    sumologicextension@v0.44.0/extension.go:517     Heartbeat request unauthorized, re-registering the collector    {"kind": "extension", "name": "sumologic", "collector_name": "pmalek_my_custom_collector_reg_030", "collector_id": "000000000123123C"}
2022-02-22T14:42:30.570+0100    info    sumologicextension@v0.44.0/extension.go:380     Calling register API    {"kind": "extension", "name": "sumologic", "collector_name": "pmalek_my_custom_collector_reg_030", "collector_id": "000000000123123C", "URL": "https://sumologic.com/api/v1/collector/register"}
2022-02-22T14:42:30.913+0100    info    sumologicextension@v0.44.0/extension.go:465     Collector registration finished successfully    {"kind": "extension", "name": "sumologic", "collector_name": "pmalek_my_custom_collector_reg_030", "collector_id": "0000000001231232"}
2022-02-22T14:42:30.914+0100    info    credentials/credentialsstore_localfs.go:199     Collector registration credentials stored locally       {"kind": "extension", "name": "sumologic", "path": "/Users/pmalek/.sumologic-otel-collector/11111111111111111111111111111111111111111111111111111111111111111"}
2022-02-22T14:42:35.372+0100    debug   sumologicexporter@v0.44.0/sender.go:156 Sending data    {"kind": "exporter", "name": "sumologic/1", "pipeline": "metrics", "headers": {"Content-Encoding":["gzip"],"Content-Type":["application/x-protobuf"],"X-Sumo-Client":["otelcol"],"X-Sumo-Host":["pmalek_source_host"],"X-Sumo-Name":["pmalek_source_name"]}}
```

### DIFF
--- a/pkg/exporter/sumologicexporter/exporter.go
+++ b/pkg/exporter/sumologicexporter/exporter.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"sync"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer/consumererror"
@@ -39,7 +40,9 @@ const (
 
 type sumologicexporter struct {
 	sources             sourceFormats
+	configLock          sync.RWMutex
 	config              *Config
+	host                component.Host
 	logger              *zap.Logger
 	client              *http.Client
 	filter              filter
@@ -257,6 +260,8 @@ func (se *sumologicexporter) pushLogsData(ctx context.Context, ld pdata.Logs) er
 	if err != nil {
 		return consumererror.NewLogs(fmt.Errorf("failed to initialize compressor: %w", err), ld)
 	}
+
+	se.configLock.RLock()
 	sdr := newSender(
 		se.logger,
 		se.config,
@@ -343,6 +348,7 @@ func (se *sumologicexporter) pushLogsData(ctx context.Context, ld pdata.Logs) er
 		errs = append(errs, err)
 	}
 
+	se.configLock.RUnlock()
 	if len(droppedRecords) > 0 {
 		// Move all dropped records to Logs
 		droppedLogs := pdata.NewLogs()
@@ -355,6 +361,7 @@ func (se *sumologicexporter) pushLogsData(ctx context.Context, ld pdata.Logs) er
 			lp.log.CopyTo(logs.AppendEmpty())
 		}
 
+		se.handleUnauthorizedErrors(ctx, errs...)
 		return consumererror.NewLogs(multierr.Combine(errs...), droppedLogs)
 	}
 
@@ -377,6 +384,8 @@ func (se *sumologicexporter) pushMetricsData(ctx context.Context, md pdata.Metri
 	if err != nil {
 		return consumererror.NewMetrics(fmt.Errorf("failed to initialize compressor: %w", err), md)
 	}
+
+	se.configLock.RLock()
 	sdr := newSender(
 		se.logger,
 		se.config,
@@ -454,6 +463,7 @@ func (se *sumologicexporter) pushMetricsData(ctx context.Context, md pdata.Metri
 		errs = append(errs, err)
 	}
 
+	se.configLock.RUnlock()
 	if len(droppedRecords) > 0 {
 		// Move all dropped records to Metrics
 		droppedMetrics := pdata.NewMetrics()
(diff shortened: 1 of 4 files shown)


swiatekm-sumo (commented):
Two questions:
1. Can we scope the locking in the exporter in such a way that it only happens in a single function? Something like `.getClient`? Right now it's all over the place and it's difficult to know what exactly the lock is protecting.
2. Do we need the lock in the extension? At a glance, it seems like the config would only be modified from the heartbeat loop, which is serial.

pmalek-sumo:
> Two questions:
> 
> 1. Can we scope the locking in the exporter in such a way that it only happens in a single function? Something like `.getClient`? Right now it's all over the place and it's difficult to know what exactly the lock is protecting.
> 2. Do we need the lock in the extension? At a glance, it seems like the config would only be modified from the heartbeat loop, which is serial.

ad 1. Changed this to lock only data URLs and client with 2 separate locks for this. It doesn't look too pretty but OTOH we're aiming to get rid of these sender functions and jump on OTLP wagon at some point.

ad 2. That might actually be true. Despite not being a big deal I've removed those locks. The subtle potential problem with this is that the code  is brittle with respect to potential future changes

swiatekm-sumo:
> ad 1. Changed this to lock only data URLs and client with 2 separate locks for this. It doesn't look too pretty but OTOH we're aiming to get rid of these sender functions and jump on OTLP wagon at some point.

I think this could be the same lock without any impact on performance, but this is good as well. Ugly and explicit is better than pretty and hidden, and all that.

I do think that you also need to protect reading data from the extension with a lock, as it can be accessed concurrently. I think protecting the entirety of `configure` would be a good choice here, 

> ad 2. That might actually be true. Despite not being a big deal I've removed those locks. The subtle potential problem with this is that the code is brittle with respect to potential future changes

Eh, I think it's fine. The extension is essentially serial - the heartbeat loop is the only thing that runs  after it's started. If someone wanted to add more concurrency to it, they should be aware that they need to watch out for race conditions.

Other than the point above about protecting access to the extension from the exporter, this all looks good to me.

pmalek-sumo:
> I do think that you also need to protect reading data from the extension with a lock, as it can be accessed concurrently. Protecting the entirety of `configure` would be a good choice here.

I've added a lock around `BaseUrl` which is the only thing accessed in configure still without a lock. PTAL @swiatekm-sumo

pkg/exporter/sumologicexporter/exporter.go
```diff
@@ -38,16 +39,22 @@ const (
...
-	sources             sourceFormats
-	config              *Config
-	logger              *zap.Logger
-	client              *http.Client
+	sources sourceFormats
+	config  *Config
+	host    component.Host
+	logger  *zap.Logger
+
+	clientLock sync.RWMutex
```
swiatekm-sumo, line 47: Can we add a comment here and for `dataUrlsLock`, and maybe also in the get/set functions, explaining why this is necessary?

swiatekm-sumo (approved):
Looks good to me. I would like more comments on the locks and methods, but the logic is sound.

pkg/exporter/sumologicexporter/exporter.go
pmalek-sumo: Added a comment around dataUrlsLock and baseUrlLock in the extension

PR #422, opened 2022-02-23
docs(troubleshooting): add section on full queue

astencel-sumo:


docs/Troubleshooting.md
```diff
@@ -0,0 +1,26 @@
...
+increasing the queue size for the exporter with `sending_queue.queue_size` property
+might be enough to accommodate temporary additional load.
+
+If you can see `429 Too Many Requests` HTTP response codes from Sumo in exporter logs,
+this means the exporter is being throttled by Sumo backend.
+In this case, you need to either decrease the volume of data sent,
+or reach out to Sumo support to increase the quota.
+
+If the exporter is not being throttled the best option might be
+to increase the number of consumers sending data from queue to Sumo with `sending_queue.num_consumers`.
```
sumo-drosiek, line 26: also playing with batching, request size and timeout properties

astencel-sumo: Added paragraph about request size https://github.com/SumoLogic/sumologic-otel-collector/pull/422/commits/246b963a3a240d67f716d29f8dd6c0aa6fc8d79a.

docs/Troubleshooting.md
```diff
@@ -0,0 +1,31 @@
...
+```
+
+This means that the `sumologicexporter` is not able to send data as quickly as it receives new data.
+There may be a couple ways to fix this is, depending on the root cause.
+
+If the problem is intermittent and caused by temporary spike in data volume,
+increasing the queue size for the exporter with `sending_queue.queue_size` property
+might be enough to accommodate temporary additional load.
+
+If you can see `429 Too Many Requests` HTTP response codes from Sumo in exporter logs,
```
pmalek-sumo, line 20:
```suggestion
If you see `429 Too Many Requests` HTTP response codes from Sumo in exporter logs,
```

pmalek-sumo (approved):

PR #427, opened 2022-02-24
feat(sumologicexporter): add better error handling for errors from Sumo API

pmalek-sumo:


sumo-drosiek (approved):

pkg/exporter/sumologicexporter/sender.go
```diff
@@ -164,47 +164,61 @@ func (s *sender) send(ctx context.Context, pipeline PipelineType, body io.Reader
...
+		type ReceiverResponse struct {
+			Status  int    `json:"status,omitempty"`
+			ID      string `json:"id,omitempty"`
+			Code    string `json:"code,omitempty"`
+			Message string `json:"message,omitempty"`
+			Errors  []struct {
+				Code    string `json:"code"`
+				Message string `json:"message"`
+			} `json:"errors,omitempty"`
+		}
```
swiatekm-sumo, line 188: Do you know if these fields are actually allowed to be empty? We could make the parsing and error-generation logic simpler if we knew they're always available.

pmalek-sumo: They are not. It depends on the error being returned.

swiatekm-sumo (approved):

PR #439, opened 2022-03-02
fix(sumologicexporter): fix batching logs from different files

astencel-sumo:
The `fields.string()` method is not good for comparisons because it ignores the `_sourceName` attribute.

Refs: SUMO-187202

### DIFF
--- a/pkg/exporter/sumologicexporter/exporter.go
+++ b/pkg/exporter/sumologicexporter/exporter.go
@@ -282,7 +282,7 @@ func (se *sumologicexporter) pushLogsData(ctx context.Context, ld pdata.Logs) er
 				}
 
 				// If metadata differs from currently buffered, flush the buffer
-				if currentMetadata.string() != previousMetadata.string() && previousMetadata.string() != "" {
+				if !currentMetadata.equals(previousMetadata) && !previousMetadata.isEmpty() {
 					var dropped []logPair
 					dropped, err = sdr.sendLogs(ctx, previousMetadata)
 					if err != nil {
@@ -397,7 +397,7 @@ func (se *sumologicexporter) pushMetricsData(ctx context.Context, md pdata.Metri
 				}
 
 				// If metadata differs from currently buffered, flush the buffer
-				if currentMetadata.string() != previousMetadata.string() && previousMetadata.string() != "" {
+				if !currentMetadata.equals(previousMetadata) && !previousMetadata.isEmpty() {
 					var dropped []metricPair
 					dropped, err = sdr.sendMetrics(ctx, previousMetadata)
 					if err != nil {
--- a/pkg/exporter/sumologicexporter/exporter_test.go
+++ b/pkg/exporter/sumologicexporter/exporter_test.go
@@ -93,7 +93,7 @@ func prepareExporterTest(t *testing.T, cfg *Config, cb []func(w http.ResponseWri
 			return
 		}
 
-		if c := int(atomic.LoadInt32(&reqCounter)); assert.Greater(t, len(cb), c) {
+		if c := int(atomic.LoadInt32(&reqCounter)); assert.Greater(t, len(cb), c, "Exporter sent more requests than the number of test callbacks defined:", len(cb)) {
 			cb[c](w, req)
 			atomic.AddInt32(&reqCounter, 1)
 		}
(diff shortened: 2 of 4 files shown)


pkg/exporter/sumologicexporter/exporter_test.go
```diff
@@ -541,11 +541,59 @@ func TestPushJSONLogs_AttributeTranslation(t *testing.T) {
...
+					// Source attributes like `_sourceName` are not included in fields
+					assert.Equal(t, "", req.Header.Get("X-Sumo-Fields"))
+					assert.Equal(t, "/file1.log", req.Header.Get("X-Sumo-Name"))
+				},
+				func(w http.ResponseWriter, req *http.Request) {
+					body := extractBody(t, req)
+
+					var regex string
+					regex += `^{"log":"log from file2","timestamp":\d{13}}$`
+					assert.Regexp(t, regex, body)
```
pmalek-sumo, line 586:
```suggestion
					assert.Regexp(t, `^{"log":"log from file2","timestamp":\d{13}}$`, body)
```

pkg/exporter/sumologicexporter/exporter.go
```diff
@@ -282,7 +282,9 @@ func (se *sumologicexporter) pushLogsData(ctx context.Context, ld pdata.Logs) er
 				}
 
 				// If metadata differs from currently buffered, flush the buffer
-				if currentMetadata.string() != previousMetadata.string() && previousMetadata.string() != "" {
+				currentMetadataString := fmt.Sprintf("%v", currentMetadata.orig.AsRaw())
+				previousMetadataString := fmt.Sprintf("%v", previousMetadata.orig.AsRaw())
```
sumo-drosiek, line 286: in this form, please add metadata function for that

astencel-sumo: I actually found another way to compare the objects, using `reflect.DeepEqual()`.

pkg/exporter/sumologicexporter/exporter_test.go
```diff
@@ -541,11 +541,55 @@ func TestPushJSONLogs_AttributeTranslation(t *testing.T) {
 
 	testcases := []struct {
 		name       string
+		logs       pdata.Logs
 		configFunc func() *Config
 		callbacks  []func(w http.ResponseWriter, req *http.Request)
 	}{
+		{
+			name: "logs from different files",
```
pmalek-sumo, line 549: This sounds like an incomplete sentence to me :) Logs from different files ... ?

pkg/exporter/sumologicexporter/exporter.go
sumo-drosiek: still, could we extract it to function? :sweat_smile:

pkg/exporter/sumologicexporter/exporter_test.go
astencel-sumo: Changed it to `logs from different files are sent with different metadata`

pkg/exporter/sumologicexporter/exporter.go
```diff
@@ -282,7 +282,7 @@ func (se *sumologicexporter) pushLogsData(ctx context.Context, ld pdata.Logs) er
 				}
 
 				// If metadata differs from currently buffered, flush the buffer
-				if currentMetadata.string() != previousMetadata.string() && previousMetadata.string() != "" {
+				if !currentMetadata.equals(previousMetadata) && !previousMetadata.isEmpty() {
```
sumo-drosiek, line 285:
```suggestion
				if currentMetadata.equals(previousMetadata) || previousMetadata.isEmpty() {
```

swiatekm-sumo:
@astencel-sumo have you tried running a simple benchmark for how fast this check is with different implementations? I vaguely recall that `reflect.DeepEqual` is slow, and we do run this for every datapoint, so it adds up.

astencel-sumo:
> @astencel-sumo have you tried running a simple benchmark for how fast this check is with different implementations? I vaguely recall that `reflect.DeepEqual` is slow, and we do run this for every datapoint, so it adds up.

The previous approach to do dump the struct to a string, which probably isn't a speed champion either. Having said that, no I haven't run any benchmarks.

pkg/exporter/sumologicexporter/exporter.go
```diff
@@ -397,7 +397,7 @@ func (se *sumologicexporter) pushMetricsData(ctx context.Context, md pdata.Metri
...
```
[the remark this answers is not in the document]
astencel-sumo: This wouldn't work. You cannot use `continue` like this. I understand the condition would be easier to read if it was positive instead of negative, but I don't see a good way to do this.

pmalek-sumo (approved):

pkg/exporter/sumologicexporter/exporter.go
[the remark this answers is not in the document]
sumo-drosiek: do not trust github :confused: 
image

sumo-drosiek (approved):

PR #443, opened 2022-03-03
fix(sumologicexporter): better receiver error handling

pmalek-sumo:
Exemplar errors to handle:

```
{
  "status" : 200,
  "id" : "YBLR1-11111-11111",
  "code" : "bad.http.header.fields",
  "message" : "X-Sumo-Fields Warning: 1 key-value pairs are dropped as their values are exceeding max value length 200. 14 key-value pairs are dropped as they are exceeding maximum key-value pair number limit 30."
}
```

Exemplar log entry that this PR introduces:

```
There was an issue sending data	{"status": "200 OK", "id": "YBLR1-11111-11111", "code": "bad.http.header.fields", "message": "X-Sumo-Fields Warning: 1 key-value pairs are dropped as their values are exceeding max value length 200. 14 key-value pairs are dropped as they are exceeding maximum key-value pair number limit 30."}
```


pkg/exporter/sumologicexporter/sender.go
```diff
@@ -168,26 +168,56 @@ func (s *sender) send(ctx context.Context, pipeline PipelineType, body io.Reader
...
+		if len(rResponse.ID) > 0 {
+			l = l.With(zap.String("id", rResponse.ID))
+		}
+		if len(rResponse.Code) > 0 {
+			l = l.With(zap.String("code", rResponse.Code))
+		}
+		if len(rResponse.Message) > 0 {
+			l = l.With(zap.String("message", rResponse.Message))
+		}
+		l.Warn("There was an issue sending data")
```
swiatekm-sumo, line 205: Aren't we going to always log this, even if we successfully sent the data?

pmalek-sumo: Nope. The contract with the API is that we get a content length of 0 and a 200 or 204 status code if the data was received successfully.

If something was off then the body will contain a json with a message.

swiatekm-sumo: Can you add a comment explaining this?

pmalek-sumo: Please check now

swiatekm-sumo (approved):

PR #446, opened 2022-03-04
fix(sumologicexporter): deduplicate error messages

astencel-sumo:
Refs: SUMO-185930


pkg/exporter/sumologicexporter/exporter.go
```diff
@@ -440,13 +441,32 @@ func (se *sumologicexporter) pushMetricsData(ctx context.Context, md pdata.Metri
...
 
+// removeDuplicates removes duplicate errors from a collection of errors.
+// Duplicate errors are errors that have the same message returned by `error.Error()`.
+func removeDuplicates(errs []error) []error {
+	if len(errs) < 2 {
+		return errs
+	}
+
+	var uniqueList []error
+	uniqueSet := make(map[string]bool)
```
pmalek-sumo, line 460:
```suggestion
	uniqueSet := make(map[string]struct{})
```

and then use not the value but the second value returned by the map check

pmalek-sumo (commented):
In general I'm sort of against this change (but I can understand its benefits like cleaner log messages) because it hides away the scale of the problem i.e. the number of errors that occurred.

This way it may be harder to correlate requests sent to errors received.

sumo-drosiek:
@pmalek-sumo so maybe we should add counter to the errors, so we won't loose this information

swiatekm-sumo:
> @pmalek-sumo so maybe we should add counter to the errors, so we won't loose this information

I like this idea. We can use a format like `{error_message} (x{error_count})`, which I think is both clear and compact.

pkg/exporter/sumologicexporter/exporter.go
astencel-sumo: Done

astencel-sumo:
> > @pmalek-sumo so maybe we should add counter to the errors, so we won't loose this information
> 
> I like this idea. We can use a format like `{error_message} (x{error_count})`, which I think is both clear and compact.

I agree, will implement this.

astencel-sumo:
Implemented counting errors in https://github.com/SumoLogic/sumologic-otel-collector/pull/446/commits/704dc75d507e29a1555c4b64782d1acf85c791c9, please review @pmalek-sumo @sumo-drosiek @swiatekm-sumo

pkg/exporter/sumologicexporter/deduplicate_errors.go
```diff
@@ -0,0 +1,60 @@
...
+	errorsWithCounts := []errorWithCount{}
+	for _, err := range errs {
+		found := false
+		for i := range errorsWithCounts {
+			if errorsWithCounts[i].err.Error() == err.Error() {
+				found = true
+				errorsWithCounts[i].count += 1
+				break
+			}
+		}
```
sumo-drosiek, line 37: wouldn't be better to use map instead of list?

astencel-sumo: I want to keep the order of the errors. With a map, I'm losing the ordering.

pmalek-sumo (approved):

PR #449, opened 2022-03-04
fix(telegrafreceiver): shutdown Telegraf agent gracefully

swiatekm-sumo:
It was possible to create panics in Telegraf agent by closing the channel before the input plugin is done closing. We fix this in two ways:
1. let Telegraf agent close the channel on its own, and wait for it to do so before shutting down the receiver
2. upgrade to a new Telegraf version where the http plugin closes more gracefully, see the change here: https://github.com/SumoLogic/telegraf/pull/36

We also add a simple test for receiver start and shutdown.

### DIFF
--- a/pkg/receiver/telegrafreceiver/go.mod
+++ b/pkg/receiver/telegrafreceiver/go.mod
@@ -276,4 +276,4 @@ require (
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
 
-replace github.com/influxdata/telegraf => github.com/sumologic/telegraf v1.21.3-sumo-0
+replace github.com/influxdata/telegraf => github.com/sumologic/telegraf v1.21.3-sumo-1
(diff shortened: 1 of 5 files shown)


pkg/receiver/telegrafreceiver/go.mod
```diff
@@ -276,4 +276,4 @@ require (
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
 
-replace github.com/influxdata/telegraf => github.com/sumologic/telegraf v1.21.3-sumo-0
+replace github.com/influxdata/telegraf => github.com/sumologic/telegraf v1.21.3-sumo-1
```
pmalek-sumo, line 279: Add this change also in builder config

swiatekm-sumo: Done.

pkg/receiver/telegrafreceiver/receiver_test.go
```diff
@@ -0,0 +1,44 @@
+// Copyright 2021, OpenTelemetry Authors
```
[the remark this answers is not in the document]
swiatekm-sumo: We can just remove it, I don't think it's necessary.

pmalek-sumo (approved):

pmalek-sumo (approved):

PR #454, opened 2022-03-08
Update docs and examples to use the most recent otelcol version

perk-sumo:


pmalek-sumo (approved):
NOTE: that 0.0.53-beta.0 is not yet ready :)

perk-sumo:
Good call! Changed that into `0.0.52-beta.0`.

swiatekm-sumo:
Should we automate this in some way? Or maybe just replace all of these concrete values with a placeholder like `<version>`?

pmalek-sumo:
> Should we automate this in some way? Or maybe just replace all of these concrete values with a placeholder like `<version>`?

No point I think. When this gets out of beta we'll use releases instead of prereleases and then https://github.com/SumoLogic/sumologic-otel-collector/releases/latest will work.

PR #467, opened 2022-03-10
docs(configuration): add example for keeping prometheus format with OTLP

sumo-drosiek:
This solution is not perfect but there is not so many alternatives

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -13,6 +13,7 @@
   - [Parsing JSON logs](#parsing-json-logs)
 - [Command-line configuration options](#command-line-configuration-options)
 - [Proxy Support](#proxy-support)
+- [Keeping Prometheus format using OTLP exporter](#keeping-prometheus-format-using-otlp-exporter)
 
 ---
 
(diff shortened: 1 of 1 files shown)


pmalek-sumo (commented):
Have you tested this with jmx metrics which have dots in field names (which are translated in telegrafreceiver to OT's data point labels)?

Wouldn't this only update the metrics names and not the data point labels?

sumo-drosiek:
@pmalek-sumo Yes, and then we have `field` dimension  (with dot of course)

pmalek-sumo:
> @pmalek-sumo Yes, and then we have `field` dimension (with dot of course)

Shouldn't we document that separately then?

sumo-drosiek:
Not sure as app dashboards do not rely on them and also I don't see possibility to change field value using regexp

pmalek-sumo:
> Not sure as app dashboards do not rely on them and also I don't see possibility to change field value using regexp

Now I'm confused 😕  Wasn't the problem about fields like e.g. `Utilization.min`? Isn't that what we're trying to cover here?

sumo-drosiek:
the problem is with field indeed, but in case this field has been added to metric name so we had:

`metric_name_Utilization.max`, but we would like to have `metric_name_Utilization_max`

so, field in telegraf but metric name in OTC

pmalek-sumo:
> the problem is with field indeed, but in case this field has been added to metric name so we had:
> 
> `metric_name_Utilization.max`, but we would like to have `metric_name_Utilization_max`
> 
> so, field in telegraf but metric name in OTC

What if someone uses `separate_field: true` in [telegrafreceiver](https://github.com/SumoLogic/sumologic-otel-collector/tree/main/pkg/receiver/telegrafreceiver)? Then the field becomes a data point label

sumo-drosiek:
> What if someone uses separate_field: true in [telegrafreceiver](https://github.com/SumoLogic/sumologic-otel-collector/tree/main/pkg/receiver/telegrafreceiver)? Then the field becomes a data point label

yes, but we don't use scenario with `separate_field: true` for application dashboards, also dot is fine as label value in prometheus format

pmalek-sumo (approved):

PR #472, opened 2022-03-14
ci: run tracing tests on demand

swiatekm-sumo:
These are legacy tracing tests that we'd like to run on our main branch until we replace them with something less bespoke. They use CircleCI and some internal test tooling.


.circleci/config.yml
```diff
@@ -0,0 +1,100 @@
...
+          command: echo "export KUBERNETES_METADATA_URL=$(kubectl -n java-app get svc k8s-feeder-svc-pub -o json | jq .status.loadBalancer.ingress[0].hostname)" >> $BASH_ENV
+      - run:
+          name: "Deploy Java App application"
+          command: /opt/tracing-tests/deployment_scripts/deploy-test-applications.sh
+      - run:
+          name: "Wait for Kubernetes Metadata Provider"
+          command: kubectl -n java-app wait --for=condition=ready --timeout=60s pod -l app=server
+      - run:
+          name: "Wait for data..."
+          command: sleep 180
```
pmalek-sumo, line 93: Isn't this brittle?

pmalek-sumo (approved):
Can we add a screenshot to the description on how to run those? Or ideally a dev doc section describing that?

.circleci/config.yml
swiatekm-sumo: It is, but it's a lot more complex to properly wait for this data, and I don't think we want to invest more time in these tests currently. As they're only run on demand, you can always rerun if you get a random failure.

pmalek-sumo (commented):
👍

PR #489, opened 2022-03-25
feat(exporter/sumologic): drop attribute used by routing processor

sumo-drosiek:
Drop metrics attribute used for routing

### DIFF
--- a/pkg/exporter/sumologicexporter/README.md
+++ b/pkg/exporter/sumologicexporter/README.md
@@ -58,7 +58,7 @@ exporters:
     # source templating is going to be applied,
     # default = `%{_metric_}`
     graphite_template: <graphite_template>
-    # name of resource attribute which should be dropped for metrics
+    # name of resource attribute which should be dropped for records
     # this is for attribute used by resource processor
     # other attributes should be removed by processors in pipelines before
     # This is workaround for the following issue:
--- a/pkg/exporter/sumologicexporter/README.md
+++ b/pkg/exporter/sumologicexporter/README.md
@@ -59,7 +59,7 @@ exporters:
     # default = `%{_metric_}`
     graphite_template: <graphite_template>
     # name of resource attribute which should be dropped for records
-    # this is for attribute used by resource processor
+    # this is for attribute used by routing processor
     # other attributes should be removed by processors in pipelines before
     # This is workaround for the following issue:
     # https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7407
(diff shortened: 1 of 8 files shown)


sumo-drosiek:
I'm going to add it for logs and traces, even if we didn't have that case for them yet

swiatekm-sumo (commented):
Can you also add a test for traces?

pkg/exporter/sumologicexporter/README.md
```diff
@@ -58,6 +58,13 @@ exporters:
     # source templating is going to be applied,
     # default = `%{_metric_}`
     graphite_template: <graphite_template>
+    # name of resource attribute which should be dropped for records
+    # this is for attribute used by routing processor
+    # other attributes should be removed by processors in pipelines before
+    # This is workaround for the following issue:
+    # https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7407
+    # default = `sumo_routing`
+    drop_routing_atttribute: <drop_routing_atttribute>
```
pmalek-sumo, line 67: The name is a bit unfortunate IMHO. It indicates that it should be a bool.

pkg/exporter/sumologicexporter/config.go
```diff
@@ -255,4 +260,6 @@ const (
 	DefaultTimestampKey string = "timestamp"
 	// DefaultFlattenBody defines default FlattenBody value
 	DefaultFlattenBody bool = false
+	// DefaultDropRoutingAttribute defines default DropRoutingAttribute
+	DefaultDropRoutingAttribute string = "sumo_routing"
```
pmalek-sumo, line 264: I wonder if we shouldn't set the default to empty string and don't drop anything then (make it conditional).

@SumoLogic/open-source-collection-team  WDYT?

pkg/exporter/sumologicexporter/README.md
sumo-drosiek: would `routing_attribute_to_drop` be a better name?

pkg/exporter/sumologicexporter/config.go
sumo-drosiek: I agree, especially if we do not target it by defaut to `http_listener_v2_path`

pkg/exporter/sumologicexporter/README.md
pmalek-sumo: Not sure about either of those 😬 . But let's go with your proposal: `routing_attribute_to_drop`

sumo-drosiek: fixed then :shrug:

pkg/exporter/sumologicexporter/config.go
sumo-drosiek: fixed

swiatekm-sumo (approved):

PR #500, opened 2022-03-29
docs(configuration): add section on setting source category

astencel-sumo:


docs/Configuration.md
```diff
@@ -470,6 +474,175 @@ Example configuration with example log can be found in [/examples/logs_json/](/e
...
+to use other attributes' values to compose source category, like this:
+
+```yaml
+exporters:
+  sumologic:
+    source_category: "%{cloud.provider}/%{cloud.region}/%{cloud.platform}"
+```
+
+If you want data from different sources to have different source categories,
+you'll need to set a resource attribute named `_sourceCategory` earlier in the pipeline.
```
pmalek-sumo, line 503:
```suggestion
you'll need to set a resource attribute named (for example) `_sourceCategory` earlier in the pipeline.
```

docs/Configuration.md
```diff
@@ -470,6 +474,175 @@ Example configuration with example log can be found in [/examples/logs_json/](/e
...
 [opentelemetry-log-collection]: https://github.com/open-telemetry/opentelemetry-log-collection
 
+## Setting source category
+
+For many Sumo Logic customers the [source category][source_category_docs] is a crucial piece of metadata
+that allows to differentiate data coming from different sources.
+This section describes how to configure the Sumo Logic Distribution of OpenTelemetry
+to decorate data with this attribute.
+
+To decorate all the data from the collector with the same source category,
```
pmalek-sumo, line 484: Maybe mention here that it's totally possible to have multiple exporters and set separate source categories for each of those.

docs/Configuration.md
```diff
@@ -470,6 +474,175 @@ Example configuration with example log can be found in [/examples/logs_json/](/e
...
+[source_category_docs]: https://help.sumologic.com/03Send-Data/Sources/04Reference-Information-for-Sources/Metadata-Naming-Conventions#source-categories
+[sumologicexporter_source_templates]: https://github.com/SumoLogic/sumologic-otel-collector/blob/main/pkg/exporter/sumologicexporter/README.md#source-templates
+
+### Setting source category on logs from files
+
+The [Filelog receiver][filelogreceiver_readme] allows to set attributes on received data.
+Here's an example on how to attach different source categories to logs from different files
+with the Filelog receiver:
+
+```yaml
```
pmalek-sumo, line 515: Perhaps add a `NOTE` section that this relies on the above mentioned exporter config?

docs/Configuration.md
```diff
@@ -470,6 +474,175 @@ Example configuration with example log can be found in [/examples/logs_json/](/e
...
+          target_label: "_sourceCategory"
+        static_configs:
+        - targets:
+          - "0.0.0.0:8888"
+
+service:
+  pipelines:
+    metrics:
+      exporters:
+      - logging
```
pmalek-sumo, line 578: Why logging instead of sumologic exporter?

docs/Configuration.md
astencel-sumo: The attribute needs to specifically be named `_sourceCategory` to be used in Sumo as the source category.

docs/Configuration.md
astencel-sumo: The Filelog receiver config does not rely on the above mentioned exporter config.

docs/Configuration.md
pmalek-sumo: Do we have that documented somewhere in https://help.sumologic.com/ ?

docs/Configuration.md
astencel-sumo: Hmm yes that's an option. I'm not sure though if we want to encourage this way.

There are many different ways to achieve the same goal, but this documentation definitely should not (and probably cannot) describe all of them.

pmalek-sumo (approved):

PR #508, opened 2022-03-30
fix(telegrafreceiver): retry recoverable errors

swiatekm-sumo:
Retry recoverable errors from the rest of the pipeline using a simple constant backoff strategy. I've made the backoff configurable primarily to make testing easier, but users shouldn't really touch these values in most circumstances, which is why I'm not including them in the README.

### DIFF
--- a/pkg/receiver/telegrafreceiver/factory.go
+++ b/pkg/receiver/telegrafreceiver/factory.go
@@ -17,6 +17,7 @@ package telegrafreceiver
 import (
 	"context"
 	"fmt"
+	"time"
 
 	telegrafagent "github.com/influxdata/telegraf/agent"
 	telegrafconfig "github.com/influxdata/telegraf/config"
@@ -47,6 +48,10 @@ func createDefaultConfig() config.Receiver {
 	return &Config{
 		ReceiverSettings: &rs,
 		SeparateField:    false,
+		// we mostly expect to get recoverable errors from the memory_limiter, which is unlikely to have
+		// a check interval less than 1s, so retrying much more frequently is pointless
+		ConsumeRetryDelay: 500 * time.Millisecond,
+		ConsumeMaxRetries: 10,
 	}
 }
 
@@ -72,9 +77,11 @@ func createMetricsReceiver(
 	}
 
 	return &telegrafreceiver{
-		agent:           tAgent,
-		consumer:        nextConsumer,
-		logger:          params.Logger,
-		metricConverter: newConverter(tCfg.SeparateField, params.Logger),
+		agent:             tAgent,
+		consumer:          nextConsumer,
+		logger:            params.Logger,
+		metricConverter:   newConverter(tCfg.SeparateField, params.Logger),
+		consumeRetryDelay: tCfg.ConsumeRetryDelay,
+		consumeMaxRetries: tCfg.ConsumeMaxRetries,
 	}, nil
 }
--- a/pkg/receiver/telegrafreceiver/receiver.go
+++ b/pkg/receiver/telegrafreceiver/receiver.go
@@ -18,11 +18,14 @@ import (
 	"context"
 	"errors"
 	"sync"
+	"time"
 
+	backoff "github.com/cenkalti/backoff/v4"
 	"github.com/influxdata/telegraf"
 	telegrafagent "github.com/influxdata/telegraf/agent"
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/consumer/consumererror"
 	"go.opentelemetry.io/collector/model/pdata"
 	"go.uber.org/zap"
 )
@@ -39,10 +42,12 @@ type telegrafreceiver struct {
 	wg        sync.WaitGroup
 	cancel    context.CancelFunc
 
-	agent           *telegrafagent.Agent
-	consumer        consumer.Metrics
-	logger          *zap.Logger
-	metricConverter MetricConverter
+	agent             *telegrafagent.Agent
+	consumer          consumer.Metrics
+	logger            *zap.Logger
+	metricConverter   MetricConverter
+	consumeRetryDelay time.Duration
+	consumeMaxRetries uint64
 }
 
 // Ensure this receiver adheres to required interface.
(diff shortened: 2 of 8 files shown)


pkg/receiver/telegrafreceiver/factory.go
```diff
@@ -47,6 +48,10 @@ func createDefaultConfig() config.Receiver {
 	return &Config{
 		ReceiverSettings: &rs,
 		SeparateField:    false,
+		// we mostly expect to get recoverable errors from the memory_limiter, which is unlikely to have
+		// a check interval less than 1s, so retrying much more frequently is pointless
+		ConsumeRetryDelay: 500 * time.Millisecond,
+		ConsumeMaxRetries: 10,
```
pmalek-sumo, line 54: Add those defaults to receiver's README

pkg/receiver/telegrafreceiver/receiver.go
```diff
@@ -105,6 +110,37 @@ func (r *telegrafreceiver) Start(ctx context.Context, host component.Host) error
 	return err
 }
 
+// Consume metrics and retry on recoverable errors
+func (r *telegrafreceiver) consumeWithRetry(ctx context.Context, metrics pdata.Metrics) error {
+	constantBackoff := backoff.WithMaxRetries(backoff.NewConstantBackOff(r.consumeRetryDelay), uint64(r.consumeMaxRetries))
+
+	// retry handling according to https://github.com/open-telemetry/opentelemetry-collector/blob/master/component/receiver.go#L45
+	fErr := backoff.RetryNotify(
```
astencel-sumo, line 118: what does this mean `fErr`? Can we make the name more desriptive?

pmalek-sumo: What do you suggest :) ? `retryNotifyErr`? The scope of this func is pretty small (nice), no need IMHO to make it any more explicit/clear

pkg/receiver/telegrafreceiver/factory.go
swiatekm-sumo: I'm not sure this is a good idea. This is largely a technical detail that users should be able to ignore most of the time. It's only really configurable in case I'm wrong about the defaults in some specific circumstance.

pmalek-sumo: Well in this case it's a bit of an easter egg or security by obscurity :)

Not a huge fan of this approach.

swiatekm-sumo: It's something that is only configurable just in case, and I'm somewhat concerned that putting it in the README will be more confusing than helpful. If someone is actually interested, they can read the source code and find out that way.

swiatekm-sumo: Allright, the consensus is that we should document this with a caveat. Added to the README.

pkg/receiver/telegrafreceiver/receiver.go
```diff
@@ -105,6 +110,37 @@ func (r *telegrafreceiver) Start(ctx context.Context, host component.Host) error
...
+
+	// retry handling according to https://github.com/open-telemetry/opentelemetry-collector/blob/master/component/receiver.go#L45
+	err := backoff.RetryNotify(
+		func() error {
+			// we need to check for context cancellation here
+			select {
+			case <-ctx.Done():
+				return backoff.Permanent(errors.New("closing"))
+			default:
+			}
```
pmalek-sumo, line 125: Be wary of the fact that this might actually go through the `default` case even though `<-ctx.Done()` can also proceed.

pkg/receiver/telegrafreceiver/receiver.go
```diff
@@ -105,6 +110,37 @@ func (r *telegrafreceiver) Start(ctx context.Context, host component.Host) error
...
+			if consumererror.IsPermanent(err) {
+				return backoff.Permanent(err)
+			} else {
+				return err
+			}
+		},
+		constantBackoff,
+		func(err error, delay time.Duration) {
+			r.logger.Warn("ConsumeMetrics() recoverable error, will retry",
+				zap.String("error", err.Error()), zap.Duration("delay", delay),
```
pmalek-sumo, line 136:
```suggestion
				zap.Error(err), zap.Duration("delay", delay),
```

pkg/receiver/telegrafreceiver/receiver_test.go
```diff
@@ -42,3 +71,47 @@ func TestStartShutdown(t *testing.T) {
...
+		consumer:          consumer,
+		logger:            logger,
+		metricConverter:   newConverter(true, logger),
+		consumeRetryDelay: time.Nanosecond,
+		consumeMaxRetries: 10,
+	}
+
+	metrics := pdata.Metrics{}
+	err := receiver.consumeWithRetry(ctx, metrics)
+	assert.NotEqual(t, nil, err)
```
pmalek-sumo, line 114:
```suggestion
	assert.NoError(t, err)
```

pkg/receiver/telegrafreceiver/receiver.go
```diff
@@ -105,6 +110,37 @@ func (r *telegrafreceiver) Start(ctx context.Context, host component.Host) error
 	return err
 }
 
+// Consume metrics and retry on recoverable errors
+func (r *telegrafreceiver) consumeWithRetry(ctx context.Context, metrics pdata.Metrics) error {
+	constantBackoff := backoff.WithMaxRetries(backoff.NewConstantBackOff(r.consumeRetryDelay), r.consumeMaxRetries)
```
pmalek-sumo, line 115: I wonder if this won't end up slowing this whole thing down. Perhaps we could try reusing those backoffs? 🤔  Are you eager to benchmark this e.g. using a https://pkg.go.dev/sync#Pool ?

pkg/receiver/telegrafreceiver/receiver.go
swiatekm-sumo: Yeah, but that's ok in this case. I'm basically making sure that this exits cleanly during shutdown. At worst it'll get a permanent error from the rest of the pipeline and exit that way, or spit out a stack trace.

pkg/receiver/telegrafreceiver/receiver.go
swiatekm-sumo: Right now, we don't even need to be threadsafe, and could just reuse one backoff held in the receiver object.

I haven't benchmarked this, but intuitively it should be quite fast (definitely faster than metrics conversion), and I haven't seen any impact in integration tests. `ConstantBackoff` is a single-field struct, so it should be allocated on the stack: https://github.com/cenkalti/backoff/blob/6b0e4ad0cd65431b217393bff47d1dff727b264b/backoff.go#L57.

pmalek-sumo: Fair enough. One `time.Duration` should be lightweight enough to not pay too much attention to this.

pmalek-sumo (approved):

PR #513, opened 2022-03-31
fix(k8sprocessor): fix metadata dependencies

### LINKED ISSUE #512: [k8sprocessor] implicit dependencies between owner metadata
OT distro version: 0.0.57-beta.0

### Steps to reproduce

1. Configure `k8s_tagger` processor with `owner_lookup_enabled: true` and with extract metadata section that includes `deploumentName` but does not include `replicaSetName`, and with `cronJobName` but without `jobName`.

  ```yaml
  processors:
    k8s_tagger:
      owner_lookup_enabled: true
      extract:
        metadata:
         - deploymentName
         - cronJobName
  ```
2. Prepare a Kubernetes cluster that includes a Deployment and a CronJob.
2. Run the collector in a Kubernetes cluster with log and/or metrics collection from pods, observe the metadata on the logs and/or metrics.

### Actual result

- Data from pods that belong to a Deployment is not tagged with deployment name
- Data from pods that belong to a CronJob is not tagged with the cronjob's name

### Expected result

- Data from pods that belong to a Deployment is tagged with deployment name `k8s.deployment.name`
- Data from pods that belong to a CronJob is tagged with the cronjob's name `k8s.cronjob.name`

### How to fix

This is because the Pods that belong to a Deployment actually belong to a ReplicaSet, and that ReplicaSet belongs to the Deployment. With the above configuration, currently the ReplicaSet information is not retrieved from k8s API server, which it should be when the Deployment extraction is defined.

Same with CronJob - Job - Pod hierarchy.

astencel-sumo:
Fixes #512

### DIFF
--- a/pkg/processor/k8sprocessor/kube/client_test.go
+++ b/pkg/processor/k8sprocessor/kube/client_test.go
@@ -40,7 +40,23 @@ import (
 )
 
 func newFakeAPIClientset(_ k8sconfig.APIConfig) (kubernetes.Interface, error) {
-	return fake.NewSimpleClientset(), nil
+	fakeClient := fake.NewSimpleClientset()
+	// Add batch/v1 CronJob resource so that setting up a CronJob informer works in testing.
+	fakeClient.Fake.Resources = append(fakeClient.Fake.Resources, &metav1.APIResourceList{
+		GroupVersion: "batch/v1",
+		APIResources: []metav1.APIResource{
+			{
+				Name:         "cronjobs",
+				SingularName: "cronjob",
+				Namespaced:   true,
+				Group:        "batch",
+				Version:      "batch/v1",
+				Kind:         "CronJob",
+				ShortNames:   []string{"cj"},
+			},
+		},
+	})
+	return fakeClient, nil
 }
 
 func podAddAndUpdateTest(t *testing.T, c *WatchClient, handler func(obj interface{})) {
--- a/pkg/processor/k8sprocessor/kube/client_test.go
+++ b/pkg/processor/k8sprocessor/kube/client_test.go
@@ -42,6 +42,8 @@ import (
 func newFakeAPIClientset(_ k8sconfig.APIConfig) (kubernetes.Interface, error) {
 	fakeClient := fake.NewSimpleClientset()
 	// Add batch/v1 CronJob resource so that setting up a CronJob informer works in testing.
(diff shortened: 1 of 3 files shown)


pkg/processor/k8sprocessor/kube/owner_test.go
```diff
@@ -47,6 +48,201 @@ func waitForWatchToBeEstablished(client *fake.Clientset, resource string) <-chan
...
+	require.NoError(t, err)
+
+	assert.Eventually(t, func() bool {
+		owners := op.GetOwners(pod)
+		if len(owners) != 1 {
+			t.Logf("owners: %v", owners)
+			return false
+		}
+
+		if uid := owners[0].UID; uid != "fb9e6935-8936-4959-bd90-4e975a4c2b07" {
```
pmalek-sumo, line 127: Maybe it would be more readable to assign the ReplicaSet to a variable and then use its ID here instead of using magic string?
And similarly in other tests that are added with this PR.

WDYT?

astencel-sumo: OK, added variable names.

pkg/processor/k8sprocessor/kube/client_test.go
```diff
@@ -40,7 +40,23 @@ import (
 )
 
 func newFakeAPIClientset(_ k8sconfig.APIConfig) (kubernetes.Interface, error) {
-	return fake.NewSimpleClientset(), nil
+	fakeClient := fake.NewSimpleClientset()
+	// Add batch/v1 CronJob resource so that setting up a CronJob informer works in testing.
```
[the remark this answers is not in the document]
astencel-sumo: Added comment.

pmalek-sumo:
> One comment w.r.t to the tests is that I'd maybe be more generous with comments, i.e. add comments where we're testing this particular feature that is added in this PR so that we know what's being tested.

ping

astencel-sumo:
> > One comment w.r.t to the tests is that I'd maybe be more generous with comments, i.e. add comments where we're testing this particular feature that is added in this PR so that we know what's being tested.
> 
> ping

Here's how I see it: there was a functionality  that was not covered by tests (retrieving metadata for deployments and cronjobs), and this functionality didn't work correctly. I fixed the bug and added the tests for this functionality at the same time. I don't think there is anything else required with regards to this bug.

PR #514, opened 2022-03-31
chore(Makefile): use bash as default shell

sumo-drosiek:
https://stackoverflow.com/a/589300

Fixes:

```
➜  ~ cd projects/sumologic-otel-collector 
➜  sumologic-otel-collector git:(main) ✗ git reset --hard
HEAD is now at 1e0f157 docs(known-issues): describe filelog multiline issue
➜  sumologic-otel-collector git:(main) ✗ export TAG=v0.0.59-beta.1 ; make add-tag 
Adding tag v0.0.59-beta.1
/bin/sh: 2: Bad substitution
make: *** [Makefile:126: add-tag] Error 2
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 GOLANGCI_LINT_VERSION ?= v1.45.2
+SHELL := /usr/bin/env bash
 
 all: markdownlint yamllint
 


pmalek-sumo (approved):
Hm this worked for me regardless if I have this change or I don't have it.

Perhaps the reason for this is that I use ZSH and even in `sh` I have `SHELL` set to it

```
sh-3.2$ echo $SHELL
/bin/zsh
```

sumo-drosiek:
I have `$SHELL` set either
```
5.3.1 Choosing the Shell
------------------------

The program used as the shell is taken from the variable `SHELL'.  If
this variable is not set in your makefile, the program `/bin/sh' is
used as the shell.

```

There is nothing about using shell's `SHELL` env

PR #519, opened 2022-04-04
refactor(sumologicexporter): use golang.org/x/exp/slices for sorting fields

pmalek-sumo:
This PR uses generic slice sorting algorithm and saves yet another allocation in `fields.string()`.

Related article: https://eli.thegreenplace.net/2022/faster-sorting-with-go-generics/

```
benchcmp old.txt new.txt
benchcmp is deprecated in favor of benchstat: https://pkg.go.dev/golang.org/x/perf/cmd/benchstat
benchmark              old ns/op     new ns/op     delta
BenchmarkFields-16     646           618           -4.32%
BenchmarkFields-16     642           603           -6.03%
BenchmarkFields-16     636           614           -3.35%
BenchmarkFields-16     640           616           -3.72%
BenchmarkFields-16     648           612           -5.56%

benchmark              old allocs     new allocs     delta
BenchmarkFields-16     11             10             -9.09%
BenchmarkFields-16     11             10             -9.09%
BenchmarkFields-16     11             10             -9.09%
BenchmarkFields-16     11             10             -9.09%
BenchmarkFields-16     11             10             -9.09%

benchmark              old bytes     new bytes     delta
BenchmarkFields-16     368           344           -6.52%
BenchmarkFields-16     368           344           -6.52%
BenchmarkFields-16     368           344           -6.52%
BenchmarkFields-16     368           344           -6.52%
BenchmarkFields-16     368           344           -6.52%
```

**EDIT**: the linter fails because we don't have full generics support in golangci-lint just yet: https://github.com/golangci/golangci-lint/issues/2649

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -25,20 +25,24 @@ linters-settings:
     tab-width: 8
 
 linters:
-  disable-all: false
+  disable-all: true
   enable:
     - gofmt
     - deadcode
-    - unused
     - errcheck
     - goimports
     - misspell
     - noctx
     - lll
-  disable:
-    - maligned
-    - prealloc
-  fast: false
+    - govet
+    - ineffassign
+    - structcheck
+    - typecheck
+    - varcheck
+    # These linters are problematic when using go1.18 and generic code like in golang.org/x/exp/slices
+    # - unused
+    # - gosimple
+    # - staticcheck
 
 issues:
   # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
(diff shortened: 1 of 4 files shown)


sumo-drosiek (approved):

swiatekm-sumo:
I'm not sure if I'm reading this right, but from https://github.com/golangci/golangci-lint/issues/2649 it sounds like the linter should work if we explicitly set go version to 1.18 in the config? The issue description claims that they automatically disable linters which can't handle generics for go 1.18. @pmalek-sumo ?

PR #530, opened 2022-04-05
refactor(sumologicextension): use bytes slices and strings.Builder to decrease allocations

pmalek-sumo:
```
$ benchcmp main.txt pr.txt
benchmark                                          old ns/op     new ns/op     delta
Benchmark_PrometheusFormatter_Metric2String-16     66901         53862         -19.49%
Benchmark_PrometheusFormatter_Metric2String-16     65607         53773         -18.04%
Benchmark_PrometheusFormatter_Metric2String-16     65205         54047         -17.11%

benchmark                                          old allocs     new allocs     delta
Benchmark_PrometheusFormatter_Metric2String-16     696            512            -26.44%
Benchmark_PrometheusFormatter_Metric2String-16     696            512            -26.44%
Benchmark_PrometheusFormatter_Metric2String-16     696            512            -26.44%

benchmark                                          old bytes     new bytes     delta
Benchmark_PrometheusFormatter_Metric2String-16     20326         15793         -22.30%
Benchmark_PrometheusFormatter_Metric2String-16     20341         15785         -22.40%
Benchmark_PrometheusFormatter_Metric2String-16     20315         15811         -22.17%
```

### DIFF
--- a/pkg/exporter/sumologicexporter/prometheus_formatter.go
+++ b/pkg/exporter/sumologicexporter/prometheus_formatter.go
@@ -57,7 +57,14 @@ func newPrometheusFormatter() (prometheusFormatter, error) {
 
 // PrometheusLabels returns all attributes as sanitized prometheus labels string
 func (f *prometheusFormatter) tags2String(attr pdata.AttributeMap, labels pdata.AttributeMap) prometheusTags {
+	attrsPlusLabelsLen := attr.Len() + labels.Len()
+	if attrsPlusLabelsLen == 0 {
+		return ""
+	}
+
 	mergedAttributes := pdata.NewAttributeMap()
+	mergedAttributes.EnsureCapacity(attrsPlusLabelsLen)
+
 	attr.CopyTo(mergedAttributes)
 	labels.Range(func(k string, v pdata.AttributeValue) bool {
 		mergedAttributes.UpsertString(k, v.StringVal())
@@ -65,30 +72,83 @@ func (f *prometheusFormatter) tags2String(attr pdata.AttributeMap, labels pdata.
 	})
 	length := mergedAttributes.Len()
 
-	if length == 0 {
-		return ""
-	}
-
 	returnValue := make([]string, 0, length)
 	mergedAttributes.Range(func(k string, v pdata.AttributeValue) bool {
+		key := f.sanitizeKeyBytes([]byte(k))
+		value := f.sanitizeValue(v.AsString())
+
 		returnValue = append(
 			returnValue,
-			fmt.Sprintf(
-				`%s="%s"`,
-				f.sanitizeKey(k),
-				f.sanitizeValue(v.AsString()),
-			),
+			formatKeyValuePair(key, value),
 		)
 		return true
 	})
 
-	return prometheusTags(fmt.Sprintf("{%s}", strings.Join(returnValue, ",")))
+	return prometheusTags(stringsJoinAndSurround(returnValue, ",", "{", "}"))
 }
 
-// sanitizeKey returns sanitized key string by replacing
+func formatKeyValuePair(key []byte, value string) string {
+	const (
+		quoteSign = `"`
+		equalSign = `=`
+	)
+
+	// Use strings.Builder and not fmt.Sprintf as it uses significantly less
+	// allocations.
+	sb := strings.Builder{}
+	// We preallocate space for key, value, equal sign and quotes.
+	sb.Grow(len(key) + len(equalSign) + 2*len(quoteSign) + len(value))
+	sb.Write(key)
+	sb.WriteString(equalSign)
+	sb.WriteString(quoteSign)
+	sb.WriteString(value)
+	sb.WriteString(quoteSign)
+	return sb.String()
+}
+
+// stringsJoinAndSurround joins the strings in s slice using the separator adds front
+// to the front of the resulting string and back at the end.
+//
+// This has a benefit over using the strings.Join() of using just one strings.Buidler
+// instance and hence using less allocations to produce the final string.
+func stringsJoinAndSurround(s []string, separator, front, back string) string {
+	switch len(s) {
+	case 0:
+		return ""
+	case 1:
+		var b strings.Builder
+		b.Grow(len(s[0]) + len(front) + len(back))
+		b.WriteString(front)
+		b.WriteString(s[0])
+		b.WriteString(back)
+		return b.String()
+	}
+
+	// Count the total strings summarized length for the preallocation.
+	n := len(front) + len(s[0])
+	for i := 1; i < len(s); i++ {
+		n += len(separator) + len(s[i])
+	}
+	n += len(back)
+
+	var b strings.Builder
+	// We preallocate space for all the entires in the provided slice together with
+	// the separator as well as the surrounding characters.
+	b.Grow(n)
+	b.WriteString(front)
+	b.WriteString(s[0])
+	for _, s := range s[1:] {
+		b.WriteString(separator)
+		b.WriteString(s)
+	}
+	b.WriteString(back)
+	return b.String()
+}
+
+// sanitizeKeyBytes returns sanitized key byte slice by replacing
 // all non-allowed chars with `_`
-func (f *prometheusFormatter) sanitizeKey(s string) string {
-	return f.sanitNameRegex.ReplaceAllString(s, "_")
+func (f *prometheusFormatter) sanitizeKeyBytes(s []byte) []byte {
+	return f.sanitNameRegex.ReplaceAll(s, []byte{'_'})
 }
 
 // sanitizeKey returns sanitized value string performing the following substitutions:
@@ -103,7 +163,7 @@ func (f *prometheusFormatter) sanitizeValue(s string) string {
 func (f *prometheusFormatter) doubleLine(name string, attributes prometheusTags, value float64, timestamp pdata.Timestamp) string {
 	return fmt.Sprintf(
 		"%s%s %g %d",
-		f.sanitizeKey(name),
+		f.sanitizeKeyBytes([]byte(name)),
 		attributes,
 		value,
 		timestamp/pdata.Timestamp(time.Millisecond),
@@ -114,7 +174,7 @@ func (f *prometheusFormatter) doubleLine(name string, attributes prometheusTags,
 func (f *prometheusFormatter) intLine(name string, attributes prometheusTags, value int64, timestamp pdata.Timestamp) string {
 	return fmt.Sprintf(
 		"%s%s %d %d",
-		f.sanitizeKey(name),
+		f.sanitizeKeyBytes([]byte(name)),
 		attributes,
 		value,
 		timestamp/pdata.Timestamp(time.Millisecond),
@@ -125,7 +185,7 @@ func (f *prometheusFormatter) intLine(name string, attributes prometheusTags, va
 func (f *prometheusFormatter) uintLine(name string, attributes prometheusTags, value uint64, timestamp pdata.Timestamp) string {
 	return fmt.Sprintf(
 		"%s%s %d %d",
-		f.sanitizeKey(name),
+		f.sanitizeKeyBytes([]byte(name)),
 		attributes,
 		value,
 		timestamp/pdata.Timestamp(time.Millisecond),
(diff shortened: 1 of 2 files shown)


swiatekm-sumo:
Have you benchmarked how much of the improvement comes from introducing StringBuilder vs the other changes? The StringBuilder code is much uglier, and intuitively it shouldn't be that much faster than fmt.Sprintf for short strings.

pmalek-sumo:
> Have you benchmarked how much of the improvement comes from introducing StringBuilder vs the other changes? The StringBuilder code is much uglier, and intuitively it shouldn't be that much faster than fmt.Sprintf for short strings.

Just those 2 changes introducing `strings.Builder` in `tags2String()` contribute ~~20-25% of allocs and similarly in time:

```
benchmark                                          old ns/op     new ns/op     delta
Benchmark_PrometheusFormatter_Metric2String-16     49478         59523         +20.30%
Benchmark_PrometheusFormatter_Metric2String-16     50142         64352         +28.34%
Benchmark_PrometheusFormatter_Metric2String-16     49223         59849         +21.59%

benchmark                                          old allocs     new allocs     delta
Benchmark_PrometheusFormatter_Metric2String-16     538            674            +25.28%
Benchmark_PrometheusFormatter_Metric2String-16     538            674            +25.28%
Benchmark_PrometheusFormatter_Metric2String-16     538            674            +25.28%

benchmark                                          old bytes     new bytes     delta
Benchmark_PrometheusFormatter_Metric2String-16     16950         19635         +15.84%
Benchmark_PrometheusFormatter_Metric2String-16     16923         19645         +16.08%
Benchmark_PrometheusFormatter_Metric2String-16     16938         19654         +16.03%
```

pmalek-sumo:
Another comparison:

```go
	returnValue := make([]string, 0, length)
	mergedAttributes.Range(func(k string, v pdata.AttributeValue) bool {
		key := f.sanitizeKeyBytes([]byte(k))
		value := f.sanitizeValue(v.AsString())

		returnValue = append(
			returnValue,
			fmt.Sprintf(
				`%s="%s"`,
				key, value,
			),
		)
		return true
	})

	return prometheusTags(fmt.Sprintf("{%s}", strings.Join(returnValue, ",")))
```

```
goos: darwin
goarch: amd64
pkg: github.com/SumoLogic/sumologic-otel-collector/pkg/exporter/sumologicexporter
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
Benchmark_PrometheusFormatter_Metric2String
Benchmark_PrometheusFormatter_Metric2String-16             20386             58591 ns/op           19630 B/op        674 allocs/op
Benchmark_PrometheusFormatter_Metric2String-16             20182             59102 ns/op           19668 B/op        674 allocs/op
Benchmark_PrometheusFormatter_Metric2String-16             19946             61476 ns/op           19647 B/op        674 allocs/op
PASS
ok      github.com/SumoLogic/sumologic-otel-collector/pkg/exporter/sumologicexporter    5.821s
```

---

using `strings.Builder`

```go
	returnValue := make([]string, 0, length)
	mergedAttributes.Range(func(k string, v pdata.AttributeValue) bool {
		key := f.sanitizeKeyBytes([]byte(k))
		value := f.sanitizeValue(v.AsString())

		sb := strings.Builder{}
		sb.Grow(len(key) + len(value) + 1 + 2)
		sb.Write(key)
		sb.WriteRune('=')
		sb.WriteRune('"')
		sb.WriteString(value)
		sb.WriteRune('"')

		returnValue = append(
			returnValue,
			fmt.Sprintf(
				"%s", sb.String(),
			),
		)
		return true
	})

	ret := strings.Join(returnValue, ",")
	sb := strings.Builder{}
	sb.Grow(2 + len(ret))
	sb.WriteRune('{')
	sb.WriteString(ret)
	sb.WriteRune('}')

	return prometheusTags(sb.String())
```

```
goos: darwin
goarch: amd64
pkg: github.com/SumoLogic/sumologic-otel-collector/pkg/exporter/sumologicexporter
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
Benchmark_PrometheusFormatter_Metric2String
Benchmark_PrometheusFormatter_Metric2String-16             21086             55722 ns/op           18930 B/op        658 allocs/op
Benchmark_PrometheusFormatter_Metric2String-16             21694             57132 ns/op           18978 B/op        658 allocs/op
Benchmark_PrometheusFormatter_Metric2String-16             21644             58168 ns/op           18953 B/op        658 allocs/op
PASS
ok      github.com/SumoLogic/sumologic-otel-collector/pkg/exporter/sumologicexporter    5.771s
```

---

using `strings.Builder` and no `fmt`

```go
	returnValue := make([]string, 0, length)
	mergedAttributes.Range(func(k string, v pdata.AttributeValue) bool {
		key := f.sanitizeKeyBytes([]byte(k))
		value := f.sanitizeValue(v.AsString())

		sb := strings.Builder{}
		sb.Grow(len(key) + len(value) + 1 + 2)
		sb.Write(key)
		sb.WriteRune('=')
		sb.WriteRune('"')
		sb.WriteString(value)
		sb.WriteRune('"')

		returnValue = append(
			returnValue,
			sb.String(),
		)
		return true
	})

	ret := strings.Join(returnValue, ",")
	sb := strings.Builder{}
	sb.Grow(2 + len(ret))
	sb.WriteRune('{')
	sb.WriteString(ret)
	sb.WriteRune('}')

	return prometheusTags(sb.String())
```

```
goos: darwin
goarch: amd64
pkg: github.com/SumoLogic/sumologic-otel-collector/pkg/exporter/sumologicexporter
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
Benchmark_PrometheusFormatter_Metric2String
Benchmark_PrometheusFormatter_Metric2String-16             24160             51110 ns/op           16954 B/op        538 allocs/op
Benchmark_PrometheusFormatter_Metric2String-16             23751             50131 ns/op           16942 B/op        538 allocs/op
Benchmark_PrometheusFormatter_Metric2String-16             24018             49481 ns/op           16944 B/op        538 allocs/op
PASS
ok      github.com/SumoLogic/sumologic-otel-collector/pkg/exporter/sumologicexporter    5.389s
```

swiatekm-sumo:
Pretty significant difference after all, then. Ok, I'm in favor of using the StringBuilders, let's just make it clear why we're doing it this way with comments and possibly moving the key=value formatting to a separate function.

pmalek-sumo:
PTAL @swiatekm-sumo

swiatekm-sumo (approved):
LGTM. Do you reckon we should add separate unit tests for `stringsJoinAndSurround` or are we sufficiently covered by existing ones?

pmalek-sumo:
> LGTM. Do you reckon we should add separate unit tests for `stringsJoinAndSurround` or are we sufficiently covered by existing ones?

We could in theory. OTOH it's tested indirectly in other tests.

sumo-drosiek (approved):

pkg/exporter/sumologicexporter/prometheus_formatter.go
```diff
@@ -57,38 +57,90 @@ func newPrometheusFormatter() (prometheusFormatter, error) {
...
-	return prometheusTags(fmt.Sprintf("{%s}", strings.Join(returnValue, ",")))
+	return prometheusTags(stringsJoinAndSurround(returnValue, ',', '{', '}'))
+}
+
+func formatKeyValuePair(key []byte, value string) string {
+	// Use strings.Builder and not fmt.Sprintf as it uses significantly less
+	// allocations.
+	sb := strings.Builder{}
+	// We preallocate space for key, value, equal sign and quotes.
+	sb.Grow(len(key) + len(value) + 1 + 2)
```
pdelewski, line 95: I think that it would be more readable to not use hardcoded values in this case
`const equalSign string = "="`
 `const quoteSign string = "\""`
 `sb.Grow(len(key) + len(value) + len(equalSign) + len(quoteSign) + len(quoteSign)`
This still does not guarantee consistency of allocated buffer and real space used to contain all characters.
One example of fixing it, would be, hiding string builder under higher level builder api with at least two functions for one allocating space and second for building string which would check (at runtime) consistency between allocated space and really used one.

swiatekm-sumo: What would be the outcome of that check in your view @pdelewski? Let's say you try to write more data to the buffer than what is allocated, StringBuilder will automatically allocate more space. Should we then panic, print a warning, or something else, instead?

pdelewski: It depends if that preallocation is critical for us and if that's the case we can return error or panic

pmalek-sumo: We can benchmark this with and without the preallocation if we really want to find the performance difference between the 2 approaches but @swiatekm-sumo has a point that `strings.Builder` does re-allocate when needed and it's its feature so there's really no point to work around that I'd say.

To address your suggestion @pdelewski we could make those `const`s but then what type should they be? If we stick to `rune`s then they are always of length 1 and **you cannot take `rune`'s length** i.e.

```go
r := '['
l := len(r) // <- this won't compile
```

In that case we'd need to resort to strings but then we'd have to measure if it makes sense from performance standpoint (probably it does).

pdelewski: Do we need to use runes? Are they better, faster in this particular case?

pmalek-sumo: It looks like it's basically the same:

This PR:

```
goos: darwin
goarch: amd64
pkg: github.com/SumoLogic/sumologic-otel-collector/pkg/exporter/sumologicexporter
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
Benchmark_PrometheusFormatter_Metric2String
Benchmark_PrometheusFormatter_Metric2String-16             22712             51177 ns/op           15821 B/op        512 allocs/op
Benchmark_PrometheusFormatter_Metric2String-16             23038             50360 ns/op           15791 B/op        512 allocs/op
Benchmark_PrometheusFormatter_Metric2String-16             23560             51604 ns/op           15810 B/op        512 allocs/op
PASS
ok      github.com/SumoLogic/sumologic-otel-collector/pkg/exporter/sumologicexporter    5.578s
```

---

This approach:

```go
	sb.WriteString("=")
	sb.WriteString(`"`)
	sb.WriteString(value)
	sb.WriteString(`"`)
```

```
goos: darwin
goarch: amd64
pkg: github.com/SumoLogic/sumologic-otel-collector/pkg/exporter/sumologicexporter
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
Benchmark_PrometheusFormatter_Metric2String
Benchmark_PrometheusFormatter_Metric2String-16             23334             50641 ns/op           15804 B/op        512 allocs/op
Benchmark_PrometheusFormatter_Metric2String-16             24022             51114 ns/op           15809 B/op        512 allocs/op
Benchmark_PrometheusFormatter_Metric2String-16             24036             49831 ns/op           15811 B/op        512 allocs/op
```

pdelewski: I would stick to string if you don't mind (it's simpler to read as we don't need to mix different types and we avoid questions why runes are there)

pmalek-sumo: Please post your review then :)  👍 / 👎

pkg/exporter/sumologicexporter/prometheus_formatter.go
```diff
@@ -57,38 +57,95 @@ func newPrometheusFormatter() (prometheusFormatter, error) {
...
+// This has a benefit over using the strings.Join() of using just one strings.Buidler
+// instance and hence using less allocations to produce the final string.
+func stringsJoinAndSurround(s []string, separator rune, front rune, back rune) string {
+	switch len(s) {
+	case 0:
+		return ""
+	case 1:
+		var b strings.Builder
+		b.Grow(len(s[0]) + 2)
+		b.WriteRune(front)
```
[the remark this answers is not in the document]
pmalek-sumo: Done.

pkg/exporter/sumologicexporter/prometheus_formatter.go
pdelewski: :)

PR #536, opened 2022-04-07
fix(sumologicexporter): treat resource attributes as fields for otlp

pmalek-sumo:
This PR aims at treating OTLP resource level attributes as [fields](https://help.sumologic.com/Manage/Fields).

It also simplifies the OTLP flows a bit by removing the unnecessary batching from their logic.

### DIFF
--- a/pkg/exporter/sumologicexporter/exporter_test.go
+++ b/pkg/exporter/sumologicexporter/exporter_test.go
@@ -1454,7 +1454,7 @@ func TestPushOTLPMetrics_AttributeTranslation(t *testing.T) {
 				cfg := createConfig()
 				// NOTE: MetadataAttributes does not have an impact on exporter
 				// behavior when using OTLP. What gets sent as fields is purely
-				// depdendent on what's in resource attributes.
+				// dependent on what's in resource attributes.
 				cfg.MetadataAttributes = []string{}
 				cfg.SourceCategory = "source_category_with_hostname_%{host.name}"
 				cfg.SourceHost = "%{host}"
--- a/pkg/exporter/sumologicexporter/exporter_test.go
+++ b/pkg/exporter/sumologicexporter/exporter_test.go
@@ -276,6 +276,8 @@ func TestPushInvalidCompressor(t *testing.T) {
 }
 
 func TestPushFailedBatch(t *testing.T) {
+	t.Parallel()
+
 	test := prepareExporterTest(t, createTestConfig(), []func(w http.ResponseWriter, req *http.Request){
 		func(w http.ResponseWriter, req *http.Request) {
 			w.WriteHeader(500)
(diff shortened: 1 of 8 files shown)


pkg/exporter/sumologicexporter/fields.go
```diff
@@ -24,15 +24,21 @@ import (
 
 // fields represents metadata
 type fields struct {
-	orig pdata.AttributeMap
+	orig        pdata.AttributeMap
+	initialized bool
```
sumo-drosiek, line 28: isn't there a way to figure it out using orig?

pmalek-sumo: Couldn't find a way. You can't access `pdata.AttributeMap` if it's uninitialized because it will cause a panic. We'd have to restructure code around `fields` a bit more to achieve this I guess.

pkg/exporter/sumologicexporter/exporter_test.go
```diff
@@ -1425,51 +1444,42 @@ func TestPushOTLPMetrics_AttributeTranslation(t *testing.T) {
...
 	}{
 		{
 			name: "enabled",
 			cfgFn: func() *Config {
 				cfg := createConfig()
-				cfg.MetadataAttributes = []string{`host\.name`}
-				cfg.SourceCategory = "%{host.name}"
+				// NOTE: MetadataAttributes does not have an impact on exporter
+				// behavior when using OTLP. What gets sent as fields is purely
+				// depdendent on what's in resource attributes.
```
sumo-drosiek, line 1457: depdendent

sumo-drosiek, line 1457:
```suggestion
				// dependent on what's in resource attributes.
```

pkg/exporter/sumologicexporter/sender.go
```diff
@@ -418,11 +419,6 @@ func isEmptyAttributeValue(att pdata.AttributeValue) bool {
 // to configured LogFormat and as the result of execution
 // returns array of records which has not been sent correctly and error
 func (s *sender) sendLogs(ctx context.Context, flds fields) ([]logPair, error) {
-	// Follow different execution path for OTLP format
-	if s.config.LogFormat == OTLPLogFormat {
-		return s.sendOTLPLogs(ctx, flds)
```
pmm-sumo, line 423: Not sure if it's plausible but maybe return an error just in case?

pmm-sumo: I.e. 
```
	if s.config.LogFormat == OTLPLogFormat {
		return nil, fmt.Errorf("OTLPLogFormat was attempted to be sent as non-OTLP data")
```

Or some sort of assertion that it is not OTLPLogFormat

pmm-sumo (approved):

pkg/exporter/sumologicexporter/sender.go
pmalek-sumo: Added

sumo-drosiek (approved):

PR #537, opened 2022-04-07
docs: describe how OpenTelemetry data model maps to Sumo Logic

pmm-sumo:
This is first take on adding docs on OTLP vs Fields and JSON model. 

Let me know if you see a section of documentation that would be a better fit


docs/Comparison.md
```diff
@@ -76,6 +76,71 @@
...
+
+- **Resource** - it includes Attributes describing the resource from which given set of data comes from. Should follow resource [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions).
+- **Instrumentation Scope** - additional information about the scope of data (e.g. instrumentation library name).
+- **Record** - specific record (Log, Span, Metric) that also includes its own set of Attributes. May follow given signal type semantic convetions (e.g. [trace](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions), [metrics](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions), [logs](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/logs/semantic_conventions)), though they may also contain key/values that are specific to the context of the record. Additionally, Logs can also include attributes in the Body.
```
pmalek-sumo, line 85:
```suggestion
- **Record** - specific record (Log, Span, Metric) that also includes its own set of Attributes. May follow given signal type semantic conventions (e.g. [trace](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions), [metrics](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions), [logs](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/logs/semantic_conventions)), though they may also contain key/values that are specific to the context of the record. Additionally, Logs can also include attributes in the Body.
```

docs/Comparison.md
```diff
@@ -76,6 +76,71 @@
...
+
+At Sumo Logic, there is a concept of [Fields](https://help.sumologic.com/Manage/Fields) for log data. Fields offer a powerful capability to associate indexable metadata with logs, though only limited number of them can be used at a given time. Also, they need to be defined first.
+
+Looking from the OpenTelemetry standpoint, Fields are a good match for Resource-level attributes, while Log Record-level attributes are good fit for [structured representation of the log via JSON](https://help.sumologic.com/05Search/Get-Started-with-Search/Search-Basics/View-Search-Results-for-JSON-Logs), which is automatically supported by Sumo Logic Search.
+
+All resource-level attributes are stored as fields. If a matching field is not defined, it will be dropped. When log contains record-level attributes, they are stored as JSON representation. Body, if present, is then stored under `log` key.
```
pmalek-sumo, line 93: Maybe link to https://help.sumologic.com/Manage/Fields#view-dropped-fields when mentioning dropped fields?

pmalek-sumo (commented):
IMHO despite being nicely written I believe this doesn't fit in this doc. In my mind there should be a new, top level doc that contain this added section, ideally linked from the top README.md explaining how OT concepts translated into Sumo Logic ones.

sumo-drosiek:
From my side just ask to wrap long lines along multiple lines

```
This
is
very
long
paragraph
```

pmm-sumo:
@pmalek-sumo @sumo-drosiek thanks for the initial review, just updated following your suggestions

pmalek-sumo:
> IMHO despite being nicely written I believe this doesn't fit in this doc. In my mind there should be a new, top level doc that contains this added section, ideally linked from the top README.md explaining how OT concepts translated into Sumo Logic ones.

What do you think about this @pmm-sumo?

pmm-sumo:
> > IMHO despite being nicely written I believe this doesn't fit in this doc. In my mind there should be a new, top level doc that contains this added section, ideally linked from the top README.md explaining how OT concepts translated into Sumo Logic ones.
> 
> What do you think about this @pmm-sumo?

I was thinking about this when writing this doc and wasn't sure. I think there's not enough contents to justify separate page, but maybe it will grow and we could add details on other signals vs Sumo (e.g. which metrics we currently support and which we don't). Do you have suggestions how such top-level doc could be named?

pmalek-sumo:
> I think there's not enough contents to justify separate page

I don't agree with that. I mean, yes, there is indeed not that much content given it would only contain what's in this PR but trying to piggy back on an existing doc - specifically here comparing Sumo Logic Installed Collector with Sumo OT distro - doesn't make much sense to me.

> but maybe it will grow and we could add details on other signals vs Sumo

Yes, given it would be related to OT features, naming etc. translated into Sumo specific features.

> Do you have suggestions how such top-level doc could be named?

Something like "Mapping OpenTelemetry concepts onto Sumo" or "How does Sumo interpret those OpenTelemetry concepts"?

pmm-sumo:
There you go @pmalek-sumo :)

pmalek-sumo (approved):
😍 Nice

PR #547, opened 2022-04-13
feat: add rawk8seventsreceiver

swiatekm-sumo:
Receiver for Kubernetes events with a data format compatible with https://github.com/SumoLogic/sumologic-kubernetes-fluentd/tree/main/fluent-plugin-events. The actual logic, and a good chunk of the code, is similar to the upstream receiver: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8seventsreceiver.

I'd like to merge this as an experimental component, and only enable it in the build after I add some more features, which I'd rather do separately to avoid making this PR even bigger.

To do:
- [x] some minor changes to the data format
- [x] more tests
- [x] missing comments

Notes: 
The `k8sconfig.go` file is copied from upstream, as it's an internal package that I'd like to depend on for standardized K8s client configuration. As such, there isn't much point reviewing it.


pkg/receiver/rawk8seventsreceiver/README.md
```diff
@@ -0,0 +1,38 @@
...
+- `namespaces` (default value is `[]`): Namespaces to collect events from. Empty array means all namespaces and is the
+  default.
+- `max_event_age` (default value is `1m`): Maximum age of collected events relative to receiver start time. The default
+  setting of 1 minute means that we're going to collect events up to 1 minute backwards in time.
+- `auth_type` (default value is `serviceAccount`): Authentication type for the Kubernetes client. Valid values are:
+  `serviceAccount`, `kubeConfig`, `tls` and `none`.
+- `consume_retry_delay` (default value is `500ms`): The retry delay for recoverable errors from the rest of the pipeline.
+  Don't change this or the related setting below unless you know what you're doing.
+- `consume_max_retries` (default value is `10`): The maximum number of retries for recoverable errors from the rest of
+  the pipeline.
```
pmalek-sumo, line 23: Don't we want to follow the same configuration doc format as in e.g. https://github.com/SumoLogic/sumologic-otel-collector/tree/main/pkg/exporter/sumologicexporter#sumo-logic-exporter ?

pkg/receiver/rawk8seventsreceiver/testdata/config.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+exporters:
+  nop:
+
+service:
+  pipelines:
+    logs:
+      receivers: [raw_k8s_events]
+      processors: [nop]
+      exporters: [nop]
```
pmalek-sumo, line 21: EOL

pkg/receiver/rawk8seventsreceiver/receiver_test.go
```diff
@@ -0,0 +1,264 @@
...
+		consumer,
+		client,
+		fakeListWatchFactory,
+	)
+	require.NoError(t, err)
+	receiver.ctx = ctx
+
+	logs := pdata.Logs{}
+	err = receiver.consumeWithRetry(ctx, logs)
+	assert.NotEqual(t, nil, err)
```
pmalek-sumo, line 137:
```suggestion
	assert.Error(t, err)
```

pkg/receiver/rawk8seventsreceiver/receiver.go
```diff
@@ -0,0 +1,240 @@
...
+	return eventTime.After(r.startTime.Add(-r.cfg.MaxEventAge))
+}
+
+func (r *rawK8sEventsReceiver) convertToLog(event *corev1.Event) pdata.Logs {
+	ld := pdata.NewLogs()
+	rl := ld.ResourceLogs().AppendEmpty()
+	sl := rl.ScopeLogs().AppendEmpty()
+	lr := sl.LogRecords().AppendEmpty()
+
+	// Convert the event into a map[string][interface{}]
```
pmalek-sumo, line 197:
```suggestion
	// Convert the event into a map[string]interface{}
```

pkg/receiver/rawk8seventsreceiver/receiver.go
```diff
@@ -0,0 +1,240 @@
...
+	eventMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(event)
+	if err != nil {
+		r.logger.Error("failed to convert event", zap.Error(err), zap.Any("event", event))
```
pmalek-sumo, line 200: We continue on in case of an error?

pkg/receiver/rawk8seventsreceiver/receiver.go
```diff
@@ -0,0 +1,240 @@
...
+	// Set the "SeverityNumber" and "SeverityText" if a known type of severity is found.
+	if severityNumber, ok := severityMap[strings.ToLower(event.Type)]; ok {
+		lr.SetSeverityNumber(severityNumber)
+		lr.SetSeverityText(event.Type)
+	} else {
+		r.logger.Debug("unknown severity type", zap.String("type", event.Type))
+	}
+
+	attrs := lr.Attributes()
+	pdataObjectMap.CopyTo(attrs)
```
pmalek-sumo, line 221:
```suggestion
	pdataObjectMap.CopyTo(lr.Attributes())
```

pkg/receiver/rawk8seventsreceiver/receiver.go
```diff
@@ -0,0 +1,240 @@
...
+	eventCh          chan *corev1.Event
+	ctx              context.Context
+	cancel           context.CancelFunc
+	startTime        time.Time
+
+	consumer consumer.Logs
+	logger   *zap.Logger
+}
+
+// Interface for creating ListerWatcher objects. Used for injecting mocks into k8s informers.
```
pmalek-sumo, line 56: It's not an interface but a type 🙃

pkg/receiver/rawk8seventsreceiver/config_test.go
```diff
@@ -0,0 +1,43 @@
...
+	assert.NoError(t, err)
+
+	factory := NewFactory()
+	factories.Receivers[typeStr] = factory
+	cfg, err := servicetest.LoadConfigAndValidate(filepath.Join("testdata", "config.yaml"), factories)
+
+	require.NoError(t, err)
+	require.NotNil(t, cfg)
+
+	assert.Equal(t, len(cfg.Receivers), 2)
```
pmalek-sumo, line 40:
```suggestion
	assert.Len(t, cfg.Receivers, 2)
```

pkg/receiver/rawk8seventsreceiver/receiver.go
swiatekm-sumo: Good catch, we should return it.

pkg/receiver/rawk8seventsreceiver/README.md
swiatekm-sumo: This is how some core components do it, though there isn't much consistency there either. I could change it, or only document the important settings here and the rest as comments in the example config.yaml.

pkg/receiver/rawk8seventsreceiver/receiver.go
swiatekm-sumo: It's colloquially an interface, because afaik you can't define an actual callable interface in Go. I can change the wording if you think it's confusing.

pmalek-sumo: > It's colloquially an interface

Don't know what that means 😆 

> afaik you can't define an actual callable interface in Go. 

Because that's not a thing. You call functions, you fulfill interfaces.

pkg/receiver/rawk8seventsreceiver/README.md
```diff
@@ -0,0 +1,38 @@
+# Raw Kubernetes Events Receiver
+
+Receiver for ingesting Kubernetes Events in their raw format, exactly as the Kubernetes API returns them. It intends
+to return exactly the same output as the following [FluentD plugin].
```
sumo-drosiek, line 4:
```suggestion
Receiver for ingesting Kubernetes Events in their raw format, exactly as the Kubernetes API returns them.
It intends to return exactly the same output as the following [FluentD plugin].
```

pkg/receiver/rawk8seventsreceiver/README.md
sumo-drosiek: I would go with "our" format. Seems to be the most readable and usable

sumo-drosiek: Also I don't see `ReceiverSettings` here

pkg/receiver/rawk8seventsreceiver/k8sconfig.go
```diff
@@ -0,0 +1,163 @@
...
+func (c APIConfig) Validate() error {
+	if !authTypes[c.AuthType] {
+		return fmt.Errorf("invalid authType for kubernetes: %v", c.AuthType)
+	}
+
+	return nil
+}
+
+// createRestConfig creates an Kubernetes API config from user configuration.
+func createRestConfig(apiConf APIConfig) (*rest.Config, error) {
```
sumo-drosiek, line 76: Could we add link to original code?

pkg/receiver/rawk8seventsreceiver/receiver.go
```diff
@@ -0,0 +1,239 @@
...
+	if listerWatcherFactory == nil {
+		listerWatcherFactory = func(
+			c cache.Getter,
+			resource string,
+			namespace string,
+			fieldSelector fields.Selector,
+		) cache.ListerWatcher {
+			return cache.NewListWatchFromClient(c, resource, namespace, fieldSelector)
+		}
+	}
```
sumo-drosiek, line 78: Is this because of test?

pkg/receiver/rawk8seventsreceiver/receiver.go
```diff
@@ -0,0 +1,239 @@
...
+	if r.isEventAccepted(event) {
+		logs := r.convertToLog(event)
+		err := r.consumeWithRetry(ctx, logs)
+		if err != nil {
+			r.logger.Error("ConsumeMetrics() error",
+				zap.String("error", err.Error()),
+			)
+		}
+	}
+}
```
sumo-drosiek, line 184: I'm not a fun of ifing all function, I would rather use

```
if !r.isEventAccepted(event) {
    return
}
```

pkg/receiver/rawk8seventsreceiver/receiver.go
pmalek-sumo: I guess because of the call in `pkg/receiver/rawk8seventsreceiver/factory.go:78` 🤔 but I'd try to avoid this kind of thing.

Perhaps we could accept an `interface` instead and just use its method(s) then the code will get much simpler? And then you can use test factory implementation(s) in tests.

sumo-drosiek: `return newRawK8sEventsReceiver(params, rCfg, consumer, k8sClient, nil)`

we always pass `nil`, so this is not a "normal" behaviour to have `listerWatcherFactory`

pkg/receiver/rawk8seventsreceiver/k8sconfig.go
swiatekm-sumo: There's a comment with a link above the package definition start.

pkg/receiver/rawk8seventsreceiver/receiver.go
swiatekm-sumo: This is because of tests. The normal fake k8s client doesn't support the REST API, so in order to test informers, we need to mock the ListerWatcher - that's what the factory is for, and it's a factory because we need to potentially create one for each namespace we're watching. See `receiver_test.go` for how it's used.

The reason we need a wrapper here is that the return value of `ListerWatcherFactory` is `cache.ListerWatcher`, which is an interface, while `cache.NewListWatchFromClient` returns a `*cache.ListWatch`, which is a concrete type. Even though `cache.ListWatch` fulfills `cache.ListerWatcher`, these are different function types in Go.

@pmalek-sumo I'm not sure how we could make this simpler with an interface, can you elaborate.

pmalek-sumo: Please see: https://github.com/SumoLogic/sumologic-otel-collector/commit/9b80f40b03a5011efd58f811b0950c3d6886df91#diff-d9e51d1374cdf68c9bda647ff549bc33993e5ad1e16b774601d2c9cfc191955e

That's why I suggested the interface 🙃 If it's a func it's a different type but if it's an interface then it's ok to return a concrete type fulfilling the interface

swiatekm-sumo: That's what I thought you meant, but in my view it's not really any simpler or more intuitive. My solution wraps the function type in another function, while yours wraps it in an interface. So you get simpler code at the cost of more complex types and more boilerplate.

With that said, I don't have a strong opinion on this. If your way is considered more idiomatic in Go, I can change to it.

pmalek-sumo: Depends on the perspective I guess. With `interface` approach you get the flexibility "at the expense" of an interface declaration. You can then pass anything you want to this function as long as it fulfils the interface and you separate the function body (as in the discussed snippet) from the actual "factory" implementation (which the caller **shouldn't** be bothered about IMHO).

swiatekm-sumo: I added a comment explaining why the wrapper exists. I generally like just using function types as factories, so I'd prefer to keep this as is.

pkg/receiver/rawk8seventsreceiver/receiver.go
```diff
@@ -0,0 +1,297 @@
...
+	logger   *zap.Logger
+}
+
+// Function type for creating ListerWatcher objects. Used for injecting mocks into k8s informers.
+type ListerWatcherFactory func(c cache.Getter, resource string, namespace string, fieldSelector fields.Selector) cache.ListerWatcher
+
+// We care about event creation and updates. This struct carries information about these changes.
+type eventChange struct {
+	event      *corev1.Event
+	changeType string // can be ADDED or MODIFIED
```
sumo-drosiek, line 64: could we use custom type instead of string?

sumo-drosiek (commented):
Generally LGTM. One more nit, and I don't understand the following comment

> Will add the changelog entry after I fix the Changelog in a separate PR.

swiatekm-sumo:
> Generally LGTM. One more nit, and I don't understand the following comment
> 
> > Will add the changelog entry after I fix the Changelog in a separate PR.

I wanted to merge #554 first, will add the changelog shortly.

pkg/receiver/rawk8seventsreceiver/receiver.go
pmalek-sumo: The problem is that with this approach you have test specific code in production code.

With interface you have a customization point, without a need to mention that it's for mocks.

🤷

swiatekm-sumo: Hm, that's a good point. I can avoid that by moving the wrapper into the factory, though. Sounds like a good idea in general.

pkg/receiver/rawk8seventsreceiver/README.md
```diff
@@ -0,0 +1,41 @@
...
+    namespaces: []
+
+    # Maximum age of collected events relative to receiver start time.
+    # The default setting of 1 minute means that we're going to collect events up to 1 minute backwards in time.
+    # default = 1m
+    max_event_age: 1m
+
+    # Authentication type for the Kubernetes client. Valid values are: `serviceAccount`, `kubeConfig`, `tls` and `none`.
+    # default = `serviceAccount`
+    auth_type: serviceAccount
```
sumo-drosiek, line 26: It would be nice to point for some configuration details (how to change host, port and so on)

pkg/receiver/rawk8seventsreceiver/factory.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	listerWatcherFactory := func(
+		c cache.Getter,
+		resource string,
+		namespace string,
+		fieldSelector fields.Selector,
+	) cache.ListerWatcher {
+		return cache.NewListWatchFromClient(c, resource, namespace, fieldSelector)
+	}
```
sumo-drosiek, line 87:
```suggestion
	listerWatcherFactory := cache.NewListWatchFromClient
```

? :thinking:

pkg/receiver/rawk8seventsreceiver/receiver.go
```diff
@@ -0,0 +1,291 @@
...
+func (r *rawK8sEventsReceiver) Shutdown(context.Context) error {
+	r.cancel()
+	return nil
+}
+
+// Consume metrics and retry on recoverable errors
+func (r *rawK8sEventsReceiver) consumeWithRetry(ctx context.Context, logs plog.Logs) error {
+	constantBackoff := backoff.WithMaxRetries(backoff.NewConstantBackOff(r.cfg.ConsumeRetryDelay), r.cfg.ConsumeMaxRetries)
+
+	// retry handling according to https://github.com/open-telemetry/opentelemetry-collector/blob/master/component/receiver.go#L45
```
sumo-drosiek, line 151: `master` => specific commit

pkg/receiver/rawk8seventsreceiver/README.md
swiatekm-sumo: You can't change any of that in this configuration though. Host and port are read from standard Kubernetes environment variables.

pkg/receiver/rawk8seventsreceiver/factory.go
swiatekm-sumo: Unfortunately, these are different types. `cache.NewListWatchFromClient` returns a `cache.ListWatch`, which is a concrete type, while `ListerWatcherFactory` returns an interface. The concrete type fulfills the interface, but Go doesn't allow for any kind of conversion between these function types as far as I'm aware. See https://github.com/SumoLogic/sumologic-otel-collector/pull/547#discussion_r851229357.

pkg/receiver/rawk8seventsreceiver/README.md
sumo-drosiek: I know it, but there is no indication which envs can be changed here

swiatekm-sumo: I don't know if I want to reproduce Kubernetes client documentation here, which this would come down to.

sumo-drosiek (approved):

PR #549, opened 2022-04-14
fix!(sumologicexporter): send resource attributes as fields for non-otlp, removing metadata_attributes

pmalek-sumo:


pkg/exporter/sumologicexporter/README.md
```diff
@@ -193,44 +184,6 @@ Below is a list of all attribute keys that are being translated.
 
 ## Source Templates
```
sumo-drosiek, line 185: We should explicitly say that this is for and only for resource attributes

pkg/exporter/sumologicexporter/config.go
```diff
@@ -68,9 +68,6 @@ type Config struct {
 	// `procstat_num_threads` => `Proc_Threads` or `cpu_usage_irq` => `CPU_Irq`).
 	TranslateTelegrafMetrics bool `mapstructure:"translate_telegraf_attributes"`
 
-	// List of regexes for attributes which should be send as metadata
-	MetadataAttributes []string `mapstructure:"metadata_attributes"`
-
```
sumo-drosiek, line 73: hmm, we could make validation and exit program with meaningful error message for few releases

pkg/exporter/sumologicexporter/exporter_test.go
```diff
@@ -654,12 +627,12 @@ func TestPushJSONLogs_AttributeTranslation(t *testing.T) {
...
-					regex += `{"host.type":"wizard","log":"Example log","timestamp":\d{13}}`
+					regex += `{"log":"Example log","timestamp":\d{13}}`
 					assert.Regexp(t, regex, body)
 
-					assert.Equal(t, "host.name=harry-potter", req.Header.Get("X-Sumo-Fields"))
-					assert.Equal(t, "harry-potter", req.Header.Get("X-Sumo-Category"))
-					assert.Equal(t, "undefined", req.Header.Get("X-Sumo-Host"))
+					assert.Equal(t, "host.name=harry-potter, host.type=wizard", req.Header.Get("X-Sumo-Fields"), "X-Sumo-Fields")
+					assert.Equal(t, "harry-potter", req.Header.Get("X-Sumo-Category"), "X-Sumo-Category")
+					assert.Equal(t, "undefined", req.Header.Get("X-Sumo-Host"), "X-Sumo-Host")
```
sumo-drosiek, line 635: I'm not sure if we need the third argument, if yes, we should use it for all comparisons in tests :shrug:

pkg/exporter/sumologicexporter/exporter_test.go
```diff
@@ -740,25 +712,43 @@ func TestPushLogs_DontRemoveSourceAttributes(t *testing.T) {
...
+			metricFunc:   exampleIntMetric,
 		},
-	})
-	test.exp.config.MetricFormat = PrometheusFormat
+		{
+			name: "gauge",
+			expectedBody: `` +
+				`gauge_metric_name{foo="bar",remote_name="156920",url="http://example_url"} 124 1608124661166` +
+				"\n" +
+				`gauge_metric_name{foo="bar",remote_name="156955",url="http://another_url"} 245 1608124662166`,
```
sumo-drosiek, line 730: not sure if this is better format than before

pkg/exporter/sumologicexporter/exporter_test.go
```diff
@@ -740,25 +712,43 @@ func TestPushLogs_DontRemoveSourceAttributes(t *testing.T) {
...
+			metricFunc: exampleIntGaugeMetric,
+		},
+	}
 
-	metrics := metricPairToMetrics([]metricPair{
-		exampleIntMetric(),
-		exampleIntGaugeMetric(),
-	})
```
sumo-drosiek, line 758: This test was design to verify sending multiple different metrics in one request AFAIR

sumo-drosiek: Could you add `sum and gauge` test as well?

pkg/exporter/sumologicexporter/graphite_formatter_test.go
```diff
@@ -225,7 +226,7 @@ func TestGraphiteMetrics(t *testing.T) {
 			gf, err := newGraphiteFormatter(tt.formatter)
 			require.NoError(t, err)
 
-			result := gf.metric2String(tt.metric)
+			result := gf.metric2String(tt.metricFunc(false))
```
sumo-drosiek, line 229: shouldnt be the same like in carbon_formatter?

pkg/exporter/sumologicexporter/sender.go
```diff
@@ -519,59 +510,65 @@ func (s *sender) sendOTLPLogs(ctx context.Context, ld pdata.Logs) error {
...
 		return nil, fmt.Errorf("Attempting to send OTLP metrics as non-OTLP data")
 	}
 
 	var (
-		body           bodyBuilder
+		body           bodyBuilder = newBodyBuilder()
 		errs           []error
-		droppedRecords []metricPair
-		currentRecords []metricPair
+		droppedRecords []pdata.Metric
```
sumo-drosiek, line 521: what about resource attributes?

sumo-drosiek (commented):
I don't see a code where we translate record attributes. Do we decided to skip translation for them?

pkg/exporter/sumologicexporter/exporter_test.go
pmalek-sumo: The point of this is that we'll get this third argument in test failure output without the need to look at the code/line number.

I agree but at the same time I didn't want to change ALL TEST THE CODE. Perhaps an improvement for another time?

pkg/exporter/sumologicexporter/graphite_formatter_test.go
pmalek-sumo: What do you mean?

pkg/exporter/sumologicexporter/sender.go
pmalek-sumo: We're sending metrics from a particular resource (see the argument to this func: `pdata.ResourceMetrics`) so no need to set its attributes aside because we'll have it readily available at the call site.

https://github.com/SumoLogic/sumologic-otel-collector/blob/726bdaa6960e7710efd8c4c85a156846407e3cef/pkg/exporter/sumologicexporter/exporter.go#L334-L337

pkg/exporter/sumologicexporter/exporter_test.go
pmalek-sumo: I didn't like the left shift 😿

pkg/exporter/sumologicexporter/exporter_test.go
sumo-drosiek: > The point of this is that we'll get this third argument in test failure output without the need to look at the code/line number.

Most of times I need to do it anyway :sweat_smile:

pkg/exporter/sumologicexporter/exporter_test.go
sumo-drosiek: I don't like any of them, but left shift was more readable imho

pkg/exporter/sumologicexporter/graphite_formatter_test.go
sumo-drosiek: I mean we are doing `metric, attributes := tt.metricFunc()` for exactly the same problem (versus `tt.metricFunc(false)`)

sumo-drosiek:
> Do we decided to skip translation for them?

Yes, we decided to skip translation for them

pkg/exporter/sumologicexporter/graphite_formatter_test.go
pmalek-sumo: changed

swiatekm-sumo (commented):
The changes look ok to me. I'm just wandering if we should add documentation regarding the breaking change in this PR (in the changelog?) or leave it for the release notes? I'd rather do it here to be honest.

pkg/exporter/sumologicexporter/exporter_test.go
swiatekm-sumo: That test is kind of complex with the current setup and doesn't really make much practical sense imo.

docs/Upgrading.md
```diff
@@ -12,3 +12,67 @@ Please refer to the [official upgrade guide][opentelemetry-log-collection-upgrad
...
+1. `metadata_attributes` no longer exists.
+1. Metadata for sent records is based on resource-level attributes.
+
+In order to retain current behaviour, processors should be used to transform the data before it is exported. This
+potentially involves two transformations:
+
+#### Removing unnecessary metadata using the [resourceprocessor][resourceprocessor]
+
+`metadata_attributes` allowed filtering based on regular expressions. An equivalent processor doesn't yet
+exist, but resource-level attributed can be dropped using the [resourceprocessor]. For example:
```
sumo-drosiek, line 36:
```suggestion
exist, but resource-level attributed can be dropped using the [resourceprocessor][resourceprocessor]. For example:
```

sumo-drosiek: I know this takes the same effect, but keep it consistent across file

sumo-drosiek: I would even change it to `[resource processor][...]...`

sumo-drosiek (approved):
Let's ship it

PR #555, opened 2022-04-22
chore: fix local Dockerfile

swiatekm-sumo:
The local Dockerfile was missing some dependencies and builder installation steps.


sumo-drosiek (approved):
Thank you

PR #556, opened 2022-04-25
docs: add breaking change description to changelog

swiatekm-sumo:
Describe the breaking change to `opentelemetry-log-collection` in the changelog.

### DIFF
--- a/docs/Upgrading.md
+++ b/docs/Upgrading.md
@@ -1,0 +1,14 @@
+# Upgrading
+
+## Upgrading to 0.49.0-sumo-0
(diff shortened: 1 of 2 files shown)


docs/Upgrading.md
```diff
@@ -0,0 +1,12 @@
+# Upgrading to 0.49.0-sumo-0
```
sumo-drosiek, line 1: Top level header should be `# Upgrading` or sth, but not specific version

astencel-sumo (approved):

PR #557, opened 2022-04-26
chore: bump Telegraf to v1.22.0-sumo-3

swiatekm-sumo:
This is only to restore the `socket_receiver` plugin as seen in this change: https://github.com/SumoLogic/telegraf/pull/39.

Also fixed a missing link in the changelog.

### DIFF
--- a/pkg/receiver/telegrafreceiver/go.mod
+++ b/pkg/receiver/telegrafreceiver/go.mod
@@ -139,4 +139,4 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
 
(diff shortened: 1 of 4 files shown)


pkg/receiver/telegrafreceiver/go.mod
```diff
@@ -139,4 +139,4 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
 
-replace github.com/influxdata/telegraf => github.com/SumoLogic/telegraf v1.22.0-sumo-2
+replace github.com/influxdata/telegraf => github.com/SumoLogic/telegraf v1.22.0-sumo-3
```
astencel-sumo, line 142: Shouldn't `go.sum` be updated as well?

PR #561, opened 2022-04-28
fix(sumologicexporter): batch metrics if source headers match

swiatekm-sumo:
#549 changed log and metrics exporting to send each Resource in a separate batch. This behaviour is correct for logs, whose metadata is passed via a X-Sumo-Fields http header, but is wasteful for metrics, where it's passed in the body. This change makes the exporter send all metrics in a batch in a single request, unless the source headers are different.

To achieve this we change several internal interfaces:
* `sendNonOtlpMetrics` now takes and returns a `pmetric.Metrics`, like the tracing equivalent
* the metrics sender now operates on resources, instead of individual datapoints
* removed some superfluous error handling around string builders

### DIFF
--- a/pkg/exporter/sumologicexporter/exporter_test.go
+++ b/pkg/exporter/sumologicexporter/exporter_test.go
@@ -818,21 +818,14 @@ gauge_metric_name{test="test_value",test2="second_value",remote_name="156955",ur
 			expectedError: "failed sending data: status: 500 Internal Server Error",
 		},
 		{
-			name: "sent separately when metrics under different resources",
+			name: "sent together when metrics under different resources",
 			callbacks: []func(w http.ResponseWriter, req *http.Request){
 				func(w http.ResponseWriter, req *http.Request) {
 					w.WriteHeader(500)
 
 					body := extractBody(t, req)
-					expected := `test.metric.data{test="test_value",test2="second_value"} 14500 1605534165000`
-					assert.Equal(t, expected, body)
-					assert.Equal(t, "application/vnd.sumologic.prometheus", req.Header.Get("Content-Type"))
-				},
-				func(w http.ResponseWriter, req *http.Request) {
-					w.WriteHeader(500)
-
-					body := extractBody(t, req)
-					expected := `gauge_metric_name{foo="bar",remote_name="156920",url="http://example_url"} 124 1608124661166
+					expected := `test.metric.data{test="test_value",test2="second_value"} 14500 1605534165000
+gauge_metric_name{foo="bar",remote_name="156920",url="http://example_url"} 124 1608124661166
 gauge_metric_name{foo="bar",remote_name="156955",url="http://another_url"} 245 1608124662166`
 					assert.Equal(t, expected, body)
 					assert.Equal(t, "application/vnd.sumologic.prometheus", req.Header.Get("Content-Type"))
@@ -853,7 +846,7 @@ gauge_metric_name{foo="bar",remote_name="156955",url="http://another_url"} 245 1
 				)
 				return metrics
 			},
-			expectedError: "failed sending data: status: 500 Internal Server Error (x2)",
+			expectedError: "failed sending data: status: 500 Internal Server Error",
 		},
 	}
 
--- a/pkg/exporter/sumologicexporter/sender.go
+++ b/pkg/exporter/sumologicexporter/sender.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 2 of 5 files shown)


pkg/exporter/sumologicexporter/exporter_test.go
```diff
@@ -818,21 +818,14 @@ gauge_metric_name{test="test_value",test2="second_value",remote_name="156955",ur
 			expectedError: "failed sending data: status: 500 Internal Server Error",
 		},
 		{
-			name: "sent separately when metrics under different resources",
+			name: "sent together when metrics under different resources",
```
sumo-drosiek, line 821: :D

pkg/exporter/sumologicexporter/sender.go
```diff
@@ -98,19 +92,30 @@ func (b *bodyBuilder) Reset() {
 }
 
 // addLine adds line to builder and increments counter
-func (b *bodyBuilder) addLine(line string) error {
-	_, err := b.builder.WriteString(line)
-	if err != nil {
-		return err
-	}
+func (b *bodyBuilder) addLine(line string) {
+	b.builder.WriteString(line) // WriteString can't actually return an error
```
sumo-drosiek, line 96: https://go.dev/src/strings/builder.go
```
// WriteString appends the contents of s to b's buffer.

// It returns the length of s and a nil error.

func (b *Builder) WriteString(s string) (int, error) {

	b.copyCheck()

	b.buf = append(b.buf, s...)

	return len(s), nil

}
```

pkg/exporter/sumologicexporter/sender.go
```diff
@@ -98,19 +92,30 @@ func (b *bodyBuilder) Reset() {
...
+// addLine adds multiple lines to builder and increments counter
+func (b *bodyBuilder) addLines(lines []string) {
+	if len(lines) == 0 {
+		return
+	}
+
+	// add the first line separately to avoid a conditional in the loop
+	b.builder.WriteString(lines[0])
+
+	for _, line := range lines[1:] {
```
sumo-drosiek, line 109: just checking, does it work as expected for one item list?

pkg/exporter/sumologicexporter/sender.go
```diff
@@ -609,44 +624,35 @@ func (s *sender) sendOTLPMetrics(ctx context.Context, md pmetric.Metrics) error
...
-		if err := s.send(ctx, pipeline, body.toCountingReader(), flds); err != nil {
-			errors = append(errors, err)
-		}
+) (sent bool, err error) {
+	var (
+		linesTotalLength int = 0
+	)
+
+	for _, line := range lines {
+		linesTotalLength += len(line)
```
sumo-drosiek, line 639: what about newline length?

pkg/exporter/sumologicexporter/sender.go
swiatekm-sumo: Yes, that's what I mean, there's no point to checking the error.

pkg/exporter/sumologicexporter/sender.go
swiatekm-sumo: Yeah, it's an empty slice then and we don't do anything.

pkg/exporter/sumologicexporter/sender.go
swiatekm-sumo: Good point, will fix.

pkg/exporter/sumologicexporter/sender.go
```diff
@@ -609,44 +624,35 @@ func (s *sender) sendOTLPMetrics(ctx context.Context, md pmetric.Metrics) error
 // appendAndSend appends line to the request body that will be sent and sends
 // the accumulated data if the internal logBuffer has been filled (with maxBufferSize elements).
 // It returns appendResponse
-func (s *sender) appendAndSend(
+func (s *sender) appendAndMaybeSend(
 	ctx context.Context,
-	line string,
+	lines []string,
```
sumo-drosiek, line 629: This is going to be an issue if `lines` are going to be a lot of metrics (way more than reasonable value). Not sure if it's real case, but in theory it's possible

pkg/exporter/sumologicexporter/sender.go
sumo-drosiek: I just pointed it for reference :+1:

pkg/exporter/sumologicexporter/sender.go
swiatekm-sumo: It can be at most as many as we can have in a single Resource, and we have the whole batch in memory anyway. So at most we use 2x the memory in the most degenerate case.

swiatekm-sumo: I think that's something that needs to be configured in the batch processor. `otlpexporter`, for example, doesn't expose any settings like this, it simply sends what it gets as-is, and that's what we want to use in the future. You can't send more records that are present in the batch here, and your configuration should be able to deal with this either way.

sumo-drosiek: It makes sense. Also for fluentd the batches were stored in separate files.

sumo-drosiek (approved):

PR #567, opened 2022-05-06
style: use VS Code file nesting to nest test files

andrzej-stencel:


astencel-sumo:
Closing this as we don't want this as the default. Anybody can set this by themselves either locally (for workspace) or globally.

PR #578, opened 2022-05-18
fix(k8sprocessor)!: deprecate clusterName setting

swiatekm-sumo:
The respective ObjectMeta field is being removed from Kubernetes. It's practically never set and only confuses users. It was also removed from the upstream processors: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/9886/.

This is technically a breaking change, but users shouldn't see any difference in practice.


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Breaking changes
+
+- fix(k8sprocessor)!: deprecate clusterName setting [#578]
```
kkujawa-sumo, line 12: If I see correctly, you are just removing `clusterName`, it is no longer used.
For me deprecation is when something can be used, it is not recommended to use it and it will be removed in the future.

kkujawa-sumo (approved):

CHANGELOG.md
swiatekm-sumo: I'm keeping the setting available, with a warning if it's used, it simply doesn't do anything anymore. So users don't have to update their configuration immediately, but the behaviour does change.

PR #588, opened 2022-05-20
chore(ci): add pipelines to test and build on Go+BoringCrypto

kkujawa-sumo:


sumo-drosiek:
Is it possible to build the FIPS complianat image locally? I think we should have separate `Dockerfile` or `ci` possibility to do that

kkujawa-sumo:
> In our CI binaries are built outside of the image and copied to the image but now I've noticed that in dev pipeline container image bases on [go image](https://github.com/SumoLogic/sumologic-otel-collector/blob/99d71f9aee9fcba3f6a4f1b7a441e4d2f5e63d5c/Dockerfile_dev#L13) so it would be good to have appropriate version of go in the containers for fips binaries.
> 
> Yes, we need to have Dockerfile to build container image locally for FIPS binaries, similar to [this](https://github.com/SumoLogic/sumologic-otel-collector/blob/main/Dockerfile_local) one.
> 
> so there are two things to do :)

This will be done in another PR.

sumo-drosiek (dismissed):

sumo-drosiek:
I see that `otelcol-sumo-darwin_arm64` do not have fips version

PR #596, opened 2022-05-31
feat(sumoicmarshaler): adding sumo ic marshaler

pdelewski:
Sumo Installed Collector marshaler takes otel log model and translate it into Installed Collector schema


pkg/translator/sumoicmarshaler/sumo_ic_marshaler.go
```diff
@@ -0,0 +1,133 @@
...
+	return marshaler.format
+}
+
+func NewSumoICMarshaler() SumoICMarshaler {
+	return SumoICMarshaler{}
+}
+
+func logEntry(buf *bytes.Buffer, format string, a ...interface{}) {
+	buf.WriteString(fmt.Sprintf(format, a...))
+	buf.WriteString("\n")
```
sumo-drosiek, line 43: We can use `WriteByte` here

pkg/translator/sumoicmarshaler/sumo_ic_marshaler.go
```diff
@@ -0,0 +1,133 @@
...
+	case pcommon.ValueTypeInt:
+		return strconv.FormatInt(v.IntVal(), 10)
+	case pcommon.ValueTypeSlice:
+		return sliceToString(v.SliceVal())
+	case pcommon.ValueTypeMap:
+		return mapToString(v.MapVal())
+	default:
+		return fmt.Sprintf("<Unknown OpenTelemetry attribute value type %q>", v.Type())
+	}
+}
```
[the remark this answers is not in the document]
sumo-drosiek: `AsString()` I suppose

pkg/translator/sumoicmarshaler/sumo_ic_marshaler.go
```diff
@@ -0,0 +1,133 @@
...
+
+func mapToString(m pcommon.Map) string {
+	var b strings.Builder
+	b.WriteString("{\n")
+
+	m.Sort().Range(func(k string, v pcommon.Value) bool {
+		fmt.Fprintf(&b, "     -> %s: %s(%s)\n", k, v.Type(), v.AsString())
+		return true
+	})
```
sumo-drosiek, line 87: Could you add UT for this code? Does IC really marshals maps that way?

sumo-drosiek: I believe we should use `json` format here

pdelewski: As we can use AsString() function to stringify attributeValue above function is no longer needed

pkg/translator/sumoicmarshaler/sumo_ic_marshaler_test.go
```diff
@@ -0,0 +1,102 @@
...
+	rls.Resource().Attributes().InsertString("_sourceHost", "testHost")
+
+	sl := rls.ScopeLogs().AppendEmpty()
+	const recordNum = 0
+
+	ts := pcommon.Timestamp(int64(recordNum) * time.Millisecond.Nanoseconds())
+	logRecord := sl.LogRecords().AppendEmpty()
+	logRecord.Body().SetStringVal("entry1")
+	logRecord.Attributes().InsertString("log.file.path_resolved", "testSourceName")
+	logRecord.SetTimestamp(ts)
```
sumo-drosiek, line 93: I would go with map containing all possible types. Can be done in this or separate test

pdelewski: ok

pkg/translator/sumoicmarshaler/sumo_ic_marshaler.go
```diff
@@ -0,0 +1,84 @@
...
+			for k := 0; k < logs.Len(); k++ {
+				lr := logs.At(k)
+				dateVal := lr.ObservedTimestamp()
+				body := lr.Body().AsString()
+				sourceName, exists := lr.Attributes().Get(SourceNameKey)
+				if exists == false {
+					return nil, errors.New("_sourceName attribute does not exists")
+				}
+				logEntry(&buf, "{\"data\": \"%s\",\"sourceName\":\"%s\",\"sourceHost\":\"%s\",\"sourceCategory\":\"%s\",\"fields\":{},\"message\":\"%s\"}",
+					dateVal, sourceName.AsString(), sourceHost.AsString(), sourceCategory.AsString(), body)
```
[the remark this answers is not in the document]
pdelewski: Yes, this is marshalling to json. We can follow your suggestion and marshal a struct. I just wanted to avoid this additional transformation, however it might be cleaner. Regarding `fields` I was not sure what be there, so for now it's a placeholder

pkg/translator/sumoicmarshaler/sumo_ic_marshaler.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"encoding/json"
+	"errors"
+
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+type SumoICMarshaler struct {
+	format string
```
swiatekm-sumo, line 26: Is this actually used for anything?

pkg/translator/sumoicmarshaler/sumo_ic_marshaler.go
```diff
@@ -0,0 +1,106 @@
...
+	SourceCategoryKey = "_sourceCategory"
+	SourceHostKey     = "_sourceHost"
+	SourceNameKey     = "log.file.path_resolved"
+)
+
+type field struct {
+}
+
+type icRecord struct {
+	Date           string `json:"data"`
```
swiatekm-sumo, line 47:
```suggestion
	Data           string `json:"data"`
```

pkg/translator/sumoicmarshaler/sumo_ic_marshaler.go
```diff
@@ -0,0 +1,106 @@
...
+	for i := 0; i < rls.Len(); i++ {
+		rl := rls.At(i)
+		sourceCategory, exists := rl.Resource().Attributes().Get(SourceCategoryKey)
+		if exists == false {
+			return nil, errors.New("_sourceCategory attribute does not exists")
+		}
+		sourceHost, exists := rl.Resource().Attributes().Get(SourceHostKey)
+		if exists == false {
+			return nil, errors.New("_sourceHost attribute does not exists")
+		}
```
swiatekm-sumo, line 67: Are you sure this is the right behaviour? I don't think there's a general requirement to set sourceHost and sourceCategory - the Sumo backend will accept log records without them afaik.

pkg/translator/sumoicmarshaler/sumo_ic_marshaler.go
```diff
@@ -0,0 +1,106 @@
...
+			logs := ils.LogRecords()
+			for k := 0; k < logs.Len(); k++ {
+				lr := logs.At(k)
+				sourceName, exists := lr.Attributes().Get(SourceNameKey)
+				if exists == false {
+					return nil, errors.New("_sourceName attribute does not exists")
+				}
+				icRecords = append(icRecords, icRecord{lr.ObservedTimestamp().String(),
+					sourceName.AsString(), sourceHost.AsString(),
+					sourceCategory.AsString(), field{}, lr.Body().AsString()})
```
swiatekm-sumo, line 80: Do we know what the format of `fields` should be? I'm pretty sure I know the values - it should be `rl.Attributes()`.

pkg/translator/sumoicmarshaler/sumo_ic_marshaler.go
pdelewski: It's used in s3exporter, for now I can remove it

pkg/translator/sumoicmarshaler/sumo_ic_marshaler.go
pdelewski: Of course, I'm not sure. That was just my assumption. If it's not correct I will remove it

sumo-drosiek: its optional AFAIR

pkg/translator/sumoicmarshaler/sumo_ic_marshaler.go
```diff
@@ -0,0 +1,112 @@
...
+	icRecords := []icRecord{}
+	rls := ld.ResourceLogs()
+
+	for i := 0; i < rls.Len(); i++ {
+		rl := rls.At(i)
+		sourceCategoryVal := ""
+		sourceCategory, exists := rl.Resource().Attributes().Get(SourceCategoryKey)
+		if exists {
+			sourceCategoryVal = sourceCategory.AsString()
+		}
```
pmm-sumo, line 61: Slightly more idiomatic way:

```suggestion
		if sourceCategory, exists := rl.Resource().Attributes().Get(SourceCategoryKey); exists {
			sourceCategoryVal = sourceCategory.AsString()
		}
```

pkg/translator/sumoicmarshaler/sumo_ic_marshaler.go
```diff
@@ -0,0 +1,112 @@
...
+				sourceName, exists := lr.Attributes().Get(SourceNameKey)
+				if exists {
+					sourceNameVal = sourceName.AsString()
+				}
+				fields := fieldSet{attributes: make(map[string]interface{})}
+				attrs := lr.Attributes().AsRaw()
+				for ak, av := range attrs {
+					fields.attributes[ak] = av
+				}
+
```
pmm-sumo, line 83: So this is not how we currently process logs when ingesting them through OTLP source. We should be doing following:

1) Fields should be taken from resource-level attributes
2) Body should be based on logrecord-level attributes + logrecord body (if body is a raw string, it should be stored under a separate key)

Here is more detail: https://github.com/SumoLogic/sumologic-otel-collector/blob/main/docs/OpenTelemetryConcepts.md

pdelewski: Actually, I don't understand 2 point, should body be concatenated with logrecord level attributes?

Log record from Installed Collector looks as follows (body is under message key):

`{"data": "2022-05-30 08:20:54.133382 +0000 UTC","sourceName":"/Users/pdelewski/Projects/logwriter/data.log","sourceHost":"pdelewski-mac","sourceCategory":"logfile","fields":{},"message":"2022-05-26 14:15:13.115274 +0200 CEST m=+9807.933100239 log entry9800"}`

Let's say that we have log record attributes, how then above log record should look like?

astencel-sumo:
Closing as stale.

Also I think this is now available in upstream in the [AWS S3 exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.94.0/exporter/awss3exporter/README.md) as the `sumo_ic` marshaler.

PR #611, opened 2022-06-10
feat: build arm64 binary for darwin

sumo-drosiek:


perk-sumo (approved):
I've tested the binary and it works OK

PR #614, opened 2022-06-13
feat: add missing core receivers, processors and extensions

sumo-drosiek:


swiatekm-sumo (approved):

sumo-drosiek:
I will, I was thinking also about auotomating it, but as I use VSCode, LibreOffice, grep, copy and paste, this doesn't seem to be legit :sweat_smile:

docs/release.md
```diff
@@ -144,6 +145,92 @@ make gotest
 
 in the repository root.
 
+## Add missing upstream components
```
sumo-drosiek, line 148: @swiatekm-sumo Please look at it :fearful:

sumo-drosiek: I updated it. This is valid point about using changelogs of upstream projects :+1:

swiatekm-sumo (approved):

PR #616, opened 2022-06-13
fix(cascading_filter): fix leak on traces where decision was made

pmm-sumo:
This fixes the case with increased memory consumption when spans for traces where decision was already made are observed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,12 +13,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - feat: add logstransformprocessor from upstream [#604]
 - feat: build arm64 binary for darwin [#611]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


sumo-drosiek:
I made some calculations and test should take about 20s, is it possible to reduce it somehow?

pmm-sumo:
> I made some calculations and test should take about 20s, is it possible to reduce it somehow?

So that's the bummer with it, the shortest interval for compaction to kick is 1s. With margin, this gets bigger. I'm thinking about either disabling the test or changing the internal API to accept shorter intervals

sumo-drosiek:
We can add some comment to the test and if it will be too annoying we can fix it then. Of course you can fix it now as well :)

pmm-sumo:
@sumo-drosiek the fix won't be that simple so I skipped the test for now. I am keen to rework the underlying architecture of the cascading filter and then we could fix that as well

PR #620, opened 2022-06-20
feat(rawk8seventsreceiver): remember last processed resource version

astencel-sumo:


pkg/receiver/rawk8seventsreceiver/config.go
```diff
@@ -38,6 +38,8 @@ type Config struct {
 
 	// ConsumeMaxRetries is the maximum number of retries for recoverable pipeline errors
 	ConsumeMaxRetries uint64 `mapstructure:"consume_max_retries"`
+
+	ResourceVersion uint64 `mapstructure:"resource_version"`
```
swiatekm-sumo, line 42: Do we want to make this configurable? Sounds like a footgun.

astencel-sumo: No, we definitely don't. This is a work in progress, far from the final version.

pkg/receiver/rawk8seventsreceiver/receiver.go
```diff
@@ -144,6 +160,64 @@ func (r *rawK8sEventsReceiver) Shutdown(context.Context) error {
...
+	if startResourceVersionBytes == nil {
+		r.logger.Info("Start resource version not found in storage")
+		return 0, nil
+	}
+
+	startResourceVersionString := string(startResourceVersionBytes)
+	startResourceVersion, err := strconv.ParseUint(startResourceVersionString, 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse start resource version '%s' to number: %s", startResourceVersionString, err)
+	}
```
swiatekm-sumo, line 215: Looks kind of complicated for reading an int from a KV store. Maybe use https://pkg.go.dev/encoding/binary instead? Probably faster as well.

pkg/receiver/rawk8seventsreceiver/receiver.go
```diff
@@ -44,13 +47,15 @@ var severityMap = map[string]plog.SeverityNumber{
...
-	startTime        time.Time
+	cfg                  *Config
+	client               k8s.Interface
+	eventControllers     []cache.Controller
+	eventCh              chan *eventChange
+	ctx                  context.Context
+	cancel               context.CancelFunc
+	startTime            time.Time
+	storage              storage.Client
+	startResourceVersion uint64
```
swiatekm-sumo, line 58: I think a better name for this would be `latestResourceVersion` or something similar. As is, we're writing to `startResourceVersion` for each processed record, which is unintuitive.

pkg/receiver/rawk8seventsreceiver/receiver.go
```diff
@@ -188,6 +262,7 @@ func (r *rawK8sEventsReceiver) processEventChangeLoop() {
 // this includes: checking if we should process the event, converting it into a plog.Logs
 // and sending it to the next consumer in the pipeline
 func (r *rawK8sEventsReceiver) processEventChange(ctx context.Context, eventChange *eventChange) {
+	r.recordEventProcessed(eventChange.event)
```
swiatekm-sumo, line 265: This should only happen after `consumeWithRetry` returns without error. Otherwise we risk losing events in case of errors from the consumer.

pkg/receiver/rawk8seventsreceiver/receiver.go
```diff
@@ -144,6 +160,64 @@ func (r *rawK8sEventsReceiver) Shutdown(context.Context) error {
...
+	return storageClient, nil
+}
+
+func (r *rawK8sEventsReceiver) getStartResourceVersion(ctx context.Context) (uint64, error) {
+	if r.storage == nil {
+		r.logger.Info("Did not find start resource version, as there is no storage.")
+		return 0, nil
+	}
+
+	startResourceVersionBytes, err := r.storage.Get(ctx, "startResourceVersion")
```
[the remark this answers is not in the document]
astencel-sumo: I agree

pkg/receiver/rawk8seventsreceiver/receiver.go
astencel-sumo: Using the `binary` package doesn't make things better, in fact I would argue it makes things worse as there is even more boilerplate code:

```patch
---
 pkg/receiver/rawk8seventsreceiver/receiver.go | 24 +++++++++++++++----
 1 file changed, 19 insertions(+), 5 deletions(-)

diff --git a/pkg/receiver/rawk8seventsreceiver/receiver.go b/pkg/receiver/rawk8seventsreceiver/receiver.go
index 43cfbbdb..8ed54e16 100644
--- a/pkg/receiver/rawk8seventsreceiver/receiver.go
+++ b/pkg/receiver/rawk8seventsreceiver/receiver.go
@@ -16,6 +16,7 @@ package rawk8seventsreceiver
 
 import (
 	"context"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"strconv"
@@ -214,10 +215,10 @@ func (r *rawK8sEventsReceiver) getLatestResourceVersion(ctx context.Context) (ui
 		return 0, nil
 	}
 
-	latestResourceVersionString := string(latestResourceVersionBytes)
-	latestResourceVersion, err := strconv.ParseUint(latestResourceVersionString, 10, 64)
-	if err != nil {
-		return 0, fmt.Errorf("failed to parse latest resource version '%s' to number: %s", latestResourceVersionString, err)
+	latestResourceVersion, length := binary.Uvarint(latestResourceVersionBytes)
+	if length <= 0 {
+		r.logger.Warn("failed to parse latest resource version", zap.Any("length", length))
+		return 0, nil
 	}
 
 	r.logger.Info("Found latest resource version in storage", zap.Any("latest_resource_version", latestResourceVersion))
@@ -293,7 +294,20 @@ func (r *rawK8sEventsReceiver) recordEventProcessed(event *corev1.Event) {
 		return
 	}
 
-	r.storage.Set(r.ctx, latestResourceVersionStorageKey, []byte(event.ResourceVersion))
+	resourceVersionNumber, err := strconv.ParseUint(event.ResourceVersion, 10, 64)
+	if err != nil {
+		r.logger.Debug("Cannot record incoming resource version, not a number.", zap.Any("incoming_resource_version", event.ResourceVersion))
+		return
+	}
+
+	bytes := make([]byte, binary.MaxVarintLen64)
+	length := binary.PutUvarint(bytes, resourceVersionNumber)
+	if length <= 0 {
+		r.logger.Warn("Cannot record incoming resource version, cannot encode.", zap.Any("length", length), zap.Any("incoming_resource_version", event.ResourceVersion))
+		return
+	}
+
+	r.storage.Set(r.ctx, latestResourceVersionStorageKey, bytes[:length])
 }
 
 // Check if we should process the event.
```

pkg/receiver/rawk8seventsreceiver/receiver.go
astencel-sumo: > Otherwise we risk losing events in case of errors from the consumer.
We do risk losing events in any case. To prevent losing events, we'd need to implement a proper persistent buffer. 

Moving this line to after successful consumption of event is a substitute of a persistent buffer that will work in certain scenarios and will not work in others.

What's more, moving this line to after checking if event is accepted means we are not storing the *latest* resource version received, but rather the *highest* resource version number received. This is some scenarios can lead to data loss where the original solution would lead to data duplication.

pkg/receiver/rawk8seventsreceiver/README.md
```diff
@@ -38,4 +38,37 @@ receivers:
...
+    logs:
+      receivers:
+      - raw_k8s_events
+      exporters:
+      - nop
+```
+
 [Fluentd plugin]: https://github.com/SumoLogic/sumologic-kubernetes-fluentd/tree/main/fluent-plugin-events
+
+[event_ttl]: https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/#options
```
sumo-drosiek, line 74:
```suggestion
[event_ttl]: https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/#options
```

pkg/receiver/rawk8seventsreceiver/receiver.go
sumo-drosiek: Maybe change the name to `recordEventReceived`

pkg/receiver/rawk8seventsreceiver/README.md
astencel-sumo: I'm sorry I'm not seeing any change here, what are you suggesting?

astencel-sumo: Oh it's the empty line 😄 I see now

astencel-sumo: Done

sumo-drosiek (approved):

sumo-drosiek:
fix lint :upside_down_face:

PR #627, opened 2022-06-28
fix: fix import paths for observer extensions

astencel-sumo:
The observer extensions have their own modules.
Without this fix, the latest package versions are imported
instead of the specified ones.


sumo-drosiek (approved):
Thanks!

PR #628, opened 2022-06-28
ci: disable MacOS builds while signing not possible

astencel-sumo:
There's an infrastructure issue with signing the MacOS binaries.
Until it is resolved, let's release new versions without MacOS binaries.


sumo-drosiek:
Please add appropriate information to Changelog

sumo-drosiek (approved):

PR #630, opened 2022-06-28
docs(release): describe fixing a failed release

astencel-sumo:


docs/release.md
```diff
@@ -50,6 +50,20 @@ make add-tag push-tag
...
+Pushing a new version tag to GitHub starts the [release build](../.github/workflows/release_builds.yml) jobs.
+
+If one of these jobs fails for whatever reason (real world example: failing to notarize the MacOS binary),
+you might need to remove the created tags, perhaps change something, and create the tags again.
+
+To delete the tags both locally and remotely, run the following commands:
+
+```shell
+export TAG=v0.51.0-sumo-0
+make delete-tag delete-remote-tag
```
pdelewski, line 64: Should not be just `make delete-remote-tag` ?

sumo-drosiek: will that work if `delete-tag` fails?

astencel-sumo: Why would `delete-tag` fail? If the first target fails, the second one fails too. Not sure what scenario you have in mind.

astencel-sumo: > Should not be just `make delete-remote-tag` ?

`delete-remote-tag` deletes the tags in the remote repository, `delete-tag` deletes the tags in local repository. Not sure what is unclear here?

sumo-drosiek: tag can be remote but not necessary locally

sumo-drosiek (approved):

PR #643, opened 2022-06-30
New MySQLRecords Receiver

JKashyap96:


pkg/receiver/mysqlrecordsreceiver/README.md
```diff
@@ -0,0 +1,134 @@
+# MySQL Records Receiver
+
+This receiver queries MySQL for database records and creates a log record (plog.Logs type) for each database record.
```
sumo-drosiek, line 3:
```suggestion
This receiver queries MySQL for database records and creates a log record for each database record.
```

pkg/receiver/mysqlrecordsreceiver/README.md
```diff
@@ -0,0 +1,134 @@
...
+      telemetry:
+        logs:
+          level: debug
+    The encrypted password will only be printed in the console with a debug log level. Once generated, the user can remove the telemetry field so as to enable logging at the default info level. To use the encrypted password, the user needs to specify password_type as 'encrypted' and also the encrypt_secret_path to the same secret file.
+
+State Management Use Case:
+
+- The receiver supports saving the state of a query fetch into a csv file where a unique/auto-increment field is present in a table of a database.
+- The unique/auto-increment field can either be of type 'NUMBER' or 'TIMESTAMP', where a 'NUMBER' should be a non-negative integer and a 'TIMESTAMP' should be of the     default timestamp storage format in mysql, i.e. '2006-01-02 15:04:05'.
+- This is basically the delta mode state management feature of the receiver where the current value/state of the unique/auto-increment field is saved in a csv file which can be retreived later so as to fetch records after the saved state value.
```
sumo-drosiek, line 23: I would add headers for the use cases to group it somehow

pkg/receiver/mysqlrecordsreceiver/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+- The receiver supports password encryption for 'BasicAuth' authentication_mode
+- To generate an encrypted password:
+  - Specify encrypt_secret_path to a secret file containing a 24 character string
+  - Include the following in the otel config:
+    service:
+      telemetry:
+        logs:
+          level: debug
+    The encrypted password will only be printed in the console with a debug log level. Once generated, the user can remove the telemetry field so as to enable logging at the default info level. To use the encrypted password, the user needs to specify password_type as 'encrypted' and also the encrypt_secret_path to the same secret file.
```
sumo-drosiek, line 17: Flow is weird. I need to run OTC to generate password and then I need to run it again in order to use it?

JKashyap96: For the first iteration of the run, the password will be generated. It will fetch database queries too if required, but the encrypted password won't be used. The encrypted password can be used from the second iteration if required via the user passing the password in the config file.

pkg/receiver/mysqlrecordsreceiver/client.go
```diff
@@ -0,0 +1,343 @@
...
+//2. With an encrypted plaintext password
+//3. With an AWS Authentication token to be used as a password
+func newMySQLClient(conf *Config, logger *zap.Logger) client {
+	var port string
+	var basicauthpassword string
+	var connStr string
+	basicauthpassword = conf.Password
+	//Encrypting a plaintext password if a 24 character secret string is provided by the user from an external file
+	if (len(conf.PasswordType) == 0 || conf.PasswordType == "plaintext") && len(conf.EncryptSecretPath) != 0 {
+		MySecret1, err := readMySecret(conf)
```
sumo-drosiek, line 91:
```suggestion
		secret, err := readMySecret(conf)
```

I don't see a reason to use `My` prefix

also variables in functions should starting with lower case

pkg/receiver/mysqlrecordsreceiver/README.md
```diff
@@ -0,0 +1,134 @@
...
+BasicAuth Password Encryption Use Case:
+
+- The receiver supports password encryption for 'BasicAuth' authentication_mode
+- To generate an encrypted password:
+  - Specify encrypt_secret_path to a secret file containing a 24 character string
+  - Include the following in the otel config:
+    service:
+      telemetry:
+        logs:
+          level: debug
```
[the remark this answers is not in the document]
JKashyap96: will modify this.

pkg/receiver/mysqlrecordsreceiver/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+This receiver queries MySQL for database records and creates a log record (plog.Logs type) for each database record.
+
+Supported pipeline types: logs
+
+BasicAuth Password Encryption Use Case:
+
+- The receiver supports password encryption for 'BasicAuth' authentication_mode
+- To generate an encrypted password:
+  - Specify encrypt_secret_path to a secret file containing a 24 character string
```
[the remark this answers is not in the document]
JKashyap96: will modify this.

pkg/receiver/mysqlrecordsreceiver/README.md
JKashyap96: will modify this.

pkg/receiver/mysqlrecordsreceiver/client.go
```diff
@@ -0,0 +1,343 @@
...
+//1. With a plaintext password
+//2. With an encrypted plaintext password
+//3. With an AWS Authentication token to be used as a password
+func newMySQLClient(conf *Config, logger *zap.Logger) client {
+	var port string
+	var basicauthpassword string
+	var connStr string
+	basicauthpassword = conf.Password
+	//Encrypting a plaintext password if a 24 character secret string is provided by the user from an external file
+	if (len(conf.PasswordType) == 0 || conf.PasswordType == "plaintext") && len(conf.EncryptSecretPath) != 0 {
```
sumo-drosiek, line 90: `len(conf.PasswordType) == 0` this should be resolved by default configuration, or config verification

`len(conf.EncryptSecretPath) != 0` this should be resolved by config verification

pkg/receiver/mysqlrecordsreceiver/client.go
sumo-drosiek: adding number to variable is not so informative

pkg/receiver/mysqlrecordsreceiver/client.go
```diff
@@ -0,0 +1,343 @@
...
+		}
+		basicauthpassword = decText
+	}
+	//Considering default port to be 3306 if not provided by user
+	if len(conf.DBPort) != 0 {
+		port = conf.DBPort
+	} else {
+		logger.Info("dbport empty, considering default 3306")
+		port = "3306"
+	}
```
sumo-drosiek, line 119: This should be resolved by defaults

pkg/receiver/mysqlrecordsreceiver/client.go
```diff
@@ -0,0 +1,343 @@
...
+	endpoint := conf.DBHost + ":" + port
```
sumo-drosiek, line 120: I there a need to specify port and host separately?

pkg/receiver/mysqlrecordsreceiver/client.go
```diff
@@ -0,0 +1,343 @@
...
+	if conf.AuthenticationMode == "IAMRDSAuth" {
```
sumo-drosiek, line 121: Please use constants for strings in control flow

pkg/receiver/mysqlrecordsreceiver/client.go
```diff
@@ -0,0 +1,343 @@
...
+	} else {
+		driverConf := mysql.Config{
+			User:                 conf.Username,
+			Passwd:               basicauthpassword,
+			Net:                  conf.Transport,
+			Addr:                 endpoint,
+			DBName:               conf.Database,
+			AllowNativePasswords: conf.AllowNativePasswords,
+		}
+		connStr = driverConf.FormatDSN()
```
sumo-drosiek, line 148: This line can be move outside of if/else as is the same in both cases

pkg/receiver/mysqlrecordsreceiver/client.go
```diff
@@ -0,0 +1,343 @@
...
+	} else {
+		driverConf := mysql.Config{
+			User:                 conf.Username,
+			Passwd:               basicauthpassword,
+			Net:                  conf.Transport,
+			Addr:                 endpoint,
+			DBName:               conf.Database,
+			AllowNativePasswords: conf.AllowNativePasswords,
+		}
```
sumo-drosiek, line 147: I would consider making common `driverConf` object, and just set distinctive properties separately

pkg/receiver/mysqlrecordsreceiver/README.md
sumo-drosiek: Is it possible to generate password externally?

pkg/receiver/mysqlrecordsreceiver/client.go
JKashyap96: This isn't some sort of validation, so included it here. It is used as encryption and decryption conditions.

pkg/receiver/mysqlrecordsreceiver/client.go
JKashyap96: yes

pkg/receiver/mysqlrecordsreceiver/client.go
JKashyap96: will do.

pkg/receiver/mysqlrecordsreceiver/client.go
JKashyap96: not really but the separation gives us more granularity.

pkg/receiver/mysqlrecordsreceiver/client.go
JKashyap96: will try to implement that.

pkg/receiver/mysqlrecordsreceiver/client.go
JKashyap96: Yes.

pkg/receiver/mysqlrecordsreceiver/client.go
JKashyap96: will modify this.

pkg/receiver/mysqlrecordsreceiver/client.go
sumo-drosiek: `conf.PasswordType` should be always set here, if user do not pass anything, we should take and set default value

if `conf.PasswordType == "plaintext"` we should ensure that `len(conf.EncryptSecretPath) != 0`, this is validation

at this point we should only have `if conf.PasswordType == "plaintext" {` here

pkg/receiver/mysqlrecordsreceiver/README.md
JKashyap96: will try to figure out.

pkg/receiver/mysqlrecordsreceiver/client.go
sumo-drosiek: Please move it to `Validate` as already you do with similar conditions:

https://github.com/SumoLogic/sumologic-otel-collector/blob/b707d591de141e928c04077abe5cee7bf4f6f2e7/pkg/receiver/mysqlrecordsreceiver/config.go#L74-L78

pkg/receiver/mysqlrecordsreceiver/client.go
sumo-drosiek: Do we support connection strings?
e.g. `mysql://user@localhost:3306`

pkg/receiver/mysqlrecordsreceiver/client.go
JKashyap96: okay

pkg/receiver/mysqlrecordsreceiver/client.go
JKashyap96: nope, we build the connection string, that's how it generally is.

sumo-drosiek: Why do we do that?

I personally prefer to have connection string in configuration
I believe it is easier for developer as well as for user

JKashyap96: I believe building the conn str gives us a larger window for feature implementation as developers, saw examples in our present python db integrator as well as the open contrib collector.

pkg/receiver/mysqlrecordsreceiver/client.go
JKashyap96: The reason i included len(conf.PasswordType) == 0 is because there might be a user entry of only an encrypt secret file, without any passwordType, then the program should produce an encrypted password as well. So, I can't remove this condition.

astencel-sumo:
Superseded by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/20730.

PR #659, opened 2022-07-07
fix(sumologicexporter): translate Telegraf metrics with OTLP format

astencel-sumo:
This change fixes a regression introduced in https://github.com/SumoLogic/sumologic-otel-collector/pull/536 that caused the metric name translations (the feature under the `translate_telegraf_attributes` setting) to not work when using the OTLP format.
This change refactors the code so that all modifications to metrics (attribute translation, metrics translation, dropping routing attribute) are done before branching the code into different sending formats.

### DIFF
--- a/pkg/exporter/sumologicexporter/sender.go
+++ b/pkg/exporter/sumologicexporter/sender.go
@@ -596,11 +596,6 @@ func (s *sender) sendOTLPMetrics(ctx context.Context, md pmetric.Metrics) error 
 	for i := 0; i < rms.Len(); i++ {
 		rm := rms.At(i)
 
-		if s.config.TranslateAttributes {
-			translateAttributes(rm.Resource().Attributes()).
-				CopyTo(rm.Resource().Attributes())
-		}
-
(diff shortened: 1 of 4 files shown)


pkg/exporter/sumologicexporter/sender.go
```diff
@@ -596,11 +596,6 @@ func (s *sender) sendOTLPMetrics(ctx context.Context, md pmetric.Metrics) error
 	for i := 0; i < rms.Len(); i++ {
 		rm := rms.At(i)
 
-		if s.config.TranslateAttributes {
-			translateAttributes(rm.Resource().Attributes()).
-				CopyTo(rm.Resource().Attributes())
-		}
-
 		s.addSourceResourceAttributes(rm.Resource().Attributes())
```
sumo-drosiek, line 599: Can we move it as well. I don't see a point to make multiple iterations with this change

sumo-drosiek: How is the performance change after this pr?

sumo-drosiek (approved):

PR #661, opened 2022-07-11
feat(sumologicexporter): do not retry on '400 Bad Request' response

astencel-sumo:
If the server responds with '400 Bad Request', the exporter should not retry the request.
This is in line what the upstream OTLP HTTP exporter does:
https://github.com/open-telemetry/opentelemetry-collector/blob/v0.55.0/exporter/otlphttpexporter/otlp.go#L181-L184

### DIFF
--- a/pkg/exporter/sumologicexporter/sender.go
+++ b/pkg/exporter/sumologicexporter/sender.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 	"time"
 
+	"go.opentelemetry.io/collector/consumer/consumererror"
 	"go.opentelemetry.io/collector/model/otlp"
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/plog"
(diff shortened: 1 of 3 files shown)


swiatekm-sumo (approved):

pkg/exporter/sumologicexporter/sender.go
```diff
@@ -312,7 +313,14 @@ func (s *sender) handleReceiverResponse(resp *http.Response) error {
 			errMsgs = append(errMsgs, fmt.Sprintf("errors: %+v", rResponse.Errors))
 		}
 
-		return fmt.Errorf("failed sending data: %s", strings.Join(errMsgs, ", "))
+		err := fmt.Errorf("failed sending data: %s", strings.Join(errMsgs, ", "))
+
+		if resp.StatusCode == http.StatusBadRequest {
+			// Report the failure as permanent if the server thinks the request is malformed.
+			return consumererror.NewPermanent(err)
+		}
```
sumo-drosiek, line 321: Shouldn't be for all 4XX?

astencel-sumo: Good question. Definitely not for all, like 401 or 429 ;) But for example 404 seems to make sense.
The OTLP spec explicitly only specified 400: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#bad-data

PR #666, opened 2022-07-13
chore(k8sprocessor): add debug logs

### LINKED ISSUE #238: [k8sprocessor] Debug logs
It would be good to have some debug logs, for example in these places:
- https://github.com/SumoLogic/sumologic-otel-collector/blob/83406585be84acccc8d9122b37fce1b37d9dd35c/pkg/processor/k8sprocessor/processor.go#L107
- https://github.com/SumoLogic/sumologic-otel-collector/blob/83406585be84acccc8d9122b37fce1b37d9dd35c/pkg/processor/k8sprocessor/pod_association.go#L91
- https://github.com/SumoLogic/sumologic-otel-collector/blob/83406585be84acccc8d9122b37fce1b37d9dd35c/pkg/processor/k8sprocessor/pod_association.go#L96
- https://github.com/SumoLogic/sumologic-otel-collector/blob/83406585be84acccc8d9122b37fce1b37d9dd35c/pkg/processor/k8sprocessor/processor.go#L126

aboguszewski-sumo:
Fixes #238

This PR adds some debug logs in the k8s processor.

### DIFF
--- a/pkg/processor/k8sprocessor/pod_association.go
+++ b/pkg/processor/k8sprocessor/pod_association.go
@@ -16,6 +16,7 @@ package k8sprocessor
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"strings"
@@ -30,7 +31,11 @@ import (
 // extractPodIds extracts IP and pod UID from attributes or request context.
 // It returns a value pair containing configured label and IP Address and/or Pod UID.
 // If empty value in return it means that attributes does not contains configured label to match resources for Pod.
-func extractPodID(ctx context.Context, attrs pcommon.Map, associations []kube.Association) (podIdentifierKey string, podIdentifierValue kube.PodIdentifier) {
+func extractPodID(
+	ctx context.Context,
+	attrs pcommon.Map,
+	associations []kube.Association,
+) (podIdentifierKey string, podIdentifierValue kube.PodIdentifier, returnErr error) {
 	connectionIP := getConnectionIP(ctx)
 	hostname := stringAttributeFromMap(attrs, conventions.AttributeHostName)
 
@@ -53,8 +58,8 @@ func extractPodID(ctx context.Context, attrs pcommon.Map, associations []kube.As
 			podIdentifierValue = kube.PodIdentifier(hostname)
 			return
 		}
-		podIdentifierKey = ""
-		return
+
+		return "", kube.PodIdentifier(""), errors.New("pod association not set, could not assign other pod id")
 	}
 
 	for _, asso := range associations {
@@ -87,20 +92,23 @@ func extractPodID(ctx context.Context, attrs pcommon.Map, associations []kube.As
 			// Build hostname from pod k8s.pod.name and k8s.namespace.name attributes
 			pod, ok := attrs.Get(conventions.AttributeK8SPodName)
 			if !ok {
-				return "", ""
+				return "", kube.PodIdentifier(""), errors.New("pod name not found in attributes")
 			}
 
 			namespace, ok := attrs.Get(conventions.AttributeK8SNamespaceName)
 			if !ok {
-				return "", ""
+				return "", kube.PodIdentifier(""), errors.New("namespace name not found in attributes")
 			}
 
 			if pod.StringVal() == "" || namespace.StringVal() != "" {
-				return asso.Name, kube.PodIdentifier(fmt.Sprintf("%s.%s", pod.StringVal(), namespace.StringVal()))
+				podIdentifierKey = asso.Name
+				podIdentifierValue = kube.PodIdentifier(fmt.Sprintf("%s.%s", pod.StringVal(), namespace.StringVal()))
+				return
 			}
 		}
 	}
-	return "", kube.PodIdentifier("")
+
+	return "", kube.PodIdentifier(""), errors.New("could not assign pod id basing on associations")
 }
 
 func getConnectionIP(ctx context.Context) kube.PodIdentifier {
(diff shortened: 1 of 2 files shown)


pkg/processor/k8sprocessor/pod_association.go
```diff
@@ -23,15 +23,21 @@ import (
...
 )
 
 // extractPodIds extracts IP and pod UID from attributes or request context.
 // It returns a value pair containing configured label and IP Address and/or Pod UID.
 // If empty value in return it means that attributes does not contains configured label to match resources for Pod.
-func extractPodID(ctx context.Context, attrs pcommon.Map, associations []kube.Association) (podIdentifierKey string, podIdentifierValue kube.PodIdentifier) {
-	connectionIP := getConnectionIP(ctx)
+func extractPodID(
+	logger *zap.Logger,
+	ctx context.Context,
```
pdelewski, line 36: nit: ctx as convention should be passes as first parameter

pkg/processor/k8sprocessor/pod_association.go
```diff
@@ -103,9 +111,10 @@ func extractPodID(ctx context.Context, attrs pcommon.Map, associations []kube.As
 	return "", kube.PodIdentifier("")
 }
 
-func getConnectionIP(ctx context.Context) kube.PodIdentifier {
+func getConnectionIP(logger *zap.Logger, ctx context.Context) kube.PodIdentifier {
```
pdelewski, line 114: nit: same as above

pkg/processor/k8sprocessor/pod_association.go
```diff
@@ -23,15 +23,21 @@ import (
...
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube"
 )
 
 // extractPodIds extracts IP and pod UID from attributes or request context.
 // It returns a value pair containing configured label and IP Address and/or Pod UID.
 // If empty value in return it means that attributes does not contains configured label to match resources for Pod.
-func extractPodID(ctx context.Context, attrs pcommon.Map, associations []kube.Association) (podIdentifierKey string, podIdentifierValue kube.PodIdentifier) {
-	connectionIP := getConnectionIP(ctx)
+func extractPodID(
```
pdelewski, line 34: We could also return error and log error outside of this function

pkg/processor/k8sprocessor/pod_association.go
swiatekm-sumo: Here specifically, I think it'd be better to avoid passing the logger into this function and instead emit the log in the caller if  it returns an empty string.

pkg/processor/k8sprocessor/pod_association.go
swiatekm-sumo: This would work for the sites where we return immediately, but not for `getConnectionIP`, where we continue even if we don't get anything. Maybe it's just not worth logging in that case?

pkg/processor/k8sprocessor/pod_association.go
sumo-drosiek: I believe there is only rule according to ctx :sweat_smile:, but I don't have any document behind that :upside_down_face:

pkg/processor/k8sprocessor/pod_association.go
aboguszewski-sumo: Well, the log in `getConnectionIp` was not mentioned in the issue (that function didn't even exist when the issue was created), so I'll just follow this suggestion and ignore that log.

pkg/processor/k8sprocessor/pod_association.go
```diff
@@ -53,8 +58,8 @@ func extractPodID(ctx context.Context, attrs pcommon.Map, associations []kube.As
 			podIdentifierValue = kube.PodIdentifier(hostname)
 			return
 		}
-		podIdentifierKey = ""
-		return
+
+		return "", "", errors.New("pod association not set, could not assign other pod id")
```
aboguszewski-sumo, line 62: I decided to clean some returns in this function, because they were inconsistent in which other. Basing on all return calls in this function:
 - if an error is returned, use `return "", "", error"
 - if everything is fine, assign a value to `podIdentifierValue` (and `podIdentifierKey` if applies) and use `return`

pkg/processor/k8sprocessor/pod_association.go
```diff
@@ -87,20 +92,22 @@ func extractPodID(ctx context.Context, attrs pcommon.Map, associations []kube.As
...
 			if pod.StringVal() == "" || namespace.StringVal() != "" {
-				return asso.Name, kube.PodIdentifier(fmt.Sprintf("%s.%s", pod.StringVal(), namespace.StringVal()))
+				podIdentifierKey = asso.Name
+				podIdentifierValue = kube.PodIdentifier(fmt.Sprintf("%s.%s", pod.StringVal(), namespace.StringVal()))
+				return
 			}
 		}
 	}
-	return "", kube.PodIdentifier("")
+	return "", "", errors.New("could not assign pod id basing on associations")
```
aboguszewski-sumo, line 110: Is there a difference between an empty string and `kube.PodIdentifier("")` since `kube.PodIdentifier` is defined as string?

swiatekm-sumo: There isn't in practice, but I'd personally prefer using the full type name `kube.PodIdentifier("")`.

This is a bit subtle, because you can do `return "", ""`, but if you did:
```go
var retval string = ""
return "", retval
```
you'd get an error due to type mismatch. Go does a bit of implicit type conversion in specific circumstances, and this is one of them.

swiatekm-sumo (approved):

PR #670, opened 2022-07-18
feat(sumologicexporter, sumologicschemaprocessor)!: move translating metadata from sumologic exporter to sumologic schema processor

### LINKED ISSUE #649: Move metadata attribute translation from sumologicexporter to sumologicschemaprocessor
Currently, our exporter takes care of translating attribute names from the OpenTelemetry schema to the Sumo schema used by apps. See the definitions [here](https://github.com/SumoLogic/sumologic-otel-collector/blob/main/pkg/exporter/sumologicexporter/translate_attributes.go).

However, exporters shouldn't modify data, as this causes race conditions and memory corruption when the same records are multiplexed into more than one exporter. Originally, the intent of this feature was to make it easier for users to send data with correct metadata to Sumo without having to configure multiple components, but in hindsight, this wasn't worth the problems it causes.

We've since introduced a [separate processor](https://github.com/SumoLogic/sumologic-otel-collector/tree/main/pkg/processor/sumologicschemaprocessor) for the purpose of translating metadata to what the Sumo backend and apps expect. The attribute translation should live in this processor instead.

This would be a breaking change and the fix should come with migration instructions.

aboguszewski-sumo:
Fixes #649 

The `sumologic` exporter modifies data it gets by translating attributes from OpenTelemetry convention to Sumo convention. This PR moves this logic to the `sumologic schema` processor.

Technically this is ready for review, but it might have not been tested properly yet, so I'm marking this as a draft.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+This release deprecates the following feature, which will be removed in `v0.59.0`:
+
+- 'sumologic' exporter: translate attributes ([upgrade guide][upgrade_guide_unreleased_moved_translation])
+
+### Added
+
+- feat(sumologicschemaprocessor): add translating attributes
+
+### Changed
+
+- feat(sumologicexporter): deprecate translating attributes ([upgrade guide][upgrade_guide_unreleased_moved_translation])
+
 ### Fixed
 
 - fix(k8sprocessor): only apply the node filter to Pods [#668]
 
 [Unreleased]: https://github.com/SumoLogic/sumologic-otel-collector/compare/v0.55.0-sumo-0...main
 [#668]: https://github.com/SumoLogic/sumologic-otel-collector/pull/668
+[upgrade_guide_unreleased_moved_translation]: ./docs/Upgrading.md#sumologic-exporter-drop-support-for-translating-attributes
 
 ## [v0.55.0-sumo-0]
 
--- a/docs/Upgrading.md
+++ b/docs/Upgrading.md
@@ -1,7 +1,9 @@
 # Upgrading
 
+- [Unreleased](#unreleased)
+  - [`sumologic` exporter: moved translating attributes to `sumologicschema` processor](#sumologic-exporter-drop-support-for-translating-attributes)
 - [Upgrading to v0.55.0-sumo-0](#upgrading-to-v0550-sumo-0)
-  - [`filter` processor: Drop support for `expr` language](#filter-processor-drop-support-for-expr-language)
+  - [`filter` processor: drop support for `expr` language](#filter-processor-drop-support-for-expr-language)
 - [Upgrading to v0.52.0-sumo-0](#upgrading-to-v0520-sumo-0)
   - [`sumologic` exporter: Removed `carbon2` and `graphite` metric formats](#sumologic-exporter-removed-carbon2-and-graphite-metric-formats)
 - [Upgrading to v0.51.0-sumo-0](#upgrading-to-v0510-sumo-0)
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -7,12 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Breaking changes
+
+- feat(sumologicexporter)!: remove translating attributes ([upgrade guide](upgrade_guide_unreleased_removed_translation))
```
sumo-drosiek, line 12: Not sure if we want to make it breaking change or deprecate it first

docs/Upgrading.md
```diff
@@ -1,19 +1,90 @@
...
-  - [`k8s_tagger` processor: removed `clusterName` metadata extraction option](#k8s_tagger-processor-removed-clustername-metadata-extraction-option)
-    - [How to upgrade](#how-to-upgrade)
-  - [`sumologic` exporter: metadata translation: changed the attribute that is translated to `_sourceName` from `file.path.resolved` to `log.file.path_resolved`](#sumologic-exporter-metadata-translation-changed-the-attribute-that-is-translated-to-_sourcename-from-filepathresolved-to-logfilepath_resolved)
-    - [How to upgrade](#how-to-upgrade-1)
-- [Upgrading to 0.49.0-sumo-0](#upgrading-to-0490-sumo-0)
-  - [Several changes to receivers using opentelemetry-log-collection](#several-changes-to-receivers-using-opentelemetry-log-collection)
-  - [Sumo Logic exporter metadata handling](#sumo-logic-exporter-metadata-handling)
-    - [Removing unnecessary metadata using the resourceprocessor](#removing-unnecessary-metadata-using-the-resourceprocessor)
-    - [Moving record-level attributes used for metadata to the resource level](#moving-record-level-attributes-used-for-metadata-to-the-resource-level)
+- [Upgrading](#upgrading)
```
sumo-drosiek, line 3: I would skip this header in ToC

aboguszewski-sumo: An oversight of mine, this is an evildoing of a vscode extension which I have missed, I'm sorry.

pkg/exporter/sumologicexporter/exporter_test.go
```diff
@@ -406,254 +405,6 @@ func TestPushOTLPLogsClearTimestamp(t *testing.T) {
 	}
 }
 
-func TestPushOTLPLogs_AttributeTranslation(t *testing.T) {
-	createLogs := func() plog.Logs {
```
aboguszewski-sumo, line 410: In my opinion, the tests removed in file `exporter_test.go` are not very important and can be removed safely (because similar tests have been added to the processor), but please check if I'm right. The reason they are so complex is the fact that this feature couldn't be tested in an easier way while it has been in the exporter.

docs/Upgrading.md
```diff
@@ -15,6 +17,76 @@
     - [Removing unnecessary metadata using the resourceprocessor](#removing-unnecessary-metadata-using-the-resourceprocessor)
     - [Moving record-level attributes used for metadata to the resource level](#moving-record-level-attributes-used-for-metadata-to-the-resource-level)
 
+## Unreleased
+
+### `sumologic` exporter: moved translating attributes to `sumologicschemaprocessor`
```
sumo-drosiek, line 22:
```suggestion
### `sumologic` exporter: moved translating attributes to `sumologicschema` processor
```

docs/Upgrading.md
```diff
@@ -15,6 +17,76 @@
...
+    endpoint: http://127.0.0.1:8080/
+    translate_attributes: true
+
+service:
+  pipelines:
+    logs:
+      receivers: [rec]
+      processors: [proc]
+      exporters: [sumologic]
+```
```
sumo-drosiek, line 50: I think we can skip non related part and change service to:

```
processors:
  # - ...
  - sumologicschema
```

pkg/processor/sumologicschemaprocessor/README.md
```diff
@@ -42,3 +48,44 @@ and if found, adds the corresponding `cloud.namespace` resource attribute.
 If the `cloud.platform` resource attribute is not found or has a value that is not in the table, nothing is added.
 
 [opentelemetry_cloud_provider_attribute]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.9.0/specification/resource/semantic_conventions/cloud.md
+
+### Attribute translation
+
+Attribute translation changes some of the attribute keys from OpenTelemetry convention to Sumo convention.
+For example, OpenTelemetry convention for the attribute containing Kubernetes pod name is `k8s.pod.name`,
+but Sumo expects it to be in attribute named `pod`.
```
sumo-drosiek, line 56:
```suggestion
but Sumo Logic expects it to be in attribute named `pod`.
```

and for other occurrences the same

sumo-drosiek (changes requested):
Overall LGTM, but I would like to discuss if we can go with this change as breaking change, or now as deprecation and breaking change in the future

sumo-drosiek:
@aboguszewski-sumo please move exporter changes (actual breaking change) to another PR and set appropriate milestone, should be `v0.59`. In addition please add deprecation log to the exporter, like in https://github.com/SumoLogic/opentelemetry-collector-contrib/commit/c9d2ba5d0f04cc3460495bb7c7018fa33b141b4f

aboguszewski-sumo:
Done. I'll create the other pull request once this gets merged, to avoid rebasing too many times. As for now, it is located here: https://github.com/aboguszewski-sumo/sumologic-otel-collector/tree/sumologic-exporter-remove-translation

CHANGELOG.md
```diff
@@ -7,12 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## [Unreleased]
 
+### Added
+
+- feat(sumologicschemaprocessor): add translating attributes
+
+### Changed
+
+- feat(sumologicexporter): deprecate translating attributes ([upgrade guide][upgrade_guide_unreleased_moved_translation])
+
```
sumo-drosiek, line 17: We can add it to description as well

sumo-drosiek: https://github.com/SumoLogic/sumologic-otel-collector/blob/main/CHANGELOG.md?plain=1#L21-L23

sumo-drosiek (approved):
Thanks :rocket:

Issue #650, opened 2022-07-04
Move Telegraf metric name translation from sumologicexporter to sumologicschemaprocessor


swiatekm-sumo:
See #649 for broader context. The same issue applies to Telegraf name translation the exporter does [here](https://github.com/SumoLogic/sumologic-otel-collector/blob/main/pkg/exporter/sumologicexporter/translate_metrics.go).

PR #672, opened 2022-07-20
feat(sumologicexporter): remove translating attributes

aboguszewski-sumo:
Ref: https://github.com/SumoLogic/sumologic-otel-collector/pull/670#issuecomment-1189598498

The feature of translating attributes in `sumologic` exporter has been deprecated and is going to be removed in v0.59.

Note: probably the same issue will occur in #650


aboguszewski-sumo:
Please note: this has been rebased onto #678. The merge order should be important.

PR #675, opened 2022-07-21
feat(sumologicexporter, sumologicschemaprocessor): move translating Telegraf metric names from sumologic exporter to sumologic schema processor

### LINKED ISSUE #650: Move Telegraf metric name translation from sumologicexporter to sumologicschemaprocessor

aboguszewski-sumo:
Fixes #650

The `sumologic` exporter translates Telegraf metric names to Sumo Logic convention. This feature has been moved to the `sumologicschema` processor and deprecated in the exporter.

Ref: #670


docs/Upgrading.md
```diff
@@ -49,16 +50,51 @@ service:
...
 ```
 
+Full list of configuration settings can be found in the [readme][sumologicschema_processor_readme] of the processor.
+
+### `sumologic` exporter: drop support for translating Telegraf metric names
+
+Similar as above, the translation should not happen in the exporter and has been moved to the [sumologicschema processor][sumologicschema_processor].
+The functionality is now deprecated and will be removed soon.
+
+However, if the attributes are not translated, some Sumo apps might not work correctly.
```
sumo-drosiek, line 70: Logic :grinning:

pkg/processor/sumologicschemaprocessor/processor.go
```diff
@@ -53,7 +69,9 @@ func newSumologicSchemaProcessor(set component.ProcessorCreateSettings, config *
 func (processor *sumologicSchemaProcessor) start(_ context.Context, host component.Host) error {
 	processor.logger.Info(
 		"Processor sumologic_schema has started.",
-		zap.Bool("add_cloud_namespace", processor.cloudNamespaceProcessor.addCloudNamespace),
+		zap.Bool("add_cloud_namespace", processor.subprocessors[0].isEnabled()),
```
sumo-drosiek, line 72: Please move `"add_cloud_namespace"` and so on into subprocessor, as now we have to remember processors order

sumo-drosiek: or maybe let's give them name?

sumo-drosiek (commented):
I like the refactor a lot :guitar:

pkg/processor/sumologicschemaprocessor/processor.go
aboguszewski-sumo: Adding these properties would be a pain, because we would have to edit every subprocessor while adding a new one with new configuration. Another idea is to make another method, for example `config_name()` which would return an adequate name (`add_cloud_namespace` for the cloudNamespaceProcessor and so on).

sumo-drosiek: > because we would have to edit every subprocessor while adding a new one with new configuration

why? we could use property the same way like mentioned `config_name()`

aboguszewski-sumo: Ok, maybe I didn't understand well. I thought you meant adding fields like `add_cloud_namespace` to every subprocessor. Could you explain your idea in more detail?

aboguszewski-sumo: Oh, I think I get it now. Sorry for misunderstanding.

sumo-drosiek: I mean more less:
```
type SubProcess struct {
  name string
}

type translateTelegrafMetricsProcessor struct {
        SubProcess
	shouldTranslate bool
}

```

pkg/processor/sumologicschemaprocessor/processor.go
```diff
@@ -64,42 +84,36 @@ func (processor *sumologicSchemaProcessor) shutdown(_ context.Context) error {
...
-	}
+	for i := 0; i < len(processor.subprocessors); i++ {
+		err := processor.subprocessors[i].processLogs(logs)
 
-	logs, err = processor.translateAttributesProcessor.processLogs(logs)
-	if err != nil {
-		return logs, err
+		if err != nil {
+			return logs, err
+		}
```
sumo-drosiek, line 92: It can be shortened:

```go
if err := processor.subprocessors[i].processLogs(logs); err != nil {
    return logs, err
}
```

aboguszewski-sumo: I guess the refactor now makes it more difficult to track which one subprocessor returned the error. I thought that the errors would contain that info (as for now, there's nothing that can go wrong inside them).

sumo-drosiek: > I guess the refactor now makes it more difficult to track which one subprocessor returned the error

Let's use the name we just added for error propagation

```
fmt.Errorf("failed to process by %s: %v", processor.ConfigPropertyName(), err)
```

sumo-drosiek (approved):
Thank you!

Issue #255, opened 2021-09-24
[sourceprocessor,sumologicexporter] Remove source templates


sumo-drosiek:
Source templates are implemented in two places:
 - sourceprocessor (with addidional support for pod_name and possibility to be replaced by pod annotations)
 - sumologicexporter
 
 This is against DRY and to resolve it I propose the following:
 
 - remove source templates from sourceprocessor
 - add ability to define source templates using attribute in sumologicexporter with fallback to current approach
 
 With such solution, sourceprocessor will be able to prepare template based on pod annotations and exporter will take care of rest.

aboguszewski-sumo:
Is the proposed solution still valid or has the idea changed?

sumo-drosiek:
This changed. We are aiming to get rid of source templates in exporter and `_source*` attributes have already special meaning in processor.

This should be removed and only sourceprocessor should modify the data:

```
    # For below described source related configuration,
    # please refer to "Source templates" documentation chapter from this document.

    # desired source category, useful if you want to override the source category
    # configured for the source.
    source_category: <source_category>
    # desired source name, useful if you want to override the source name
    # configured for the source.
    source_name: <source_name>
    # desired host name, useful if you want to override the source host
    # configured for the source.
    source_host: <source_host>
```

PR #685, opened 2022-08-01
feat(sumologicexporter): deprecate setting source headers

aboguszewski-sumo:
Ref: #255 (this PR does not resolve this yet, since it needs to be removed)

The feature of setting the source headers (and source templates) is now deprecated and is going to be removed in v0.60.0.

### DIFF
--- a/pkg/exporter/sumologicexporter/exporter.go
+++ b/pkg/exporter/sumologicexporter/exporter.go
@@ -43,7 +43,7 @@ const (
 
 const translationDeprecationBanner = `
 ***********************************************************************************************************************************************************
-***    Translating attributes is deprecated and is going to be dropped soon. Please see the migration document:                                  	    ***
+***    Translating attributes is deprecated and is going to be dropped soon. Please see the migration document:                                         ***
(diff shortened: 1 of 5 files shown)


pkg/exporter/sumologicexporter/exporter.go
```diff
@@ -55,6 +55,13 @@ const telegrafTranslationDeprecationBanner = `
 ***********************************************************************************************************************************************************
 `
 
+const sourceTemplatesDeprecationBanner = `
+***********************************************************************************************************************************************************
+***    Adding source headers is deprecated and is going to be dropped soon. Please see the migration document:                              			***
+***    https://github.com/SumoLogic/sumologic-otel-collector/blob/main/docs/Upgrading.md#sumologic-exporter-drop-support-for-source-headers.    		***
```
mat-rumian, line 61: Some spaces to fix :)

mat-rumian (approved):
Overall LGTM, just one lint comment 👍

pkg/exporter/sumologicexporter/exporter.go
aboguszewski-sumo: I guess that VS Code tricks me, it looks fine in the IDE 😦

PR #686, opened 2022-08-01
feat(sumologicexporter): remove setting source headers

### LINKED ISSUE #255: [sourceprocessor,sumologicexporter] Remove source templates

aboguszewski-sumo:
Resolves #255 
Ref: #685 

This PR removes the functionality of setting source headers in the `sumologic` exporter, which has been deprecated some time ago.


pkg/exporter/sumologicexporter/config.go
```diff
@@ -144,6 +132,12 @@ Please consult the changelog at https://github.com/SumoLogic/sumologic-otel-coll
 		)
 	}
 
+	if cfg.SourceCategory != "" || cfg.SourceHost != "" || cfg.SourceName != "" {
+		return fmt.Errorf(`*Deprecation warning*: setting source headers is not supported anymore.
+Please consult the changelog at https://github.com/SumoLogic/sumologic-otel-collector/releases/tag/v0.60.0-sumo-0`,
+		)
+	}
+
```
sumo-drosiek, line 140: Do we need it?

aboguszewski-sumo: I'm not sure if we really need it. I found some time ago that such warnings have been printed so I added it. 
It was also added for attribute translation and telegraf metrics translation.

sumo-drosiek, line 140: Do we need it?

sumo-drosiek: We added that because it was deprecated, now it's going to be removed :)

aboguszewski-sumo: The first such warning was this: https://github.com/SumoLogic/sumologic-otel-collector/blob/19c766e288aacae3e1a757a74a87ffc6a824af8b/pkg/exporter/sumologicexporter/config.go#L128-L134

swiatekm-sumo: I think this is ok, but it probably shouldn't say `deprecation warning`.

aboguszewski-sumo: @sumo-drosiek @swiatekm-sumo what's the conclusion then? I'm not against any of the options, but if we don't want it here, we should probably also remove these warnings for other deleted options.

sumo-drosiek: @swiatekm-sumo we can do that

sumo-drosiek (approved):

PR #693, opened 2022-08-03
feat(build): validate FIPS mode at build time and runtime

swiatekm-sumo:
This adds an additional file to the main package which:
1. Imports a module that is supposed to guarantee that we're always running in FIPS mode: https://go.googlesource.com/go/+/dev.boringcrypto/src/crypto/tls/fipsonly/fipsonly.go
2. Adds an init check for whether we're using BoringSSL and prints a log line if so

It's not clear whether this is a good enough solution to attest that we're running in FIPS mode. Maybe we should also add a test that tries to use a FIPS-incompatible cipher for TLS and fails?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - feat: add glob config provider [#713]
+- feat(build): validate FIPS mode at build time and runtime [#693]
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


kkujawa-sumo (approved):

sumo-drosiek (approved):
> Maybe we should also add a test that tries to use a FIPS-incompatible cipher for TLS and fails?

That would be nice, but I think it can be done as separate PR

swiatekm-sumo:
> > Maybe we should also add a test that tries to use a FIPS-incompatible cipher for TLS and fails?
> 
> That would be nice, but I think it can be done as separate PR

I actually checked this, and it requires doing a more involved integration test - we'd need to start the collector with a receiver/exporter and make assertions about what they send over TLS.

sumo-drosiek:
> > > Maybe we should also add a test that tries to use a FIPS-incompatible cipher for TLS and fails?
> > 
> > 
> > That would be nice, but I think it can be done as separate PR
> 
> I actually checked this, and it requires doing a more involved integration test - we'd need to start the collector with a receiver/exporter and make assertions about what they send over TLS.

So, let's create an issue :sweat_smile:

PR #694, opened 2022-08-04
docs(readme): add a table with supported OS and architectures

### LINKED ISSUE #682: Add a nice arch/OS support matrix
Right now we only have instructions for all the different OS/arch combinations in the README. We should have that information organized in a support matrix so it's easy for users to tell at first glance what exactly is supported.

aboguszewski-sumo:
Resolves #682


kkujawa-sumo (approved):

aboguszewski-sumo:
@swiatekm-sumo could you check if this resolves the issue?

swiatekm-sumo (approved):

PR #697, opened 2022-08-05
docs(sumologicextension): add information about running the collector as systemd service

### LINKED ISSUE #683: Document running as a systemd service
Notably, systemd services are often run as users without a home folder, which causes the extension to not save local credentials. At minimum we should make it clear in the documentation that the user needs a home folder, or the credential store location should be explicitly set to some other location.

aboguszewski-sumo:
Resolves #683

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -6,6 +6,7 @@
   - [Basic configuration for traces](#basic-configuration-for-traces)
   - [Putting it all together](#putting-it-all-together)
 - [Authentication](#authentication)
+  - [Running the collector as systemd service](#running-the-collector-as-systemd-service)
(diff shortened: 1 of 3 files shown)


swiatekm-sumo (commented):
I would make these separate sections, or at least make the note more visible. As is, it's very easy to miss.

jspaleta:
One more change to consider, update the systemd section in Installation.md and have the user confirm that the opentelemetry user has a home directory defined and if not explicitly set a directory for storing the creds in the otel config.

If for some reason the user's system already has an opentelemetry user defined without a home directory the user creation instructions won't generate a home directory.

swiatekm-sumo (approved):

PR #701, opened 2022-08-10
feat: define stability levels for components

aboguszewski-sumo:
As for upstream version v0.57.2, the components must have a stability level defined. This PR does this.

I couldn't recall what exactly have we agreed on, so I have set all the levels to **beta**. Feel free to disagree with this.


swiatekm-sumo (approved):
Beta is correct

PR #702, opened 2022-08-10
chore(sumologicexporter,sumologicextension): replace deprecated functions

aboguszewski-sumo:
Methods from pdata referring to mutable slices and HTTPClientSettings's ToClient methods have been deprecated.
They have been removed in v0.57.2, so in order to upgrade, they must be replaced.


swiatekm-sumo (approved):
I don't think this needs a changelog entry.

PR #710, opened 2022-08-22
fix(k8sprocessor): keep pod's services information up to date

aboguszewski-sumo:
Information about pod's services can arrive late, i.e. after the pod has been marked as ready. If that happens, the data won't be marked with this information until another update of that pod.
This PR changes this behavior - information about services is no longer cached and is recalculated every time when `GetPod` is called.
This is a less intrusive solution for this problem - an alternative is to remove the whole `map[string]*Pod` from `WatchClient` and recalculate all the attributes every time.

cc: @swiatekm-sumo

### DIFF
--- a/pkg/processor/k8sprocessor/kube/client.go
+++ b/pkg/processor/k8sprocessor/kube/client.go
@@ -247,12 +247,14 @@ func (c *WatchClient) deleteLoop(interval time.Duration, gracePeriod time.Durati
 // GetPod takes an IP address or Pod UID and returns the pod the identifier is associated with.
 func (c *WatchClient) GetPod(identifier PodIdentifier) (*Pod, bool) {
 	c.m.RLock()
+	defer c.m.RUnlock()
 	pod, ok := c.Pods[identifier]
-	c.m.RUnlock()
 	if ok {
 		if pod.Ignore {
 			return nil, false
 		}
+
+		c.updatePodOwnerMetadata(pod)
 		return pod, ok
 	}
 	observability.RecordIPLookupMiss()
@@ -296,52 +298,7 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
 		}
 	}
 
-	if c.Rules.OwnerLookupEnabled {
-		c.logger.Debug("pod owner lookup",
-			zap.String("pod.Name", pod.Name),
-			zap.Any("pod.OwnerReferences", pod.OwnerReferences),
-		)
-		owners := c.op.GetOwners(pod)
-
-		for _, owner := range owners {
-			switch owner.kind {
-			case "DaemonSet":
-				if c.Rules.DaemonSetName {
-					tags[c.Rules.Tags.DaemonSetName] = owner.name
-				}
-			case "Deployment":
-				if c.Rules.DeploymentName {
-					tags[c.Rules.Tags.DeploymentName] = owner.name
-				}
-			case "ReplicaSet":
-				if c.Rules.ReplicaSetName {
-					tags[c.Rules.Tags.ReplicaSetName] = owner.name
-				}
-			case "StatefulSet":
-				if c.Rules.StatefulSetName {
-					tags[c.Rules.Tags.StatefulSetName] = owner.name
-				}
-			case "Job":
-				if c.Rules.JobName {
-					tags[c.Rules.Tags.JobName] = owner.name
-				}
-			case "CronJob":
-				if c.Rules.CronJobName {
-					tags[c.Rules.Tags.CronJobName] = owner.name
-				}
-
-			default:
-				// Do nothing
-			}
-		}
-
-		if c.Rules.ServiceName {
-			if services := c.op.GetServices(pod); len(services) > 0 {
-				tags[c.Rules.Tags.ServiceName] = strings.Join(services, c.delimiter)
-			}
-		}
-
-	}
+	// Owner metadata is updated on every query.
 
 	if len(pod.Status.ContainerStatuses) > 0 {
 		cs := pod.Status.ContainerStatuses[0]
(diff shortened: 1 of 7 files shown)


pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -247,13 +247,15 @@ func (c *WatchClient) deleteLoop(interval time.Duration, gracePeriod time.Durati
...
 	c.m.RLock()
+	defer c.m.RUnlock()
 	pod, ok := c.Pods[identifier]
-	c.m.RUnlock()
 	if ok {
 		if pod.Ignore {
 			return nil, false
+		} else {
+			c.updatePodServices(pod)
+			return pod, ok
```
sumo-drosiek, line 257: i would put it outside of `if-else`

pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -380,6 +377,14 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
 	return tags
 }
 
+func (c *WatchClient) updatePodServices(pod *Pod) {
+	if c.Rules.ServiceName {
+		if services := c.op.GetServices(pod.Name); len(services) > 0 {
```
sumo-drosiek, line 382: if there is no services, we should set empty string or remove the attribute

sumo-drosiek: Out of curiosity. can we shorten it to just

```
pod.Attributes[c.Rules.Tags.ServiceName] = strings.Join(services, c.delimiter)
```

?

aboguszewski-sumo: According to the implementation, yes: https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/strings/strings.go;l=430
I did not change it, because I was unsure if such way of coding is idiomatic in Go. But I also like it better.

sumo-drosiek: I would go with it

swiatekm-sumo:
What do you think of doing this for all owner metadata?

sumo-drosiek (approved):

aboguszewski-sumo:
> What do you think of doing this for all owner metadata?
> 
> As in, we'd skip calculating owner metadata on Pod updates, and only calculate it in `GetPod` on demand. We'd need to check of how this affects performance, but it'd make the code simpler and use less memory.

Actually, this is a question of "where do we draw the line". We can do this only for service data, only for owner metadata (including services) or even for all data. However, removing this feature completely would require some more changes. 

However, are you sure you mean only owner metadata? There is not that much owner metadata, so I don't think this would affect performance and memory usage.

In terms of performance, are there any ready benchmarks or tools for making them?

swiatekm-sumo:
> > What do you think of doing this for all owner metadata?
> > As in, we'd skip calculating owner metadata on Pod updates, and only calculate it in `GetPod` on demand. We'd need to check of how this affects performance, but it'd make the code simpler and use less memory.
> 
> Actually, this is a question of "where do we draw the line". We can do this only for service data, only for owner metadata (including services) or even for all data. However, removing this feature completely would require some more changes.
> 
> However, are you sure you mean only owner metadata? There is not that much owner metadata, so I don't think this would affect performance and memory usage.

I'd like to do it for all owner metadata. Singling out Services is a bit arbitrary, and I think doing a write on every records as we are in this PR is going to be slower than calculating the tags on the fly.

I'm actually not that concerned about performance - I primarily want to do this to simplify the code. We should check what the peformance impact is, but I don't expect it to be much either way.

> In terms of performance, are there any ready benchmarks or tools for making them?

There's some commented out benchmarks here: https://github.com/SumoLogic/sumologic-otel-collector/blob/138cfa99f021554b3930f205e8df087e645f933c/pkg/processor/k8sprocessor/kube/client_test.go#L1175. 

Generally speaking, what you'd want is a test where we seed the client with, say, 1000 Pods and then benchmark `GetPod`.

aboguszewski-sumo:
I ran it against 1000 pods, like you suggested.

Five consecutive runs before changes (current main):
```
BenchmarkSomePodsPerOwner-16              167499              6629 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.485s

BenchmarkSomePodsPerOwner-16              163314              6636 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.446s

BenchmarkSomePodsPerOwner-16              175059              6681 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   3.460s

BenchmarkSomePodsPerOwner-16              181920              6979 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.648s

BenchmarkSomePodsPerOwner-16              180145              7417 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.702s
```

Five consecutive runs after changes:
```
BenchmarkSomePodsPerOwner-16              164652              7840 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.712s

BenchmarkSomePodsPerOwner-16              157642              8150 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   3.785s

BenchmarkSomePodsPerOwner-16              150303              8949 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.758s

BenchmarkSomePodsPerOwner-16              148719              7408 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.487s
```

aboguszewski-sumo:
The benchmarks have been run on the newest commit (with added adding all owner metadata to a pod)

swiatekm-sumo:
Hm, so around 20% more CPU time per operation on average, it looks like. @SumoLogic/open-source-collection-team WDYT, is this worth the added reliability? Or should we try to keep the more efficient approach, and make it more complex to account for the Owner updates?

sumo-drosiek:
20% is something. Will the more complex approach be much more effective?

aboguszewski-sumo:
I'm sorry, I have made a mistake while running the benchmarks (`OwnerLookupEnabled` was not set to `true`). New results:

Current main:
```
BenchmarkSomePodsPerOwner-16              146277              7815 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.594s

BenchmarkSomePodsPerOwner-16              160446              7641 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.674s

BenchmarkSomePodsPerOwner-16              145642              7674 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.563s

BenchmarkSomePodsPerOwner-16              169126              7625 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.719s
```

HEAD~1 of this branch (update only services in `GetPod`):
```
BenchmarkSomePodsPerOwner-16              155786              7725 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.646s

BenchmarkSomePodsPerOwner-16              158461              7674 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.668s

BenchmarkSomePodsPerOwner-16              147369              7875 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.579s

BenchmarkSomePodsPerOwner-16              154473              7779 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.625s
```


HEAD of this branch (update all owner metadata):
```
BenchmarkSomePodsPerOwner-16              136042              8100 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.546s

BenchmarkSomePodsPerOwner-16              142065              8193 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.556s

BenchmarkSomePodsPerOwner-16              149042              8599 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.709s

BenchmarkSomePodsPerOwner-16              146815              8300 ns/op
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/kube   2.637s
```

PR #724, opened 2022-09-14
 test(k8sprocessor): do not skip if other associations are already up to date

sumo-drosiek:
Backporting https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/14097

### DIFF
--- a/pkg/processor/k8sprocessor/kube/client.go
+++ b/pkg/processor/k8sprocessor/kube/client.go
@@ -501,10 +501,10 @@ func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
 
 	for _, identifier := range identifiers {
 		if identifier != "" {
-			// compare initial scheduled timestamp for existing pod and new pod with same IP
-			// and only replace old pod if scheduled time of new pod is newer? This should fix
-			// the case where scheduler has assigned the same IP to a new pod but update event for
-			// the old pod came in later.
+			// compare initial scheduled timestamp for existing pod and new pod with same identifier
+			// and only replace old pod if scheduled time of new pod is newer or equal.
+			// This should fix the case where scheduler has assigned the same attributes (like IP address)
+			// to a new pod but update event for the old pod came in later.
 			if p, ok := c.Pods[identifier]; ok {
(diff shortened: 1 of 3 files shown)


pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -490,24 +490,28 @@ func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
...
+
 	if newPod.Name != "" && newPod.Namespace != "" {
-		c.Pods[generatePodIDFromName(newPod)] = newPod
+		identifiers = append(identifiers, generatePodIDFromName(newPod))
+	}
+
+	for _, identifier := range identifiers {
+		if identifier != "" {
+			// compare initial scheduled timestamp for existing pod and new pod with same IP
+			// and only replace old pod if scheduled time of new pod is newer? This should fix
```
kkujawa-sumo, line 505: Why is there a question mark after `newer`?

kkujawa-sumo (approved):

pkg/processor/k8sprocessor/kube/client.go
sumo-drosiek: I think its typo

sumo-drosiek, line 505:
```suggestion
			// and only replace old pod if scheduled time of new pod is newer. This should fix
```

pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -490,24 +490,28 @@ func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
...
+			// the case where scheduler has assigned the same IP to a new pod but update event for
+			// the old pod came in later.
```
sumo-drosiek, line 507:
```suggestion
			// compare initial scheduled timestamp for existing pod and new pod with same identifier
			// and only replace old pod if scheduled time of new pod is newer or equal.
			// This should fix the case where scheduler has assigned the same attributes (like IP address)
			// to a new pod but update event for the old pod came in later.
```

PR #727, opened 2022-09-14
fix(k8sprocessor): fix metadata enrichment (#725)

sumo-drosiek:
Backport #725


astencel-sumo (approved):

swiatekm-sumo:
I may be out of the loop on this, but why do we need to backport this change?

PR #729, opened 2022-09-19
chore(examples): add default configuration

sumo-drosiek:
It adds default configuration to examples, which:

* scrapes host metrics
* scrapes logs from the following destinations:

  * /var/log/auth.log
  * /var/log/syslog
  * /var/log/daemon.log
  * /var/log/kern.log
  * /var/log/dpkg.log

* uses file storage to buffer data

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -90,15 +90,15 @@ processors:
   ## The memory_limiter processor is used to prevent out of memory situations on the collector.
   ## ref: https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/memorylimiterprocessor
   memory_limiter:
-  ## check_interval is the time between measurements of memory usage for the
-  ## purposes of avoiding going over the limits. Defaults to zero, so no
-  ## checks will be performed. Values below 1 second are not recommended since
-  ## it can result in unnecessary CPU consumption.
-  check_interval: 5s
-  ## Maximum amount of memory, in %, targeted to be allocated by the process heap.
-  limit_percentage: 75
-  ## Spike limit (calculated from available memory). Must be less than limit_percentage.
-  spike_limit_percentage: 20
+    ## check_interval is the time between measurements of memory usage for the
+    ## purposes of avoiding going over the limits. Defaults to zero, so no
+    ## checks will be performed. Values below 1 second are not recommended since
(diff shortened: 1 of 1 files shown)


examples/default.yaml
```diff
@@ -0,0 +1,64 @@
+extensions:
+  sumologic:
+    install_token: ${INSTALL_TOKEN}
+    collector_credentials_directory: ~/.sumologic-otel-collector
+  health_check:
+  file_storage:
+    directory: ${FILE_STORAGE}
```
sumo-drosiek, line 7: This is planned to be filled in by install script

examples/default.yaml
```diff
@@ -0,0 +1,64 @@
+extensions:
+  sumologic:
+    install_token: ${INSTALL_TOKEN}
```
sumo-drosiek, line 3: This is planned to be filled in by install script

astencel-sumo (approved):

sumo-drosiek:
@astencel-sumo I added a lot of comments. Please check

astencel-sumo (approved):

examples/default.yaml
```diff
@@ -0,0 +1,147 @@
...
+  memory_limiter:
+  ## check_interval is the time between measurements of memory usage for the
+  ## purposes of avoiding going over the limits. Defaults to zero, so no
+  ## checks will be performed. Values below 1 second are not recommended since
+  ## it can result in unnecessary CPU consumption.
+  check_interval: 5s
+  ## Maximum amount of memory, in %, targeted to be allocated by the process heap.
+  limit_percentage: 75
+  ## Spike limit (calculated from available memory). Must be less than limit_percentage.
+  spike_limit_percentage: 20
```
sumo-drosiek, line 101:
```suggestion
    ## check_interval is the time between measurements of memory usage for the
    ## purposes of avoiding going over the limits. Defaults to zero, so no
    ## checks will be performed. Values below 1 second are not recommended since
    ## it can result in unnecessary CPU consumption.
    check_interval: 5s
    ## Maximum amount of memory, in %, targeted to be allocated by the process heap.
    limit_percentage: 75
    ## Spike limit (calculated from available memory). Must be less than limit_percentage.
    spike_limit_percentage: 20
```

PR #735, opened 2022-09-26
feat: use /etc/otelcol-sumo/conf.d as default systemd config directory

sumo-drosiek:


portertech:
@sumo-drosiek does seem like something that needs a changelog entry, no? What does testing look like for this kind of change? Do we have any particular process/rigor?

sumo-drosiek:
@portertech I keep the install script out of the changelog entries as this is more like infra, but yes, I think there should be some mention about it during release

> What does testing look like for this kind of change? 
For now I test in manually. Not ideal but doesn't have idea for handy functional tests (for both mac and linux). We are going to use this script in our e2e, so it should cover some of the cases

sumo-drosiek:
@portertech
Started working on tests in https://github.com/SumoLogic/sumologic-otel-collector/pull/739

sumo-drosiek:
superseded by https://github.com/SumoLogic/sumologic-otel-collector/pull/742

PR #739, opened 2022-09-28
test: init install script tests

sumo-drosiek:

### DIFF
--- a/pkg/scripts_test/go.mod
+++ b/pkg/scripts_test/go.mod
@@ -2,10 +2,7 @@ module sumologic_scripts_tests
 
 go 1.18
 
-require (
-	github.com/influxdata/influxdb v1.10.0
-	github.com/stretchr/testify v1.7.0
-)
+require github.com/stretchr/testify v1.7.0
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
--- a/pkg/scripts_test/go.mod
+++ b/pkg/scripts_test/go.mod
@@ -1,0 +1,14 @@
+module sumologic_scripts_tests
+
+go 1.18
+
+require (
+	github.com/influxdata/influxdb v1.10.0
+	github.com/stretchr/testify v1.7.0
+)
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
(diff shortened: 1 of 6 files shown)


pkg/scripts_test/go.mod
```diff
@@ -0,0 +1,14 @@
+module sumologic_scripts_tests
+
+go 1.18
+
+require (
+	github.com/influxdata/influxdb v1.10.0
```
swiatekm-sumo, line 6: Why is this in the direct dependencies?

pkg/scripts_test/install_test.go
```diff
@@ -0,0 +1,119 @@
...
+		return exiterr.ExitCode(), nil
+	}
+
+	return 0, fmt.Errorf("cannot obtain exit code: %v", err)
+}
+
+func TestInstallation(t *testing.T) {
+	defer tearDown(t)
+
+	_, err := os.Stat("/usr/local/bin/otelcol-sumo")
```
swiatekm-sumo, line 105: Shouldn't we be running the whole thing in an chroot? As is, it requires root and modifies system directories, which seems dangerous.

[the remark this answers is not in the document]
sumo-drosiek: to ensure that it is not created at beginning of test

pkg/scripts_test/go.mod
sumo-drosiek: good question

pkg/scripts_test/install_test.go
sumo-drosiek: How do you think that should work. Should we prepare tree with all binary dependencies and then run test in chrooted environment?

Should we do that inside or outside of the go binary?

sumo-drosiek: ok, i did some research and making chroot environment manually seems to be counterproductive. I would leave it as it is, as test have to be run as root to make some impact on the system

sumo-drosiek: ref: https://github.com/SumoLogic/sumologic-otel-collector/blob/5e7d67f52075f688368bb3a83f6334fd00222edd/pkg/scripts_test/install_test.go#L129-L216

swiatekm-sumo (approved):

PR #749, opened 2022-10-07
feat: use default config if present

swiatekm-sumo:


swiatekm-sumo:
We decided not to do this in the end.

PR #752, opened 2022-10-11
feat(scripts/install.sh): require install token

sumo-drosiek:


sumo-drosiek:
require #751

PR #760, opened 2022-10-13
feat(scripts/install.sh): validate arguments against user configuration

sumo-drosiek:
- If user install without providing any information we do nothing
- If user provides token, tags or api url we are validating it agains common file configuration:

  - If value exists and is the same we continue
  - if value exists and is different we exits
  - if value in file is empty we write it


astencel-sumo (approved):
🙈

PR #762, opened 2022-10-16
feat(ci): add windows builds to dev & pr workflows

amdprophet:

Adds Windows builds to the `Dev builds` & `PRs checks` GitHub Actions CI workflows.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,6 +9,37 @@
 
 ## Setting up development environment
 
+### Preparing the development environment (Windows)
+
+Several packages & environment variables are necessary to build in a Windows
+environment:
+
+1. Download & install the latest [Go MSI][go-msi].
+
+1. Download & install [Make for Windows][make-for-windows].
+
+1. Open `cmd` and add make to the user PATH:
+
+  ```bat
+  setx path "%PATH%;C:\Program Files (x86)\GnuWin32\bin"
+  ```
+
+1. Download & install [64-bit Git for Windows Setup][git-for-windows].
+
+1. Open gitbash.
+
+1. Navigate to the otelcolbuilder directory:
+
+  ```bash
+  cd sumologic-otel-collector/otelcolbuilder
+  ```
+
+1. Install otelcolbuilder:
+
+  ```bash
+  make install-builder
+  ```
+
 ### How to build
 
 ```bash
@@ -93,3 +124,8 @@ To contribute you will need to ensure you have the following setup:
   make -C otelcolbuilder install-builder \
     BUILDER_BIN_PATH=/custom/dir/bin/opentelemetry-collector-builder
   ```
+
+[go-msi]: https://go.dev/dl/
+[make-for-windows]: https://gnuwin32.sourceforge.net/downlinks/make.php
+[windows-terminal]: https://apps.microsoft.com/store/detail/windows-terminal/9N0DX20HK701
+[git-for-windows]: https://git-scm.com/download/win
--- a/otelcolbuilder/Makefile
+++ b/otelcolbuilder/Makefile
@@ -182,4 +182,4 @@ otelcol-sumo-linux_arm:
 
 .PHONY: otelcol-sumo-windows_amd64
 otelcol-sumo-windows_amd64:
-	GOOS=windows GOARCH=amd64 $(MAKE) build BINARY_NAME=$(BINARY_NAME)$(FIPS_SUFFIX)-windows_amd64.exe
+	GOOS=windows GOARCH=amd64 $(MAKE) build BINARY_NAME=$(BINARY_NAME)$(FIPS_SUFFIX)-windows_amd64
--- a/otelcolbuilder/Makefile
+++ b/otelcolbuilder/Makefile
@@ -90,9 +90,9 @@ _gobuild:
 		-tags enable_unstable \
 		-ldflags=$(LDFLAGS) \
 		-trimpath \
-		-o ./$(BINARY_NAME) . \
+		-o ./$(BINARY_NAME)$(BUILDER_BIN_EXT) . \
 	)
-	chmod +x ./cmd/$(BINARY_NAME)
+	chmod +x ./cmd/$(BINARY_NAME)$(BUILDER_BIN_EXT)
 
 .PHONY: _gobuild_debug
 _gobuild_debug:
@@ -101,9 +101,9 @@ _gobuild_debug:
 		-tags enable_unstable \
 		-race \
 		-gcflags "all=-N -l" \
-		-o ./$(BINARY_NAME)-debug . \
+		-o ./$(BINARY_NAME)-debug$(BUILDER_BIN_EXT) . \
 	)
-	chmod +x ./cmd/$(BINARY_NAME)-debug
+	chmod +x ./cmd/$(BINARY_NAME)-debug$(BUILDER_BIN_EXT)
 
 .PHONY: build
 build: ensure-correct-builder-version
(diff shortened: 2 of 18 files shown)


astencel-sumo:
👏 Impressive to see the engineering that went into this, but also scary to see how much more complicated the build process is becoming, given it was already quite complex. 😞 

Probably the most important requirement is to be able to perform the build on a machine other than a GitHub Actions runner (for example a local laptop, or an EC2). Some documentation would probably be needed for this.

perk-sumo:
Agreed with @astencel-sumo - document describing how to build it locally would be a very nice addition here.

Ideally this would work outside of Github Actions, eg. with bash script or sth.

amdprophet:
@astencel-sumo @perk-sumo I've started throwing together some notes to document how to build locally on Windows. Should I put it in the otelcolbuilder README.md file or somewhere in the docs directory?

I'm working on getting `make gotest` to pass at the moment. I've come across a test that checks the permissions of a temporary directory created by the tests (https://github.com/SumoLogic/sumologic-otel-collector/blob/main/pkg/extension/sumologicextension/extension_test.go#L251). Temporary directory permissions differ between Linux & Windows. I'm wondering if we can remove the check and instead run `chmod 700` against the temp directory.

astencel-sumo:
> @astencel-sumo @perk-sumo I've started throwing together some notes to document how to build locally on Windows.

Can't thank you enough for your work @amdprophet 🙇 

> Should I put it in the otelcolbuilder README.md file or somewhere in the docs directory?

My first thought is CONTRIBUTING.md in the root directory, what do you think?

> I'm working on getting `make gotest` to pass at the moment. I've come across a test that checks the permissions of a temporary directory created by the tests (https://github.com/SumoLogic/sumologic-otel-collector/blob/main/pkg/extension/sumologicextension/extension_test.go#L251). Temporary directory permissions differ between Linux & Windows. I'm wondering if we can remove the check and instead run `chmod 700` against the temp directory.

Hm yeah it doesn't seem that this check makes any sense. It doesn't seem to be testing any functionality of the extension, it's reather an assertion on how the `os.MkdirTmp()` function works, which I don't think is relevant to anything. I support removing this check.

amdprophet:
@astencel-sumo CONTRIBUTING.md makes sense to me. I've added some documentation to it; please let me know if you think it needs any changes.

amdprophet:
I believe this is ready for review again. There are two tests that are failing intermittently with `rawk8seventsreceiver` but they don't seem to be Windows-specific as they're also occasionally failing in macOS.

otelcolbuilder/Makefile
```diff
@@ -2,13 +2,31 @@ BINARY_NAME ?= otelcol-sumo
...
+endif
+
 OS ?= $(shell uname -s | tr A-Z a-z)
 ARCH ?= $(shell uname -m | sed s/aarch64/arm64/ | sed s/x86_64/amd64/)
 
+BUILDER_BIN_NAME ?= opentelemetry-collector-builder$(BUILDER_BIN_EXT)
+BUILDER_BIN_PATH ?= $(HOME)/bin
+
+ifeq ($(OS),windows)
+	BUILDER_BIN_UNIX_PATH ?= $(shell cygpath -u $(BUILDER_BIN_PATH))
```
astencel-sumo, line 23: Is this env var `BUILDER_BIN_UNIX_PATH` used for anything? I cannot see any references to it.

amdprophet: It was being used by curl previously but it seems to be working ok without it. I'll remove it.

astencel-sumo:
Thanks a lot @amdprophet, this looks much much better. Following the instructions, I was able to build the binary on Windows and it seems to work perfectly fine. 👌 I might have just a few nits for the instructions.

Here's a silly question though: I just noticed there's an `otelcol-sumo-windows_amd64` target in the [otelcolbuilder/Makefile](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.57.2-sumo-1/otelcolbuilder/Makefile#L167) that seems to also build a Windows binary. What's the benefit of running the build on an actual Windows runner, if there is any?

CONTRIBUTING.md
```diff
@@ -9,6 +9,47 @@
...
+
+1. Download & install [Make for Windows][make-for-windows].
+
+1. Open `cmd` and add make to the user PATH:
+
+  ```bat
+  setx path "%PATH%;C:\Program Files (x86)\GnuWin32\bin"
+  ```
+
+1. (Optional) Install [Windows Terminal][windows-terminal].
```
astencel-sumo, line 27: I don't think I would bother the user with installing the Windows Terminal. I mean, it's a good app that I use and would recommend to anyone, but it's orthogonal to the instructions on building the binary and it's another step, even if optional, which you could argue is even worse, because the user now needs to make a decision, which is the harder to make the more uninformed the user is. 🙂 

I'd rather keep the steps as minimal as possible.

CONTRIBUTING.md
```diff
@@ -9,6 +9,47 @@
...
+
+1. Download & install [64-bit Git for Windows Setup][git-for-windows]. If Windows
+  Terminal was installed in the previous step, enable the
+  `Add a Git Bash Profile` component in the installer.
+
+1. Open gitbash (or Windows Terminal using gitbash shell).
+
+1. Clone this repository:
```
astencel-sumo, line 35: I think I would assume the user has the repository already cloned, I would skip this step and straight to `cd otelcolbuilder`, the assumption being the user is already in the root of the repo. This is what is being done in the `How to build` steps below and I think it's enough.

amdprophet:
> Thanks a lot @amdprophet, this looks much much better. Following the instructions, I was able to build the binary on Windows and it seems to work perfectly fine. 👌 I might have just a few nits for the instructions.
> 
> Here's a silly question though: I just noticed there's an `otelcol-sumo-windows_amd64` target in the [otelcolbuilder/Makefile](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.57.2-sumo-1/otelcolbuilder/Makefile#L167) that seems to also build a Windows binary. What's the benefit of running the build on an actual Windows runner, if there is any?

The binary doesn't have to be built on a Windows runner but the MSI packaging will. Rather than setting up a complex pipeline that can pass binaries from one runner to another, I thought it would be best to run everything on a single WIndows runner.

astencel-sumo:
> The binary doesn't have to be built on a Windows runner but the MSI packaging will.

Oh I see, so we want to expand this in the future to things that are easier/more feasible on Windows OS (like building the installer). Thanks for the explanation :+1:

astencel-sumo (approved):

swiatekm-sumo:
@amdprophet We need to prevent the install script tests from running on Windows as well.

swiatekm-sumo (changes requested):
@amdprophet We need to prevent the install script tests from running on Windows as well.

amdprophet:
@swiatekm-sumo I've already added build tags to prevent execution of those tests on Windows. It seems like this is due to the Makefile attempting to run the tests with sudo which doesn't exist on the Windows runner. I've used a variable called `SUDO` in the past which contains the name or path to sudo and then I overridde it to an empty string on platforms where there is no sudo. What do you think?

amdprophet:
It seems that the `env` command is problematic on Windows as well https://github.com/SumoLogic/sumologic-otel-collector/blob/main/pkg/scripts_test/Makefile#L5. Perhaps conditional behaviour for that make target is a better choice.

swiatekm-sumo:
```
chmod +x ./cmd/otelcol-sumo-windows_amd64.exe.exe
```
Looks like you've got an extraneous .exe somewhere in there.

swiatekm-sumo (approved):
LGTM

amdprophet:
I believe the failing test just needs to be rerun. I saw intermittent failures of that test on both Windows & macOS.

PR #767, opened 2022-10-17
feat(build): use patched versions of four upstream receivers

swiatekm-sumo:
Use patched versions of:

* hostmetricsreceiver
* apachereceiver
* mysqlreceiver
* elasticsearchreceiver

The intent is to move back to using upstream as the necessary changes are merged.


swiatekm-sumo:
This should wait for #769

sumo-drosiek (approved):

PR #774, opened 2022-10-19
fix(scripts/install.sh): fix default configuration

sumo-drosiek:


examples/sumologic.yaml
```diff
@@ -37,7 +37,7 @@ exporters:
   sumologic:
     sending_queue:
       enabled: true
-      storage: file_storage
+      persistent_storage_enabled: true
```
sumo-drosiek, line 40: this is because actually installed binary do not support `storage` property

swiatekm-sumo (approved):

PR #778, opened 2022-10-20
feat(scripts/install.sh): remove user and group during purging

sumo-drosiek:


pkg/scripts_test/check.go
```diff
@@ -250,3 +251,13 @@ func checkAbortedDueToDifferentTags(c check) {
 	require.Greater(c.test, len(c.output), 0)
 	require.Contains(c.test, c.output[len(c.output)-1], "You are trying to install with different tags than in your configuration file!")
 }
+
+func checkUserExists(c check) {
+	_, err := user.Lookup("otelcol-sumo")
```
pdelewski, line 256: nit: Would not make sense extract this string as const?

pdelewski (approved):

pkg/scripts_test/check.go
sumo-drosiek: I'll change it

PR #785, opened 2022-10-21
fix(scripts/install.sh): follow systemd flow if user token specified in file only

sumo-drosiek:


swiatekm-sumo (approved):
Are we going to exhaustively document the behaviour and options of the script? This sounds like it could be confusing for users.

PR #788, opened 2022-10-21
fix(scripts_test): use /var/lib/otelcol-sumo as home directory

sumo-drosiek:


swiatekm-sumo (approved):
Should we explicitly create this directory as well? I know we create the FILE_STORAGE dir which includes the HOME_DIRECTORY, but that's more implicit than I like.

sumo-drosiek:
`-m` is for creating directory

PR #792, opened 2022-10-25
feat(examples): use the install script in Ansible playbook

swiatekm-sumo:


perk-sumo (approved):
LGTM

examples/ansible/vars/default.yaml
```diff
@@ -1,6 +1,2 @@
-sumologic_otel_collector:
-  version: 0.50.0-sumo-0
-  src_config_path: config.yaml
-  memory_high: 200M
-  memory_max: 300M
-  systemd_service: false
+collector_tags: {}
```
kkujawa-sumo, line 1: Content of this file is different than mentioned in https://github.com/SumoLogic/sumologic-otel-collector/pull/792/files#diff-86b02de024e35f262267d5401c2bdc11bb36047fff823fa1ae05516813fd8ebbR20-R27. 
Is it desired?

examples/ansible/README.md
```diff
@@ -4,17 +4,27 @@ This playbook will install Sumo Logic Distro of [OpenTelemetry Collector][otc_li
...
+- `api_url`: Sumo Logic API url. You shouldn't need to set this in most normal circumstances.
+- `version`: version of Sumo Logic Distribution for OpenTelemetry Collector
+- `systemd_service`: enables creation of Systemd Service for Sumo Logic Distribution for OpenTelemetry Collector. Enabled by default. Note that this playbook will not start the collector if you disable this.
+- `src_config_path`: path to configuration directory for Sumo Logic Distribution for OpenTelemetry Collector
 
 ## Running playbook
 
-- Prepare [configuration](../../docs/Configuration.md) for Sumo Logic Distribution for OpenTelemetry Collector
+- Get an install token from Sumo Logic, see [install_token]
+- Prepare [configuration](../../docs/Configuration.md) for Sumo Logic Distribution for OpenTelemetry Collector and put the file in a folder of your choice
```
kkujawa-sumo, line 17: Maybe it is worthy to mention that all files from this directory will be copied to machines where otelcol will be installed.

examples/ansible/vars/default.yaml
swiatekm-sumo: Yes. Technically this is the set of default arguments. Maybe we shouldn't be telling users to modify it at all?

swiatekm-sumo: I've changed the instructions to pass these arguments on the command line.

PR #795, opened 2022-10-26
fix(scripts/install.sh): don't use sudo

swiatekm-sumo:
Don't use sudo in the script and have it exit if it's not run as root. This way the user has much better control over what it does, and won't be surprised by the script trying to run commands as root quietly.

I haven't added a test for this because it's a major pain to manage Linux users in Go, and the change is simple enough. Existing tests pass because they run as root to begin with.


perk-sumo (approved):
LGTM

PR #796, opened 2022-10-26
feat(scripts/install.sh): do not prompt user for confirmation

swiatekm-sumo:
Instead of prompting for confirmation, we should install everything by default and provide flags to disable specific features.

I'm planning to add flags to disable default config creation and systemd configuration in separate PRs.


perk-sumo (approved):
👍

PR #800, opened 2022-10-26
Add EnvironmentFile support into systemd service File

### LINKED ISSUE #799: Update systemd service to make use of optional EvironmentFile for holding environment variables references in otelcol config
otelcol configs can and will reference environment variables for sensitive information.

systemd service unit needs to use an EnvironmentFile directive to optionally read a set of files containing environment variable values as part of service start up that define all needed environment variables referenced in configs.

Install script needs to make sure the directory structure for these environement files is correctly installed with treated as sensitive by restrictive permissions as narrowly as is reasonable to give systemd and otelcol service user access.

jspaleta:
Closes #799


examples/systemd/otelcol-sumo.service
```diff
@@ -8,6 +8,7 @@ Group=otelcol-sumo
 MemoryHigh=2000M
 MemoryMax=3000M
 TimeoutStopSec=20
+EnvironmentFile=-/etc/otelcol-sumo/env/*.env
```
portertech, line 11: How does systemd behave if this directory does not exist or is empty?

scripts/install.sh
```diff
@@ -931,6 +935,13 @@ echo 'Changing permissions for config file and storage'
 chmod 440 "${CONFIG_PATH}"
 chmod -R 750 "${FILE_STORAGE}"
 
+echo 'Changing permissions for user env directory'
+chmod 440 "${USER_ENV_DIRECTORY}"
+chmod g+s "${USER_ENV_DIRECTORY}"
+
+echo 'Changing ownership for user env directory'
+chown -R "${SYSTEM_USER}":"${SYSTEM_USER}" "${USER_ENV_DIRECTORY}"
```
portertech, line 943: I like this separation (from conf.d), OT configuration can remain visible to most system users while sensitive values are obfuscated/protected.

scripts/install.sh
```diff
@@ -989,6 +1000,7 @@ fi
 echo 'Changing ownership for config and storage'
 chown -R "${SYSTEM_USER}":"${SYSTEM_USER}" "${CONFIG_PATH}" "${FILE_STORAGE}"
 
+
```
portertech, line 1003: Just like newlines?

examples/systemd/otelcol-sumo.service
jspaleta: =-  makes the directive optional, so if there is no file to read it... systemd just keeps going.

jspaleta: Just confirmed I rename the directory restarted my local otelcol-sumo service..
I'm getting the heartbeat http package in my mock sumologic endpoint but no mysql metrics.
I interpret this to mean.. mysql receiver isn't authing..no password defined... but the other otelcol functionality is working as expected.  systemd isn't trying to do its restaart loop

scripts/install.sh
jspaleta: left over from when i moved the env directory chown further up

scripts/install.sh
portertech: @jspaleta the system user doesn't yet exist.

portertech (changes requested):

scripts/install.sh
```diff
@@ -989,13 +997,18 @@ fi
...
 SYSTEMD_CONFIG_URL="https://raw.githubusercontent.com/SumoLogic/sumologic-otel-collector/${CONFIG_BRANCH}/examples/systemd/otelcol-sumo.service"
 
 TMP_SYSTEMD_CONFIG="otelcol-sumo.service"
 TMP_SYSTEMD_CONFIG_BAK="${TMP_SYSTEMD_CONFIG}.bak"
 echo 'Getting service configuration'
 curl -fL "${SYSTEMD_CONFIG_URL}" --output "${TMP_SYSTEMD_CONFIG}" --progress-bar
 sed -i.bak -e "s%/etc/otelcol-sumo%'${CONFIG_DIRECTORY}'%" "${TMP_SYSTEMD_CONFIG}"
+sed -i.bak -e "s%/etc/otelcol-sumo/env%'${USER_ENV_DIRECTORY}'%" "${TMP_SYSTEMD_CONFIG}"
+
```
sumo-drosiek, line 1011:
```suggestion
sed -i.bak -e "s%/etc/otelcol-sumo/env%'${USER_ENV_DIRECTORY}'%" "${TMP_SYSTEMD_CONFIG}"
```

scripts/install.sh
```diff
@@ -989,13 +997,18 @@ fi
...
```
sumo-drosiek, line 1012:
```suggestion
sed -i.bak -e "s%/etc/otelcol-sumo/env%'${USER_ENV_DIRECTORY}'%" "${TMP_SYSTEMD_CONFIG}"

```

scripts/install.sh
```diff
@@ -989,13 +997,18 @@ fi
...
 
 # Remove glob for versions up to 0.57
```
sumo-drosiek, line 1013:
```suggestion
sed -i.bak -e "s%/etc/otelcol-sumo/env%'${USER_ENV_DIRECTORY}'%" "${TMP_SYSTEMD_CONFIG}"

# Remove glob for versions up to 0.57
```

scripts/install.sh
```diff
@@ -1073,6 +1074,65 @@ if [[ "${DOWNLOAD_ONLY}" == "true" ]]; then
     exit 0
 fi
 
+<<<<<<< HEAD
```
portertech, line 1077: @jspaleta needs some cleanup.

portertech (approved):

sumo-drosiek:
Please add integration tests for it: https://github.com/SumoLogic/sumologic-otel-collector/tree/main/pkg/scripts_test

portertech:
@sumo-drosiek now checking env dir existence and permissions.

swiatekm-sumo (approved):

sumo-drosiek (approved):

PR #804, opened 2022-10-28
Fix dev binary downloads and add retries for curl calls in install script

perk-sumo:


scripts/install.sh
```diff
@@ -922,7 +947,7 @@ mkdir -p "${USER_CONFIG_DIRECTORY}"
 echo "Generating configuration and saving as ${CONFIG_PATH}"
 
 CONFIG_URL="https://raw.githubusercontent.com/SumoLogic/sumologic-otel-collector/${CONFIG_BRANCH}/examples/sumologic.yaml"
-if ! curl -f -s "${CONFIG_URL}" -o "${CONFIG_PATH}"; then
+if ! curl --retry 5 --connect-timeout 5 --max-time 30 --retry-delay 0 --retry-max-time 150 -f -s "${CONFIG_URL}" -o "${CONFIG_PATH}"; then
```
rnishtala-sumo, line 950: should there be a retry-delay > 0?

rnishtala-sumo (approved):

scripts/install.sh
perk-sumo: Hm I'm not sure we need to wait with another retry here. We might if we see problems though. I've tested this one and it seems to work OK with Github.

PR #810, opened 2022-11-02
fix(scripts/install.sh): use correct input for user prompts

swiatekm-sumo:
The change to the script is pretty straightforward. If we do `curl -s ... | sudo bash`, then `bash` reads commands from stdin and we can't prompt the user for input using that fd. Instead, we detect if we're running in a terminal, and if we aren't, we explicitly read from `/dev/tty`. This fixes the problem.

The change to tests is more complicated. Tests do not run in a terminal, and the script would now expect input from them via `/dev/tty`. In order to test both cases, I've done the following:
1. Existing tests run with `--yes`, which sidesteps the problem.
2. A new option was added to the test runner, which wraps the test in a fake terminal and uses that for communication.
3. Added new versions of existing tests which use the fake terminal and don't use `--yes`.

### DIFF
--- a/pkg/scripts_test/check.go
+++ b/pkg/scripts_test/check.go
@@ -19,6 +19,7 @@ import (
 type check struct {
 	test                *testing.T
 	installOptions      installOptions
+	fakeTTY             bool
 	code                int
 	err                 error
 	expectedInstallCode int
--- a/pkg/scripts_test/command.go
+++ b/pkg/scripts_test/command.go
@@ -6,10 +6,12 @@ import (
 	"bufio"
 	"fmt"
 	"io"
+
 	"os"
 	"os/exec"
 	"strings"
 
+	"github.com/creack/pty"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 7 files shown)


sumo-drosiek (approved):
Looks fine, but should be carefully tested :grimacing:

swiatekm-sumo:
I'm not sure this is a good idea. Might be better to just download the script separately and run things manually.

sumo-drosiek:
> I'm not sure this is a good idea. Might be better to just download the script separately and run things manually.

It shouldn't matter for customers if they do `curl | ...` or `curl && ...`

swiatekm-sumo:
> > I'm not sure this is a good idea. Might be better to just download the script separately and run things manually.
> 
> It shouldn't matter for customers if they do `curl | ...` or `curl && ...`

Maybe, but unfortunately it does matter for bash if we want to prompt the user for input. I think it's much easier to just run the script normally and use stdin for user input. It also has the benefit that we could pipe in the install token for improved security.

PR #811, opened 2022-11-02
feat(scripts/install.sh): add --skip-config option

swiatekm-sumo:


scripts/install.sh
```diff
@@ -112,6 +115,7 @@ Supported arguments:
 
   -${ARG_SHORT_API}, --${ARG_LONG_API} <url>                       Api URL
   -${ARG_SHORT_SYSTEMD}, --${ARG_LONG_SYSTEMD}    Preserves from Systemd service installation.
+  -${ARG_SHORT_SKIP_CONFIG}, --${ARG_LONG_SKIP_CONFIG}    Do not create default configuration.
```
sumo-drosiek, line 118: Just to ensure, did you run install script with `--help` to ensure alignment is correct?

sumo-drosiek (approved):
LGTM, one question

scripts/install.sh
swiatekm-sumo: Nope, fixed now.

PR #815, opened 2022-11-03
update install script to include setting ACL grants for typical linux log directories

jspaleta:
This is addresses the need to provide the sumo-otelcol permissions to read log files without having to manipulate multiple user or group definitions.

_most_ linux distributions _should_ have acl support enabled by default on their mounted filesystems...but its not universal.  This needs to be tested across officially supported linux distribution target configurations.

This PR assumes includes code to test for availability of the setfacl command, and if not found will skip the ACL manipulation.


scripts/install.sh
```diff
@@ -855,6 +856,14 @@ function get_binary_from_url() {
     fi
 }
 
+function set_acl_on_log_paths() {
+for path in ${LOG_FILE_PATHS}; do
+    echo -e "Running: setfacl -R -m d:u:${SYSTEM_USER}:r-x,u:${SYSTEM_USER}:r-x,g:${SYSTEM_USER}:r-x ${path}"
+    setfacl -R -m d:u:${SYSTEM_USER}:r-x,u:${SYSTEM_USER}:r-x,g:${SYSTEM_USER}:r-x ${path}
```
sumo-drosiek, line 862: is `setfacl` command always installed? If not please handle scenario when its missing

jspaleta: setfacl command conditionally tested for in latest commit

sumo-drosiek:
Sorry for additional asks. Could init variable on the [top of the script](https://github.com/SumoLogic/sumologic-otel-collector/blob/cde83fff401ae898c7528805fc69a2e160b8d5a3/scripts/install.sh#L58-L77) and make it readonly after [setting defaults](https://github.com/SumoLogic/sumologic-otel-collector/blob/cde83fff401ae898c7528805fc69a2e160b8d5a3/scripts/install.sh#L926-L927)?

sumo-drosiek (approved):

jspaleta:
@sumo-drosiek 
Do the install related tests need to be updated to make sure the setfacl command is installed so this new codepath gets tested?

sumo-drosiek:
I believe so. Do not require it in this PR, but if change is not urgent tests would be very helpful

swiatekm-sumo:
@jspaleta can you resolve conflicts on this? Looks like we're good to go otherwise.

sumo-drosiek (commented):
Could we add tests for it?

portertech:
We could add tests for it, but it would likely require `sudo` or `su` and a consistent existing log file to test the users access. Alternatively, we could run the ACL tool to inspect the ACL on `/var/log`.

portertech:
I don't believe ACL info will be represented in https://pkg.go.dev/io/fs#FileMode, I'll test and tinker.

PR #819, opened 2022-11-07
test(scripts/install.sh): check config path permissions and ownership

### LINKED ISSUE #817: Add file permission and ownership tests for install script
We should verify that all the files created by the install script have the right ownership and permissions. The lack of these has already caused bugs via unrelated changes. For example: #816 

- [x] System config file (#819)
- [ ] User config file
- [ ] File storage
- [ ] Extension credentials directory

swiatekm-sumo:
This partially resolves #817

### DIFF
--- a/pkg/scripts_test/check.go
+++ b/pkg/scripts_test/check.go
@@ -1,9 +1,12 @@
 package sumologic_scripts_tests
 
 import (
+	"io/fs"
 	"os"
 	"os/exec"
 	"os/user"
+	"strconv"
+	"syscall"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


sumo-drosiek (approved):
LGTM

btw, I was thinking about chaining checks, as some checks require the other one. e.g. checking configuration requires configuration file to exists. I believe it would make tests a bit more cleaner

swiatekm-sumo:
> LGTM
> 
> btw, I was thinking about chaining checks, as some checks require the other one. e.g. checking configuration requires configuration file to exists. I believe it would make tests a bit more cleaner

Chaining, as in, how exactly would that look? The problem with config files specifically is that the script doesn't touch existing ones, so if you set them up outside of the script, the permissions will stay as they are. That's why I didn't simply make the permissions check part of the existence check.

PR #823, opened 2022-11-09
docs: rename docs to kebab-case

### LINKED ISSUE #822: docs links should be in kebab-case
The files in the /docs directory should be kebab-cased, that is all lowercase with hyphens between words.

For example, `KnownIssues.md` should become `known-issues.md`.

astencel-sumo:
Fixes #822


swiatekm-sumo (approved):
Nice!

I looked at the Sumo docs, and it seems like there's only one place we need to update; https://github.com/SumoLogic/sumologic-documentation/search?q=sumologic-otel-collector

astencel-sumo:
> Nice!
> 
> I looked at the Sumo docs, and it seems like there's only one place we need to update; https://github.com/SumoLogic/sumologic-documentation/search?q=sumologic-otel-collector

Three https://github.com/SumoLogic/sumologic-documentation/pull/897/files

PR #826, opened 2022-11-10
chore: update OT core to v0.64.0

aboguszewski-sumo:
Important thing to note: Telegraf receiver has been updated, but disabled, because of dependency version conflict. It should be enabled again before the next release.


sumo-drosiek (changes requested):
I would wait for telegraf :grimacing:

docs/configuration.md
```diff
@@ -1,831 +1 @@
-# Configuration
```
swiatekm-sumo, line 1: Why are you deleting this file?

swiatekm-sumo (changes requested):

docs/configuration.md
aboguszewski-sumo: No remote idea. I guess this is the combination of macos filesystem issues and `git add *`.

swiatekm-sumo (approved):

sumo-drosiek (approved):

PR #836, opened 2022-11-17
fix(ci): remove --version flag in test for built image

aboguszewski-sumo:


aboguszewski-sumo:
This doesn't seem to work, closing it.

PR #838, opened 2022-11-17
docs: add installation troubleshooting

astencel-sumo:
Here's a minimal set of troubleshooting docs I came up with. We can always add more as we (or other users) hit issues installing the collector.


swiatekm-sumo (approved):
Nice!

PR #839, opened 2022-11-17
feat(receiver/apache)!: enable feature gates by default

aboguszewski-sumo:


docs/upgrading.md
```diff
@@ -51,8 +52,26 @@ extensions:
     install_token: ${SUMOLOGIC_INSTALL_TOKEN}
 ```
 
+### `apache` receiver: turn on feature gates for resource attributes
+
+The metrics in this receiver are now being sent with two resource attributes: `apache.server.name` and `apache.server.port`.
+Additionally, `apache.server.name` replaces `server_name` metric-level attribute.
+
+Both features are hidden behind feature gates, but because they are important for Sumo Logic apps,
+they have been enabled by default earlier than expected.
```
swiatekm-sumo, line 61:
```suggestion
they have been enabled by default ahead of the normal deprecation timeline.
```

docs/upgrading.md
```diff
@@ -51,8 +52,26 @@ extensions:
...
+
+To disable the new features, disable the feature gates in otelcol's arguments:
+
+```bash
+otelcol-sumo --config=file:config.yaml --feature-gates=-receiver.apache.emitServerNameAsResourceAttribute,-receiver.apache.emitPortAsResourceAttribute
+```
+
+More information about the feature gates can is [here][apache-feature-gates].
```
swiatekm-sumo, line 69:
```suggestion
More information about the feature gates can be found [here][apache-feature-gates].
```

swiatekm-sumo (commented):
Some small nitpicks about phrasing, looks good otherwise.

otelcolbuilder/cmd/configprovider.go
```diff
@@ -34,9 +36,18 @@ import (
...
+	// feature flags, which are enabled by default in our distro
+	err := featuregate.GetRegistry().Apply(map[string]bool{
+		"receiver.apache.emitServerNameAsResourceAttribute": true,
+		"receiver.apache.emitPortAsResourceAttribute":       true,
+	})
+
+	if err != nil {
+		return fmt.Errorf("setting feature gate flags failed: %s", err)
+	}
+
```
aboguszewski-sumo, line 48: Is it okay to return an error here instead of continuing? Afaik there's no logger in this function, so log and continue wouldn't work.

swiatekm-sumo: This will cause otel to terminate with this error, which I think is fine, as this can only error if we try to set flags which don't exist, afaik.

swiatekm-sumo (approved):

PR #841, opened 2022-11-17
docs(installation): add `--preserve-env` to installation command

astencel-sumo:


docs/installation.md
```diff
@@ -36,10 +36,19 @@ To run it as a standalone process you only need to run the binary file downloade
 
 1. Run installation script:
 
+    Either by piping `curl` straight into `bash`:
+
     ```bash
-    curl -s https://raw.githubusercontent.com/SumoLogic/sumologic-otel-collector/main/scripts/install.sh | sudo bash -s -- --installation-token "${SUMOLOGIC_INSTALL_TOKEN}"
+    curl -s https://raw.githubusercontent.com/SumoLogic/sumologic-otel-collector/main/scripts/install.sh | sudo --preserve-env=SUMOLOGIC_INSTALL_TOKEN bash
```
swiatekm-sumo, line 42: Should we just do `-E` instead to take the whole environment?

astencel-sumo: I wanted to explicitly only transfer this env var, as this is the only one used by the installation script.

swiatekm-sumo: I'm worried that we'll forget to update this command after adding more.

astencel-sumo: The form `--preserve-env=SUMOLOGIC_INSTALL_TOKEN` makes it explicit why we're using this argument. But I can change it to just `-E` and just explain in the document why it's needed.

astencel-sumo: Done

swiatekm-sumo (approved):

docs/installation.md
```diff
@@ -36,10 +36,21 @@ To run it as a standalone process you only need to run the binary file downloade
 
 1. Run installation script:
 
+    Either by piping `curl` straight into `bash`:
+
     ```bash
-    curl -s https://raw.githubusercontent.com/SumoLogic/sumologic-otel-collector/main/scripts/install.sh | sudo bash -s -- --installation-token "${SUMOLOGIC_INSTALL_TOKEN}"
+    curl -s https://raw.githubusercontent.com/SumoLogic/sumologic-otel-collector/main/scripts/install.sh | sudo -E bash
```
swiatekm-sumo, line 42: Actually, why'd you change the other arguments here?

astencel-sumo: I removed the `--installation-token "${SUMOLOGIC_INSTALL_TOKEN}"` because it's not needed and makes the command unnecessarily longer.

swiatekm-sumo: The way you've changed it makes it difficult to know how to actually pass arguments to the script when invoked this way. I'd rather it stay the way it is for this reason.

astencel-sumo: Oh right, this is a good argument.

astencel-sumo: Restored the `--installation-token "${SUMOLOGIC_INSTALL_TOKEN}"` part of the installation command.

swiatekm-sumo (approved):

PR #852, opened 2022-11-28
feat(packaging/msi): add msi packaging

amdprophet:

Adds code for building MSI packages with WiX and a C# DLL for manipulating YAML files when the MSI is executed. Only amd64 is tested and supported but it shouldn't be much work to add x86 and arm64 down the road.


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This release introduces the following breaking changes:
 - feat: add glob config provider [#713]
 - feat(build): validate FIPS mode at build time and runtime [#693]
 - feat(ci): add windows builds to dev & pr jobs [#762]
+- feat(packaging/msi): add msi packaging [#852]
```
astencel-sumo, line 22: You probably need to add the definition of the link [#852] down below, like this:

```
[#852]: https://github.com/SumoLogic/sumologic-otel-collector/pull/852
```

CONTRIBUTING.md
```diff
@@ -72,6 +72,28 @@ opentelemetry-collector-builder \
 2021-05-24T16:32:37.326+0200    INFO    builder/main.go:113     Compiled        {"binary": "./cmd/otelcol-sumo-linux_arm64"}
 ```
 
+### How to build packages (Windows MSI)
+
+1. Install Visual Studio 2019 or newer with the  `.NET desktop development`
+   workload selected in the installer.
+
+1. Open `cmd` and add msbuild to the user PATH. The path may differ depending on
```
astencel-sumo, line 80: Perhaps use the "Developer Command Prompt for VS 2019/2022" instead of using the regular "cmd" and adding MSBuild to path?

CONTRIBUTING.md
```diff
@@ -72,6 +72,28 @@ opentelemetry-collector-builder \
...
+   which version and edition of Visual Studio and which version and
+   platform of Windows are installed:
+
+  ```bat
+  setx path "%PATH%;%ProgramFiles(x86)%\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin"
+  ```
+
+1. Navigate to the `packaging/msi/wix` directory and run:
```
astencel-sumo, line 88: Perhaps worth mentioning that you need to `msbuild -Restore /p:Platform=x64` before that, and that you need to have the otelcol binary built first for that to work. At least this was my experience.

astencel-sumo (approved):
I tried it on a fresh Windows virtual machine downloaded from [here](https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/) and it worked :tada: :trophy: :clap:

CONTRIBUTING.md
amdprophet: Good catch, those steps are indeed required.

PR #858, opened 2022-12-03
Integrate collector with new metadata api

portertech:
Opening a draft PR for visibility and early feedback.

[Spec doc](https://docs.google.com/document/d/1wPtsmC3OU3W298oQanMeCB4vrgeFdJK6_-NpmLhf4-4/edit).

[Jira ticket](https://jira.kumoroku.com/jira/browse/SUMO-207538).

### DIFF
--- a/pkg/extension/sumologicextension/api/metadata.go
+++ b/pkg/extension/sumologicextension/api/metadata.go
@@ -1,0 +1,38 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+type OpenMetadataHostDetails struct {
+	Name      string `json:"Name"`
+	OsName    string `json:"OsName,omitempty"`
+	OsVersion string `json:"OsVersion,omitempty"`
+}
+
+type OpenMetadataAgentDetails struct {
+	RunningVersion string `json:"RunningVersion"`
+}
+
+type OpenMetadataNetworkDetails struct {
+	HostIpAddress string `json:"HostIpAddress"`
+	ProxyAddress  string `json:"ProxyAddress"`
+	ProxyPort     int    `json:"ProxyPort"`
+}
+
+type OpenMetadataRequestPayload struct {
+	HostDetails    OpenMetadataHostDetails    `json:"HostDetails"`
+	AgentDetails   OpenMetadataAgentDetails   `json:"AgentDetails"`
+	NetworkDetails OpenMetadataNetworkDetails `json:"NetworkDetails"`
+	TagDetails     map[string]string          `json:"TagDetails"`
+}
--- a/pkg/extension/sumologicextension/extension.go
+++ b/pkg/extension/sumologicextension/extension.go
@@ -65,6 +65,7 @@ type SumologicExtension struct {
 
 const (
 	heartbeatUrl = "/api/v1/collector/heartbeat"
+	metadataUrl  = "/api/v1/otCollector/metadata"
 	registerUrl  = "/api/v1/collector/register"
 
 	collectorIdField           = "collector_id"
@@ -537,6 +538,7 @@ func (se *SumologicExtension) heartbeatLoop() {
 }
 
 var errUnauthorizedHeartbeat = errors.New("heartbeat unauthorized")
+var errUnauthorizedMetadata = errors.New("metadata update unauthorized")
 
 type ErrorAPI struct {
 	status int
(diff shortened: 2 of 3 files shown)


pkg/extension/sumologicextension/extension.go
```diff
@@ -166,6 +170,11 @@ func (se *SumologicExtension) Start(ctx context.Context, host component.Host) er
 		zap.String(collectorIdField, colCreds.Credentials.CollectorId),
 	)
 
+	err = se.updateMetadataWithHTTPClient(ctx, se.httpClient)
+	if err != nil {
+		return err
```
portertech, line 175: Should failure to update metadata result in an error? Should we only log it?

portertech: Should we make further attempts (with backoff)? How important do we consider the metadata?

VibhavariMandal-sumo: metaData has many sub parts, like agent version, hostDetails and tags.
And it is used to populate the Agent List page in UI. Without these details, List page is not useful.
And updating tags has also other use cases , like tag will be used later for source, subscription and searching/sorting.
So, We should make attempts with backoff in case of failure.
On server side, we can also plan to generate a low priority alert if metadata update fails for more than x times for a customer.

VibhavariMandal-sumo: yes, failure to update metadata must be an error

perk-sumo: Hm, if this results in an error it might be a major pain for customers.

As long as we can send the data to Sumo I believe we should, even if we have degraded funcionality because collectors metadata are not there.

I believe that telemetry data are more important than collectors metadata.

portertech: I am going to update this PR to implement retry w/ backoff and log any/all related errors ("loudly" 😅).

pkg/extension/sumologicextension/extension.go
```diff
@@ -592,6 +602,157 @@ func (se *SumologicExtension) sendHeartbeatWithHTTPClient(ctx context.Context, h
...
+	var buff bytes.Buffer
+	if err = json.NewEncoder(&buff).Encode(api.OpenMetadataRequestPayload{
+		HostDetails: api.OpenMetadataHostDetails{
+			Name:        info.Hostname,
+			OsName:      info.OS,
+			OsVersion:   info.PlatformVersion,
+			Environment: se.conf.CollectorDescription, // TODO: Replace with new configuration attribute?
+		},
+		AgentDetails: api.OpenMetadataAgentDetails{
+			RunningVersion: "1.0.0",
```
portertech, line 674: @astencel-sumo How can I determine the agent running version?

portertech: @sumo-drosiek any ideas?

pkg/extension/sumologicextension/api/metadata.go
```diff
@@ -0,0 +1,37 @@
...
+type OpenMetadataNetworkDetails struct {
+	HostIpAddress string `json:"hostIpAddress"`
+}
+
+type OpenMetadataRequestPayload struct {
+	HostDetails      OpenMetadataHostDetails      `json:"hostDetails"`
+	CollectorDetails OpenMetadataCollectorDetails `json:"collectorDetails"`
+	NetworkDetails   OpenMetadataNetworkDetails   `json:"networkDetails"`
+	TagDetails       map[string]interface{}       `json:"tagDetails"`
+}
```
sumo-drosiek, line 37: Are we going to change api in the futher? Do we need any kind of versioning?

pkg/extension/sumologicextension/extension.go
```diff
@@ -592,6 +618,135 @@ func (se *SumologicExtension) sendHeartbeatWithHTTPClient(ctx context.Context, h
 	return nil
 }
 
+func getHostIpAddress() (string, error) {
+	c, err := net.Dial("udp", "8.8.8.8:53")
```
sumo-drosiek, line 622: What if this specific DNS server is filtered in customers environment. Do we care about such situations?
Maybe the address should be configurable at some point, and/or we should use some fallback if this method won't work

pkg/extension/sumologicextension/extension.go
sumo-drosiek: in `createExtension` you have access to `params.BuildInfo.Version`.

working example:
```
func createExtension(_ context.Context, params extension.CreateSettings, cfg component.Config) (extension.Extension, error) {
	fmt.Printf("current version: %s\n", params.BuildInfo.Version)
	config := cfg.(*Config)
	return newSumologicExtension(config, params.Logger, params.ID)
}
```

pkg/extension/sumologicextension/extension.go
```diff
@@ -74,9 +79,21 @@ const (
...
+	updateCollectorMetadataStage = featuregate.StageAlpha
+
 	DefaultHeartbeatInterval = 15 * time.Second
 )
 
+func init() {
+	featuregate.GetRegistry().MustRegisterID(
+		updateCollectorMetadataID,
+		updateCollectorMetadataStage,
+		featuregate.WithRegisterDescription("When enabled, the collector will updated its Sumo Logic metadata."),
```
swiatekm-sumo, line 92:
```suggestion
		featuregate.WithRegisterDescription("When enabled, the collector will update its Sumo Logic metadata on startup."),
```

pkg/extension/sumologicextension/extension.go
swiatekm-sumo: I don't think we actually connect to this address, we just need the connection object to check the local address. I'd change it to an obviously invalid address (`255.255.255.255` maybe?) and add a comment to make this clear.

pkg/extension/sumologicextension/extension.go
```diff
@@ -592,6 +618,135 @@ func (se *SumologicExtension) sendHeartbeatWithHTTPClient(ctx context.Context, h
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), &buff)
+	if err != nil {
+		return fmt.Errorf("unable to create HTTP request %w", err)
+	}
+
+	addJSONHeaders(req)
+
+	se.logger.Info("Updating collector metadata",
+		zap.String("URL", u.String()),
+		zap.String("body", buff.String()))
```
swiatekm-sumo, line 686: If we're going to log this, we should make sure that it's human-readable. Setting an indentation level on the encoder should suffice.

pkg/extension/sumologicextension/extension.go
```diff
@@ -592,6 +618,135 @@ func (se *SumologicExtension) sendHeartbeatWithHTTPClient(ctx context.Context, h
...
+			ErrorAPI{
+				status: res.StatusCode,
+				body:   buff.String(),
+			},
+		)
+
+	case http.StatusUnauthorized:
+		return errUnauthorizedMetadata
+
+	case http.StatusNoContent:
```
swiatekm-sumo, line 718: What does it mean when the API returns a 204? When no changes happened? It's a bit peculiar, because we're using a POST here, which in principle shouldn't ever result in a 204.

swiatekm-sumo (commented):
Looks good to me overall, left some minor comments.

pkg/extension/sumologicextension/extension.go
portertech: I was just sloppy, will remove 👍

pkg/extension/sumologicextension/extension.go
portertech: Correct, we don't connect. I liked the idea of using a known/typical DNS host/port, but this causes more confusion 😄 I'll use your suggestion.

pkg/extension/sumologicextension/api/metadata.go
portertech: I don't believe so, there may be additional fields. I used registration for reference.

pkg/extension/sumologicextension/extension.go
portertech: Zap String() doesn't pretty print it as expected, for example: `{\n\t\t\"runningVersion\":`

portertech: This logging is really for our own testing, it should move to debug or be removed when we remove the feature gate.

PR #871, opened 2022-12-15
[processor/sumologicschema]: allow aggregating attributes with given name patterns

### LINKED ISSUE #866: sumologicschemaprocessor: allow aggregating attributes with given name patterns
## Problem
The processor should have a possibility to aggregate attributes with given name pattern into a single key-value pair. 

This functionality can't be achieved now by the `transform` processor, because some attribute names are created dynamically.

## Example
Given the following list of attributes:
```
[
  "pod_labels_app": "demo",
  "pod_labels_pod_template_hash": "123456#&",
  "pod_labels_foo": "bar"
]
```
let's say that the user wants to map all pods with name `pod_labels_*` to a name `kubernetes.labels`. The result should be:
```
"kubernetes.labels": {
  "app": "demo",
  "pod_template_hash": "123456#&",
  "foo": "bar"
}
```

## Notes

It is assumed that only wildcards (`*`) are supported. If any other kind pattern matching should be supported (eg. regexes), it should be clear what the result names should be.

_Original upstream issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/14946_

aboguszewski-sumo:
Updates #866

Done. This is a basic version, which supports only prefixes.

### DIFF
--- a/pkg/processor/sumologicschemaprocessor/aggregate_attributes_processor.go
+++ b/pkg/processor/sumologicschemaprocessor/aggregate_attributes_processor.go
@@ -1,0 +1,179 @@
+// Copyright 2022 Sumo Logic, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sumologicschemaprocessor
+
+import (
+	"fmt"
+	"strings"
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+// aggregateAttributesProcessor
+type aggregateAttributesProcessor struct {
+	aggregations []*aggregation
+}
+
+type aggregation struct {
+	attribute string
+	prefixes  []string
+}
+
+func newAggregateAttributesProcessor(config []aggregationPair) (*aggregateAttributesProcessor, error) {
+	aggregations := []*aggregation{}
+
+	for i := 0; i < len(config); i++ {
+		pair := &aggregation{
+			attribute: config[i].Attribute,
+			prefixes:  config[i].Patterns,
+		}
+		aggregations = append(aggregations, pair)
+	}
+
+	return &aggregateAttributesProcessor{aggregations: aggregations}, nil
+}
+
+func (proc *aggregateAttributesProcessor) processLogs(logs plog.Logs) error {
+	for i := 0; i < logs.ResourceLogs().Len(); i++ {
+		resourceLogs := logs.ResourceLogs().At(i)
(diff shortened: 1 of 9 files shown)


pkg/processor/sumologicschemaprocessor/aggregate_attributes_processor.go
```diff
@@ -0,0 +1,255 @@
...
+		regex, err := regexp.Compile(regexStrWithWildcard)
+		if err != nil {
+			return nil, err
+		}
+
+		regexes = append(regexes, regex)
+	}
+
+	return &aggregation{attribute: pair.Attribute, patternRegexes: regexes}, nil
+}
```
aboguszewski-sumo, line 71: This does not check for exact matches, for example if user specifies pattern `pod_*`, we will match strings like `opod_123`. We should probably discard that, shouldn't we?

pkg/processor/sumologicschemaprocessor/aggregate_attributes_processor.go
```diff
@@ -0,0 +1,255 @@
...
+					// this string will be the name of this key in the new map.
+					// TODO: Potential name conflict to resolve, eg.:
+					// pod_*_bar_* matches pod_foo_bar_baz
+					// pod2_*_bar_* matches pod2_foo_bar_baz
+					// both will be renamed to foo_baz
+					name := strings.Join(match[1:], "_")
+					names = append(names, name)
+					val := pcommon.NewValueEmpty()
+					value.CopyTo(val)
+					attrs = append(attrs, val)
```
aboguszewski-sumo, line 227: As mentioned in the comment. This seems to be a general problem with naming, which I'm not sure how to solve.

pkg/processor/sumologicschemaprocessor/aggregate_attributes_processor.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	return &aggregateAttributesProcessor{aggregations: aggregations}, nil
+}
+
+func pairToAggregation(pair *aggregationPair) (*aggregation, error) {
+	regexes := []*regexp.Regexp{}
+
+	for i := 0; i < len(pair.Patterns); i++ {
+		// We do not support regexes - only wildcards (*). Escape all regex special characters.
```
sumo-drosiek, line 56: why not?

pkg/processor/sumologicschemaprocessor/aggregate_attributes_processor.go
swiatekm-sumo: Yes, we should check for exact matches. Also see my comment here: https://github.com/SumoLogic/sumologic-otel-collector/issues/866#issuecomment-1357672421.

pkg/processor/sumologicschemaprocessor/aggregate_attributes_processor.go
swiatekm-sumo: See my comment here: https://github.com/SumoLogic/sumologic-otel-collector/issues/866#issuecomment-1357672421

pkg/processor/sumologicschemaprocessor/aggregate_attributes_processor.go
swiatekm-sumo: We should either support regexes and require a named capture group, or only support prefixes. See: https://github.com/SumoLogic/sumologic-otel-collector/issues/866#issuecomment-1357672421

aboguszewski-sumo: There's a problem with naming in that case. But yes, now that I think about it, we can just force the user to specify at least one capture group

aboguszewski-sumo:
PR reworked.
Now only prefixes are supported.

sumo-drosiek (approved):

PR #874, opened 2022-12-19
feat: report attributes as fields

pdelewski:
There is opened bug https://jira.kumoroku.com/jira/browse/SUMO-207676, which is about reporting severity as a field. This PR does similar thing as LokiExporter, adding severity into set of attributes. It's preliminary solution, showing only possible way, not necessarily final solution and question is how do we want to tackle it.


swiatekm-sumo:
This shouldn't happen in the exporter, as exporters shouldn't modify data. I think this should be a feature in the schema processor, if anything, and should probably be configurable.

aboguszewski-sumo:
I agree that this should be in a processor. It also seems to be totally independent from what I'm doing right now in `sumologicschema` processor, so feel free to add it there as soon as you need it.

pdelewski:
> This shouldn't happen in the exporter, as exporters shouldn't modify data. I think this should be a feature in the schema processor, if anything, and should probably be configurable.

Thx. Agree, that it should be configurable

pkg/processor/sumologicschemaprocessor/log_fields_conversion_processor.go
```diff
@@ -0,0 +1,114 @@
...
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+func (proc *logFieldsConversionProcessor) processMetrics(metrics pmetric.Metrics) error {
+	// No-op. Traces should not be translated.
```
aboguszewski-sumo, line 99:
```suggestion
	// No-op. Metrics should not be translated.
```

pkg/processor/sumologicschemaprocessor/processor.go
```diff
@@ -75,6 +81,7 @@ func (processor *sumologicSchemaProcessor) start(_ context.Context, host compone
 		zap.Bool(procs[0].ConfigPropertyName(), procs[0].isEnabled()),
 		zap.Bool(procs[1].ConfigPropertyName(), procs[1].isEnabled()),
 		zap.Bool(procs[2].ConfigPropertyName(), procs[2].isEnabled()),
+		zap.Bool(procs[2].ConfigPropertyName(), procs[3].isEnabled()),
```
aboguszewski-sumo, line 84:
```suggestion
		zap.Bool(procs[3].ConfigPropertyName(), procs[3].isEnabled()),
```

pkg/processor/sumologicschemaprocessor/log_fields_conversion_processor.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func newLogFieldConversionProcessor(shouldConvert bool) (*logFieldsConversionProcessor, error) {
+	return &logFieldsConversionProcessor{
+		shouldConvert: shouldConvert,
+	}, nil
+}
+
+func addLogLevelAttribute(log plog.LogRecord) {
+
```
aboguszewski-sumo, line 68:
```suggestion
```

pkg/processor/sumologicschemaprocessor/testdata/config.yaml
```diff
@@ -9,7 +9,8 @@ processors:
     translate_attributes: false
   sumologic_schema/disabled-telegraf-attribute-translation:
     translate_telegraf_attributes: false
-
+  sumologic_schema/disabled-log-fields-convertion:
+    convert_attributes_to_fields: false
```
aboguszewski-sumo, line 13: I think that this name is too general - it suggests that this feature allows to converts all attributes to fields, but we only add severity number here. Also, we're converting a severity level into an attribute, so `attributes_to_fields` is misleading imo (if I'm not mistaken, fields in Sumo are the same thing as attributes in otlp, but correct me if I'm wrong)
I'd suggest something like `add_severity_level_attribute`

aboguszewski-sumo:
A few nits. Please modify `README.md` and add a sample test to `processor_test.go`

aboguszewski-sumo (changes requested):

pkg/processor/sumologicschemaprocessor/log_fields_conversion_processor.go
```diff
@@ -0,0 +1,114 @@
...
+	return nil
+}
+
+func (proc *logFieldsConversionProcessor) isEnabled() bool {
+	return proc.shouldConvert
+}
+
+func (*logFieldsConversionProcessor) ConfigPropertyName() string {
+	return "convert_to_fields"
+}
```
aboguszewski-sumo, line 114: IMO, it should be the same as the config property name (`convert_attributes_to_fields`), but check my other PR.

pkg/processor/sumologicschemaprocessor/testdata/config.yaml
pdelewski: true, that was my first idea, but then I decided to generalize that to maybe handle more attributes, however as it is now it make more sense to change as you suggest

pkg/processor/sumologicschemaprocessor/processor.go
pdelewski: Thx for spotting that

pkg/processor/sumologicschemaprocessor/README.md
```diff
@@ -67,6 +67,10 @@ processors:
       - attribute: <attribute>
         prefixes: [<prefix>]
       - ...
+
+    # Defines whether `severity` attribute should be added to record attributes.
+    # default = true
+    add_severity_level_attribute: {true, false}
```
astencel-sumo, line 73: I propose the following configuration, or something similar:

```yaml
add_severity_attribute:
  enabled: true
  attribute_name: loglevel
```

pdelewski: Changed configuration and extended to use other fields as well, eg. `severity text` `span id` and `trace id`. That was asked by our customer support, however I'm still waiting for confirmation if client needs that. Anyway, we can always leave it or remove.

sumo-drosiek: shouldn't this be updated?

pdelewski: @sumo-drosiek what do you mean? merge/rebase with main?

sumo-drosiek: I agree with @astencel-sumo that the configuration added by this PR should be grouped in one major key, instead of producing 4 new config options

pdelewski: I missed this one

pdelewski: @sumo-drosiek I don't think @astencel-sumo suggested grouping, but maybe I'm wrong?

pkg/processor/sumologicschemaprocessor/README.md
```diff
@@ -271,3 +278,21 @@ The result is:
...
+It's possible to convert specific attributes to fields. List of currently supported attributes is below:
+
+`add_severity_number_attribute`
+`add_severity_text_attribute`
+`add_space_id_attribute`
+`add_trace_id_attribute`
+
+In order to report one of them as field, following configuration is needed:
+
+```json
```
sumo-drosiek, line 294:
```suggestion
```yaml
```

pkg/processor/sumologicschemaprocessor/README.md
sumo-drosiek: @astencel-sumo? 🤔

bnartiff: Hi @pdelewski any update on this? The customer is inquiring for updates and I don't see any recent movement. Thank you.

pdelewski: Hi @bnartiff , I will update PR tomorrow. It still requires aprovals. I was also waiting for confirmation regarding fields that customer expects (as originally there were mentioned only severity, recently I updated that to `spanid` and `traceid`

bnartiff: Thanks for the update, I am going to check w/ the customer to confirm the desired fields. Thank you for adding spanid and traceid as well.

bnartiff: Per customer, "I think severity, spanID, and traceID are a good start.  I don’t have any other suggestions/needs at this point. " @pdelewski

pdelewski:
@astencel-sumo @sumo-drosiek @aboguszewski-sumo Could you look at this PR?

pkg/processor/sumologicschemaprocessor/README.md
```diff
@@ -271,3 +278,21 @@ The result is:
   }
 }
 ```
+
+### Severity Attribute
+
+Some fields that log entries consist are not displayed as fields in Sumo Logic out of the box.
```
aboguszewski-sumo, line 284:
```suggestion
Some fields that log entries consist of are not displayed as fields in Sumo Logic out of the box.
```

pkg/processor/sumologicschemaprocessor/README.md
```diff
@@ -67,6 +67,13 @@ processors:
       - attribute: <attribute>
         prefixes: [<prefix>]
       - ...
+
+    # Defines whether `severity` attribute should be added to record attributes.
+    add_severity_number_attribute:
+        enabled: true
+        attribute_name: "loglevel"
+
```
aboguszewski-sumo, line 75: Keep it consistent with other examples above in this file, eg.:
```
add_severity_number_attribute:
    # if enabled, then XYZ
    # default = true
     enabled: {true, false}
    # The name that is something
    # default = "loglevel"
     attribute_name: "loglevel"

File: pkg/processor/sumologicschemaprocessor/README.md [hunk 10]
```diff
@@ -271,3 +278,21 @@ The result is:
...
+`add_space_id_attribute`
+`add_trace_id_attribute`
+
+In order to report one of them as field, following configuration is needed:
+
+```yaml
+    add_severity_number_attribute:
+        enabled: true
+        attribute_name: "loglevel"
+```
```
aboguszewski-sumo, line 298: Can all of them be enabled or only one at one time? Maybe it's better to squash this into one attribute and disable it by selecting `none`, empty string or something like that?

pkg/processor/sumologicschemaprocessor/processor_test.go
```diff
@@ -1237,6 +1237,61 @@ func TestAggregateAttributesForTraces(t *testing.T) {
...
+				assert.Equal(t, "severity", attribute2.Str())
+				attribute3, found := outputLogs.ResourceLogs().At(0).ScopeLogs().At(0).LogRecords().At(0).Attributes().Get("spanid")
+				assert.True(t, found)
+				assert.Equal(t, "0101010101010101", attribute3.Str())
+				attribute4, found := outputLogs.ResourceLogs().At(0).ScopeLogs().At(0).LogRecords().At(0).Attributes().Get("traceid")
+				assert.True(t, found)
+				assert.Equal(t, "01010101010101010101010101010101", attribute4.Str())
+
+			},
+		},
```
aboguszewski-sumo, line 1276: I'm not sure how this works.
In the config (lines 1356+), you set only `SeverityNumberAttributeName` to be transformed, but here everything is transformed.

aboguszewski-sumo (commented):
I have some problems with understanding how should this work in general. As I understand - there are some OTLP fields in logs that are not shown in Sumo, and to fix that, we transform them into regular attributes. Should every of these fields be translated, or only one, selected in the config? The implementation suggests that all of them are translated, but the config suggests that only one of them.

pkg/processor/sumologicschemaprocessor/README.md
```diff
@@ -271,3 +278,21 @@ The result is:
...
 ```
+
+### Severity Attribute
+
+Some fields that log entries consist are not displayed as fields in Sumo Logic out of the box.
+It's possible to convert specific attributes to fields. List of currently supported attributes is below:
+
+`add_severity_number_attribute`
+`add_severity_text_attribute`
+`add_space_id_attribute`
```
aboguszewski-sumo, line 289:
```suggestion
`add_span_id_attribute`
```

pkg/processor/sumologicschemaprocessor/README.md
pdelewski: My idea was to be able to enable one by one, however I have not seen requirements regarding this from customer side.

pkg/processor/sumologicschemaprocessor/processor_test.go
pdelewski: Will check that

pkg/processor/sumologicschemaprocessor/README.md
aboguszewski-sumo: Ok, I've read the discussion above. In my opinion, the best thing would be to define something like this (the name should probably be different, I'm not sure how to call all these attributes):
```yaml
add_log_metadata_attributes:
  metadata: ["severity_no", "severity_text", "span_id", "trace_id"]
```

pdelewski: ok, I can update implementation according to your suggestion, @astencel-sumo @sumo-drosiek What's your opinion on that? I would like to avoid yet another round of changes regarding configuration.

pkg/processor/sumologicschemaprocessor/processor_test.go
```diff
@@ -1298,3 +1353,15 @@ func newAggregateAttributesConfig(aggregations []aggregationPair) *Config {
 	config.AggregateAttributes = aggregations
 	return config
 }
+
+func newLogFieldsConversionConfig(logFieldsConversion bool) *Config {
```
astencel-sumo, line 1357: The parameter `logFieldsConversion` does not seem to be used?

pkg/processor/sumologicschemaprocessor/README.md
astencel-sumo: I'm not convinced that squashing all configurations into a single property is any better than the current solution. We can discuss in a meeting.

sumo-drosiek: > I'm not convinced that squashing all configurations into a single property is any better than the current solution. We can discuss in a meeting.

same as we skip `attribute_name`

pkg/processor/sumologicschemaprocessor/processor_test.go
pdelewski: I updated PR according to our today discussion as well as fixed some bugs. Current configuration looks as suggested by @sumo-drosiek, however I have not change main key `field_attributes`
```
field_attributes:
  trace_id:
    enabled: true
    name: traceid
  span_id:
    enabled: true
    name: spanid
```

sumo-drosiek (approved):

pkg/processor/sumologicschemaprocessor/README.md
```diff
@@ -67,6 +67,16 @@ processors:
...
+
+    # Defines whether attribute should be added to record attributes.
+    # Currently available attributes are `severity_number`, `severity_text`,
+    # `trace_id`, `span_id`
+    field_attributes:
+      severity_number:
+        enabled: true
+        name: "loglevel"
+
+    - ...
```
aboguszewski-sumo, line 79: this line seems redundant

pkg/processor/sumologicschemaprocessor/README.md
```diff
@@ -67,6 +67,16 @@ processors:
...
         prefixes: [<prefix>]
       - ...
+
+    # Defines whether attribute should be added to record attributes.
+    # Currently available attributes are `severity_number`, `severity_text`,
+    # `trace_id`, `span_id`
+    field_attributes:
+      severity_number:
+        enabled: true
+        name: "loglevel"
```
aboguszewski-sumo, line 77: Info about defaults is missing

aboguszewski-sumo (changes requested):
Could you add a test that checks if this is correct?
```
    field_attributes:
      severity_number:
        enabled: true
        name: "definitely_not_default_name"
```

pkg/processor/sumologicschemaprocessor/config.go
```diff
@@ -66,6 +72,12 @@ func createDefaultConfig() component.Config {
 			SquashSingleValues: defaultNestingSquashSingleValues,
 		},
 		AggregateAttributes: defaultAggregateAttributes,
+		LogFieldsAttributes: &logFieldAttributesConfig{
+			SeverityNumberAttribute: &logFieldAttribute{defaultAddSeverityNumberAttribute, "loglevel"},
+			SeverityTextAttribute:   &logFieldAttribute{defaultAddSeverityTextAttribute, "severitytext"},
+			SpanIdAttribute:         &logFieldAttribute{defaultAddSpanIdAttribute, "spanid"},
+			TraceIdAttribute:        &logFieldAttribute{defaultAddTraceIdAttribute, "traceid"},
+		},
```
aboguszewski-sumo, line 80: This seems fishy (attribute names are hardcoded), this is what I want you to test

pkg/processor/sumologicschemaprocessor/README.md
pdelewski: update

pkg/processor/sumologicschemaprocessor/README.md
pdelewski: ok

pdelewski:
> Could you add a test that checks if this is correct?
> 
> ```
>     field_attributes:
>       severity_number:
>         enabled: true
>         name: "definitely_not_default_name"
> ```

I updated current test using suggested name `definitely_not_default_name`. Will this satisfy your requirements?

aboguszewski-sumo (approved):
I overlooked one thing when I asked for that, sorry for inconvienience.

PR #877, opened 2022-12-20
[processor/sumologicschema]: add nesting processor

### LINKED ISSUE #865: sumologicschemaprocessor: transform attributes into JSON-like structure
## Problem
The processor should gain possibility to transform set of attributes with common namespace prefix in the attribute name, to a nested key-value structure. 

## Example

The following list of attributes:
```
[
  "kubernetes.container_name": "xyz",
  "kubernetes.host.name": "the host",
  "kubernetes.host.address": "127.0.0.1",
  "kubernetes.namespace_name": "sumologic",
  "another_attr": "42" 
]
```
Should be translated into such list:
```
[
  "kubernetes": {
    "container_name": "xyz",
    "namespace_name": "sumologic",
    "host": {
      "name": "the host",
      "address": "127.0.0.1"
    }
  },
 "another_attr": "42"
]
```

## Notes

Is is assumed, that by "prefix", we mean the namespace path separated by `.`, for example `xyz` is a common prefix path for `xyz.abc.qwe` and `xyz.foo`, but `sumo` **is not** a prefix for `sumologic.foo.baz` and `sumologic.bar.baz`.

Optional config options that are not necessary, but might be useful:

- changing the separator from `.` to any string
- setting the translation on/off for both resource attributes and regular attributes
- denylist of prefixes that shouldn't be transformed

_Original upstream issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/14977_

aboguszewski-sumo:
Fixes #865 
Practically done, e2e tests are missing.


pkg/processor/sumologicschemaprocessor/README.md
```diff
@@ -98,3 +109,42 @@ Below is a list of all attribute keys that are being translated.
 | `k8s.statefulset.name`    | `statefulset`       |
 | `service.name`            | `service`           |
 | `log.file.path_resolved`  | `_sourceName`       |
+
+### Nesting attributes
+
+Nesting attributes allows to change the structure of attributes (both resource attributes and normal ones)
```
astencel-sumo, line 115: Aren't resource attributes normal? :smile:
```suggestion
Nesting attributes allows to change the structure of attributes (both resource level and record level attributes)
```

aboguszewski-sumo: They always felt special for me 🙂

pkg/processor/sumologicschemaprocessor/README.md
```diff
@@ -98,3 +109,72 @@ Below is a list of all attribute keys that are being translated.
...
+{
+  "a": {
+    "b": {
+      "c": ...
+    }
+  }
+}
+```
+
+However, it is not possible to know a priori if the value under key `c` will be equal to `d` or `e`.
```
aboguszewski-sumo, line 180: There is one issue that I encountered and I think that we should definitely make it a UB.

sumo-drosiek: what is happening in current implementation?
IMHO, We shouldn't override existing keys

aboguszewski-sumo: In general, keys are overwritten. If there was a map, the maps are merged (for example `"a.b": "c"` and `"a": {"d":"e"}` get merged into `"a": {"b": "c", "d":"e"}`).
The problem is that we can't easily decide whether we will be overwriting a key, because it depends on the order in which `Range` method of `pcommon.Map` traverses the attributes. We would have to make some kind of preprocessing to check all existing keys.

pkg/processor/sumologicschemaprocessor/README.md
```diff
@@ -32,6 +32,17 @@ processors:
...
+    # Specifies if attributes should be nested, basing on their keys.
+    # See "Nesting attributes" documentation chapter from this document.
+    nest_attributes:
+      # Defines whether attributes should be nested.
+      # default = false
+      enabled: {true, false}
+
+      # Defines the string used to separate key names in attributes that are to be nested.
+      # default = "."
+      separator: <separator>
```
sumo-drosiek, line 45: In my opinion, we could have some pattern matching here. So in case customer wants to leave some keys which has separator in it he can do i, but current behavior seems to be reasonable for first iteration

pkg/processor/sumologicschemaprocessor/README.md
sumo-drosiek: makes sense, I think for the first iteration it should be fine. Let's follow the KISS principle

pkg/processor/sumologicschemaprocessor/README.md
aboguszewski-sumo: This has been already commented this on Slack. IMO we should add this in another PR (it is also arguable what should the filters be: we can do an allowlist or a denylist, with advanced matching options or not... imo it's the best to just add a denylist of prefixes)

sumo-drosiek (approved):

PR #880, opened 2022-12-23
feat(sumologicschemaprocessor): add allowlist and denylist to nesting processor

aboguszewski-sumo:
Updates #865 

Nesting subprocessor has gained two config options. It now supports an allowlist and a denylist of prefixes for which the attributes should or should not be translated.


sumo-drosiek (approved):
LGTM

PR #881, opened 2022-12-23
feat(sumologicschemaprocessor): add squashing single values in nesting processor

aboguszewski-sumo:
Updates #865 

This PR adds a feature to nesting processor that squashes maps that have only one key-value pair. For example:
```
{
"k8s": {
  "pods": {
    "a" : "A",
    "b": "B"
    }
  }
}
```
gets squashed into
```
{
"k8s.pods": {
    "a" : "A",
    "b": "B"
    }
}
```

This feature is **disabled** by default. 

Two things to note:
- Perhaps this implementation contains too much map copying, but I wanted to finish this before Christmas and eventually fix later. 
- I tried to create an algorithm that does this while doing the initial nesting, but I found it to be too complex and not necessarily faster.


sumo-drosiek (approved):
:shrug:

PR #891, opened 2023-01-12
chore: update OT core to 0.69.0

swiatekm-sumo:
I'm not sure why `staticcheck` didn't find uses of deprecated `config.*Settings` fields.

There's one other problem here. Some our components use a [test package](https://github.com/open-telemetry/opentelemetry-collector/tree/v0.69.0/otelcol/otelcoltest) from upstream, and this test package has a dependency on the [`zpagesextension`](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.69.0/extension/zpagesextension/go.mod#L8), whose dependencies weren't upgraded in `0.69.0` core. I don't understand go packaging with multiple replaces that well, but the end result was that the tests wanted to use version `0.68.0` of that package, which isn't compatible with `0.69.0`. So I had to manually replace it in each component that uses it. Not sure if there's a more graceful way.


otelcolbuilder/cmd/configprovider.go
```diff
@@ -38,8 +38,6 @@ import (
 func UseCustomConfigProvider(params *otelcol.CollectorSettings) error {
 	// feature flags, which are enabled by default in our distro
 	err := featuregate.GetRegistry().Apply(map[string]bool{
-		"receiver.apache.emitServerNameAsResourceAttribute":            true,
-		"receiver.apache.emitPortAsResourceAttribute":                  true,
 		"receiver.elasticsearch.emitClusterHealthDetailedShardMetrics": true,
 		"receiver.elasticsearch.emitAllIndexOperationMetrics":          true,
 	})
```
aboguszewski-sumo, line 43: The feature gates for elasticsearch are in beta now (enabled by default), so technically we could remove this, but I'm not sure if in this PR.

pkg/processor/k8sprocessor/go.mod
```diff
@@ -104,3 +104,8 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+// This component uses the otelcoltest package, which depends on zpagesextension, but uses a replace to refer to it in its go.mod
+// As a result, we pull the wrong version here without this replace
+// This can be removed once upstream updates the versions in zpagesextension go.mod or removes the dependency
+replace go.opentelemetry.io/collector/extension/zpagesextension => go.opentelemetry.io/collector/extension/zpagesextension v0.69.0
```
aboguszewski-sumo, line 111: Isn't this the same problem as with importing `sumologicextension` in `sumologicexporter`? I remember having similar issues when I used `sumologicexporter` externally. Idk if there is a reasonable solution for that, doesn't seem so.

aboguszewski-sumo (approved):
lgtm, but read my comment about feature gates

pkg/processor/k8sprocessor/go.mod
swiatekm-sumo: There probably isn't, but we have a new release fixing this here: https://github.com/open-telemetry/opentelemetry-collector/releases/tag/cmd%2Fbuilder%2Fv0.69.1, so I'll update this PR.

swiatekm-sumo: Actually, it looks like 0.69.1 is not out for contrib, so we'd need to have two different versions for different components. I'll just leave this as is and add an issue to remove the replaces for 0.70.0.

swiatekm-sumo: https://github.com/SumoLogic/sumologic-otel-collector/issues/894

otelcolbuilder/cmd/configprovider.go
aboguszewski-sumo: https://github.com/SumoLogic/sumologic-otel-collector/issues/896

PR #893, opened 2023-01-12
Collector metadata tags auto-discovery (local services, e.g. mysql)

portertech:
This pull-request demonstrates how easy it is to discover local processes, filter the process list, and create the appropriate collector metadata tags in order to drive Sumo platform "insights" (e.g. "We've noticed that these hosts are running MySQL, would you like to install the Sumo MySQL app to monitoring them?"). This feature leverages the existing/new Sumo metadata API.

This pull-request currently produces auto-discovery tags like these on my development host:

```
{
  "sumo.disco.enabled": "true",
  "sumo.disco.docker": "",
  "...": "..."
}
```


pkg/extension/sumologicextension/config.go
```diff
@@ -49,6 +49,9 @@ type Config struct {
 	// https://help.sumologic.com/docs/manage/fields
 	CollectorFields map[string]interface{} `mapstructure:"collector_fields"`
 
+	// DiscoverCollectorTags enables collector metadata tag auto-discovery.
+	DiscoverCollectorTags bool `mapstructure:"discover_collector_tags"`
```
portertech, line 53: Not sure if we want to first introduce this feature w/ "Tags" or "Fields" 🤔

portertech: Team confirmed that "Tags" is the way 👍

perk-sumo (approved):
LGTM pending README with information that this is an alpha feature :)

PR #902, opened 2023-01-23
docs: FIPS support

swiatekm-sumo:
Officially claim FIPS 140-3 support. Add a document explaining how the binaries can be obtained, and how to verify their correctness.

Delete remaining references to FIPS builds on darwin. BoringCrypto is only certified for Linux on amd64.


docs/fips.md
```diff
@@ -0,0 +1,194 @@
...
+While [FIPS 140-3][fips] doesn't, strictly speaking, mandate a specific TLS version, it is nonetheless a good practice to use as recent of a version as possible. Otelcol allows TLS to be configured for each component individually, in a standardized manner. You can find the details [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/README.md).
+
+For example, here is how we can set a minimum TLS version of 1.2 in the `otlp` exporter:
+
+```yaml
+exporters:
+  otlp:
+    endpoint: myserver.local:55690
+    tls:
+      min_version: "1.2"
```
perk, line 191: Isn't that the default actually?
If yes I think we should call it out.

Right now, the way it's written one might understand that they need to set it explicitly to have TLS 1.2

perk (approved):
LGTM except for one comment.

docs/fips.md
swiatekm-sumo: It is the default, yeah. I'll call it out as you suggest.

swiatekm-sumo: Done.

perk-sumo (approved):
LGTM 🚀

docs/fips.md
```diff
@@ -0,0 +1,195 @@
+# BoringCrypto and FIPS 140-3 compliance
```
dochel, line 1: Where did you get FIPS 140-3 from? The last module validation of Boring I see is still against FIPS 140-2? (https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4407)

Maybe just say "FIPS compliance" and omit the 140-3?

docs/fips.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Introduction
+
+Sumo OT distro can be built to use BoringCrypto instead of Go standard library crypto implementations.
```
dochel, line 5: Maybe add something like "BoringCrypto is a cryptographic module validated against FIPS 140."

(Would it help to add a link to https://csrc.nist.gov/projects/cryptographic-module-validation-program/validated-modules/search?SearchMode=Basic&ModuleName=boring&CertificateStatus=Active&ValidationYear=0?)

docs/fips.md
```diff
@@ -0,0 +1,195 @@
...
+```text
+docker pull public.ecr.aws/sumologic/sumologic-otel-collector:0.69.0-sumo-0-fips
+```
+
+> **Note**
+> Currently FIPS-compliant builds are only available for Linux on amd64.
+
+## Running in FIPS-approved mode
+
+The FIPS-compliant build statically enforces use of [FIPS 140-3][fips] compliant TLS settings, and doesn't require any
```
dochel, line 48: This is confusing, as it conflates two different things: FIPS 140 validation of the BoringCrypto module, and settings that ensure that TLS 1.2 is used only with cipher suites that use approved primitives. (FIPS 140 does not regulate TLS.)

Maybe say something like "The FIPS-compliant build statically enforces the use of FIPS-approved algorithms in TLS cipher suites, and..." ?

docs/fips.md
```diff
@@ -0,0 +1,195 @@
...
+client curve: secp384r1 (24)
+client curve: secp521r1 (25)
+client point: uncompressed (0)
+
+[...]
+```
+
+## Setting minimum TLS version
+
+While [FIPS 140-3][fips] doesn't, strictly speaking, mandate a specific TLS version, it is nonetheless a good practice to use as recent of a version as possible. Otelcol allows TLS to be configured for each component individually, in a standardized manner. You can find the details [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/README.md).
```
dochel, line 182: Specifically, versions earlier than 1.2 are considered unsafe. You might find differing opinions in the security/compliance community over whether 1.2 or 1.3 is the best option, currently.

docs/fips.md
```diff
@@ -0,0 +1,195 @@
...
+
+By default, the minimum TLS version for otel components is `1.2`. If you want to set a higher one, here's an example of using `1.3`
+for the `otlp` exporter:
```
dochel, line 185: Interesting discussion here that hints that 1.3 might use non-approved algorithms? I know nothing about it, just happened to come across this write-up: https://kupczynski.info/posts/fips-golang/

Doesn't invalidate what you're saying, though. :)

dochel (commented):
Added some comments...

docs/fips.md
swiatekm-sumo: I swear I saw a cert for 140-3, but it's clearly not actually there. Sorry for the confusion.

docs/fips.md
swiatekm-sumo: I specifically noted that `1.2` is safe to use, and that the status of `1.3` is uncertain as of right now.

dochel (approved):
Looks good to me. Thank you! :)

PR #909, opened 2023-01-27
fix(install/script): config dir permissions

swiatekm-sumo:
Directories need a +x permission. Fixes a regression introduced by #906


mat-rumian (approved):
LGTM

PR #910, opened 2023-01-27
fix(scripts/install.sh): use a safe dir for temp files

swiatekm-sumo:
Storing temporary files in `/tmp/` in general is unsafe. A malicious user can prepare files there and make us install their own binary. We've also been putting backups of config files in the working directory, which can leak secrets.

To prevent this, use a safe directory for temporary files. 

As we have a feature that can keep downloaded binaries after installation, I put those in `/var/cache/otelcol-sumo`, which is a standard location for system caches, and only writable by root.


sumo-drosiek (approved):

aiwaniuk (approved):
Thanks!

PR #915, opened 2023-01-31
chore: update OT core to v0.70.0

aboguszewski-sumo:
Important changes:
- `pdata.map.Sort()` was deprecated, for example source processor's tests were affected
- `otelcolbuilder`'s CLI options were deprecated, they have been moved to `otelcol-builder.yaml`


sumo-drosiek (approved):
LGTM

PR #918, opened 2023-01-31
feat(sumologicextension): use hostname as default collector name

swiatekm-sumo:
This ends up causing more problems than it solves. In most circumstances the hostname is unique, and where it isn't, the user should set the collector name explicitly. And just using the hostname makes the user experience in the majority of cases better.


sumo-drosiek (approved):
Should we clobber by default then?

swiatekm-sumo:
> Should we clobber by default then?

I think the default behaviour is better in this case. Clobber would actually delete the existing collector, which is a destructive action I'd rather not do without explicit user opt-in. Maybe we should log a warning in this case?

swiatekm-sumo:
> > Should we clobber by default then?
> 
> I think the default behaviour is better in this case. Clobber would actually delete the existing collector, which is a destructive action I'd rather not do without explicit user opt-in. Maybe we should log a warning in this case?

I added it.

pkg/extension/sumologicextension/extension.go
```diff
@@ -435,6 +434,10 @@ func (se *SumologicExtension) registerCollector(ctx context.Context, collectorNa
 		return credentials.CollectorCredentials{}, err
 	}
 
+	if collectorName != resp.CollectorName {
+		se.logger.Warn("Collector name already in use, registered modified name", zap.String("registered_name", resp.CollectorName))
+	}
+
```
sumo-drosiek, line 440: Nice, but not sure if we should mention `clobber` or any documentation here

swiatekm-sumo: Hm, I probably should add this behaviour to the documentation to begin with. I don't want to make this log message too long.

swiatekm-sumo: Updated documentation.

sumo-drosiek (approved):

Issue #879, opened 2022-12-22
[receiver/rawk8sevents]: fix flaky tests


aboguszewski-sumo:
One of the tests fails with a terrifying frequency:
```
--- FAIL: TestProcessEventE2E (0.02s)
    receiver_test.go:124: 
        	Error Trace:	D:\a\sumologic-otel-collector\sumologic-otel-collector\pkg\receiver\rawk8seventsreceiver\receiver_test.go:124
        	            				D:\a\sumologic-otel-collector\sumologic-otel-collector\pkg\receiver\rawk8seventsreceiver\asm_amd64.s:1594
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 0
        	Test:       	TestProcessEventE2E
FAIL
FAIL	github.com/SumoLogic/sumologic-otel-collector/pkg/receiver/rawk8seventsreceiver	0.233s
```
example failed run: https://github.com/SumoLogic/sumologic-otel-collector/actions/runs/3757804988/jobs/6385415495

PR #919, opened 2023-01-31
test(rawk8seventsreceiver): fix flaky test

swiatekm-sumo:
This test was flaky because it was erroneously using an assertion inside an `assert.Eventually` call. If the inner assertion failed even once, the whole test would fail.


sumo-drosiek (approved):

aboguszewski-sumo:
Fixes #879

PR #920, opened 2023-02-01
fix(scripts/install.sh): config dir permissions

swiatekm-sumo:
Set permissions for config directory in `/etc/` before putting config files there. `/etc/` is world-readable by default, and credentials can be leaked this way.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -557,6 +557,13 @@ function setup_config() {
     echo -e "Creating user env directory (${USER_ENV_DIRECTORY})"
     mkdir -p "${USER_ENV_DIRECTORY}"
 
+    echo 'Changing permissions for config files and storage'
+    chmod 551 "${CONFIG_DIRECTORY}"  # config directory world traversable, as is the /etc/ standard
+
+    echo 'Changing permissions for user env directory'
+    chmod 550 "${USER_ENV_DIRECTORY}"
+    chmod g+s "${USER_ENV_DIRECTORY}"
+
     echo "Generating configuration and saving as ${CONFIG_PATH}"
 
     CONFIG_URL="https://raw.githubusercontent.com/SumoLogic/sumologic-otel-collector/${CONFIG_BRANCH}/examples/sumologic.yaml"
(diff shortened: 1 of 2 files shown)


sumo-drosiek (approved):

scripts/install.sh
```diff
@@ -557,6 +557,13 @@ function setup_config() {
     echo -e "Creating user env directory (${USER_ENV_DIRECTORY})"
     mkdir -p "${USER_ENV_DIRECTORY}"
 
+    echo 'Changing permissions for config files and storage'
+    chmod 551 "${CONFIG_DIRECTORY}"  # config directory world traversable, as is the /etc/ standard
```
aiwaniuk, line 561: does it block exploit from ticket?
if "all" have still "x", I belive it will still work
I think access was blocked aftert installation by this line:
```
chmod -R 440 "${CONFIG_DIRECTORY}"/*
```
which removed "x" from USER_CONFIG_DIRECTORY
and this line is still after writing token to file

aiwaniuk (changes requested):
I think ti may not work

scripts/install.sh
swiatekm-sumo: Files and directories created in this directory will have 551 set as their permissions by default. So an unprivileged user will be able to traverse the directories, but not read the files.

At least that's the theory, in practice I think something else is messing with the permissions, as the exploit still works. I'll try to figure out why.

swiatekm-sumo: Turns out curl creates files with 644 permissions by default: https://curl.se/libcurl/c/CURLOPT_NEW_FILE_PERMS.html. 

I'll just set the variable for the whole script.

swiatekm-sumo: Ok, I've explicitly set the ACLs for `/etc/otelcol-sumo/conf.d/common.yaml` after it's created, but before it has the token written into it. I tested your simple exploit on it and it doesn't work anymore.

aiwaniuk: I did some addiational verification what permssions will have '.bak' file in this case
```
root@h1aaah1:/etc/otelcol-sumo/conf.d# sed -i.bak -e "s/install_token:.*$/install_token: aaa/" common.yaml
root@h1aaah1:/etc/otelcol-sumo/conf.d# ls -al
total 16
dr-xr-x--- 2 otelcol-sumo otelcol-sumo 4096 Feb  1 15:09 .
dr-xr-x--x 4 root         root         4096 Jan 31 13:57 ..
-r--r----- 1 otelcol-sumo otelcol-sumo  106 Feb  1 15:09 common.yaml
-r--r----- 1 otelcol-sumo otelcol-sumo  106 Feb  1 15:09 common.yaml.bak
```
and looks good, looks like sed will set the same permssion as original file

aiwaniuk (approved):

PR #921, opened 2023-02-01
chore: use go dns for all binaries

sumo-drosiek:

https://pkg.go.dev/net#hdr-Name_Resolution

### DIFF
--- a/otelcolbuilder/Makefile
+++ b/otelcolbuilder/Makefile
@@ -43,6 +43,11 @@ else
 LDFLAGS="-w"
 endif
 
+# We want to enforce go dns for all types of binaries, as cgo has been used by fips binaries (which supports CGO)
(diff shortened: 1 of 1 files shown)


otelcolbuilder/Makefile
```diff
@@ -83,7 +83,7 @@ _builder:
 .PHONY: _gobuild
 _gobuild:
 	(cd cmd && \
-		CGO_ENABLED=$(CGO_ENABLED) go build -v \
+		GODEBUG=netdns=go CGO_ENABLED=$(CGO_ENABLED) go build -v \
```
swiatekm-sumo, line 86: I'd prefer if we had the `GODEBUG` variable defined at the top of the Makefile, and add a comment there explaining why we set this.

swiatekm-sumo (approved):

PR #926, opened 2023-02-02
feat(scripts/install.sh): don't pass install token via command line

swiatekm-sumo:
I'm leaving the actual option in, as it's currently still displayed in the UI, but I'm removing it from documentation and manifests for Chef, Puppet and Ansible.

### DIFF
--- a/examples/chef/sumologic-otel-collector/resources/default.rb
+++ b/examples/chef/sumologic-otel-collector/resources/default.rb
@@ -27,6 +27,7 @@ action :default do
   install_command = get_install_script_command(new_resource)
   execute INSTALL_SCRIPT_PATH do
     command install_command
+    environment ({"SUMOLOGIC_INSTALL_TOKEN" => resource.install_token})
   end
   run_action :prepare_config
   if new_resource.systemd_service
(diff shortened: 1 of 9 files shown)


aiwaniuk (approved):
Thanks

swiatekm-sumo:
Modified the tests as well.

examples/chef/sumologic-otel-collector/resources/default.rb
```diff
@@ -58,7 +58,7 @@
 
 
 def get_install_script_command(resource)
-  command_parts = ["bash", INSTALL_SCRIPT_PATH, "--installation-token #{resource.install_token}"]
+  command_parts = ["SUMOLOGIC_INSTALL_TOKEN=#{resource.install_token}", "bash", INSTALL_SCRIPT_PATH]
```
sumo-drosiek, line 61: is it working if we pass `ENV` in chef as normal command?

sumo-drosiek (approved):
LGTM, one chef question

examples/chef/sumologic-otel-collector/resources/default.rb
swiatekm-sumo: That's supposed to work, but I replaced it with a more idiomatic approach from Chef's docs.

PR #931, opened 2023-02-03
fix(ci): fix failing devbuilds

aboguszewski-sumo:
Working run: https://github.com/SumoLogic/sumologic-otel-collector/actions/runs/4085445456

Tested locally and pulling both images works for amd64 (both fips and normal, tested with `docker pull public.ecr.aws/sumologic/sumologic-otel-collector-dev:70c26f6906b46795a0850d42c3e9848e7af32920`)


swiatekm-sumo (approved):
LGTM

But we should rethink how this is done, ideally just use `docker buildx bake`.

PR #938, opened 2023-02-07
fix(k8sprocessor): race condition when getting Pod data

swiatekm-sumo:
We had a race condition where we'd update Pod metadata in `GetPod` while only holding a read lock. In circumstances where two batches of data would arrive from the same Pod in quick succession, we'd get a panic due to concurrent map writes.

I fixed this by changing the interface to only return an attributes map and explicitly making that a copy of what's in the Pod attributes.

### DIFF
--- a/pkg/processor/k8sprocessor/kube/client.go
+++ b/pkg/processor/k8sprocessor/kube/client.go
@@ -247,21 +247,42 @@ func (c *WatchClient) deleteLoop(interval time.Duration, gracePeriod time.Durati
 	}
 }
 
-// GetPod takes an IP address or Pod UID and returns the pod the identifier is associated with.
-func (c *WatchClient) GetPod(identifier PodIdentifier) (*Pod, bool) {
+// getPod takes an IP address or Pod UID and returns the pod the identifier is associated with.
+func (c *WatchClient) getPod(identifier PodIdentifier) (*Pod, bool) {
 	c.m.RLock()
 	defer c.m.RUnlock()
 	pod, ok := c.Pods[identifier]
-	if ok {
-		if pod.Ignore {
-			return nil, false
-		}
-
-		c.updatePodOwnerMetadata(pod)
-		return pod, ok
+	if !ok {
+		observability.RecordIPLookupMiss()
+		return nil, ok
 	}
-	observability.RecordIPLookupMiss()
-	return nil, false
+	if pod.Ignore {
+		return nil, false
+	}
+	return pod, true
+}
+
+// GetPodAttributes takes an IP address or Pod UID and returns the metadata attributes of the Pod the
+// identifier is associated with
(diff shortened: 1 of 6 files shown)


pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -247,21 +247,35 @@ func (c *WatchClient) deleteLoop(interval time.Duration, gracePeriod time.Durati
...
+	attributes := make(map[string]string, len(pod.Attributes))
+	for key, value := range pod.Attributes {
+		attributes[key] = value
+	}
+	for key, value := range c.getPodOwnerMetadataAttributes(pod) {
+		attributes[key] = value
 	}
-	observability.RecordIPLookupMiss()
-	return nil, false
+	return attributes, ok
```
sumo-drosiek, line 278: Could we pass target map to this function so we avoid double copying?

pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -247,21 +247,35 @@ func (c *WatchClient) deleteLoop(interval time.Duration, gracePeriod time.Durati
...
+	}
+	if pod.Ignore {
+		return nil, false
+	}
+	return pod, true
+}
 
-		c.updatePodOwnerMetadata(pod)
-		return pod, ok
+// GetPod takes an IP address or Pod UID and returns the pod the identifier is associated with.
```
sumo-drosiek, line 265: Please update description

pkg/processor/k8sprocessor/kube/client.go
swiatekm-sumo: I'd rather not unless we can show a performance increase. Doing it this way is easier to understand, and I'd be surprised if copying like 16 strings twice makes any significant difference.

Either way, for this fix I'd rather keep things simple, we can revisit this later.

pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -247,21 +247,42 @@ func (c *WatchClient) deleteLoop(interval time.Duration, gracePeriod time.Durati
...
+	pod, ok := c.getPod(identifier)
+	if !ok {
+		return nil, false
 	}
-	observability.RecordIPLookupMiss()
-	return nil, false
+	ownerAttributes := c.getPodOwnerMetadataAttributes(pod)
+
+	// we need to take a lock here because pod.Attributes may be modified concurrently
+	// TODO: clean up the locking in these functions
```
sumo-drosiek, line 275: which functions?

sumo-drosiek (approved):

pkg/processor/k8sprocessor/kube/client.go
swiatekm-sumo: This one and the ones it calls, made that clearer in the comment.

PR #942, opened 2023-02-07
Revert "chore(otelcolbuilder): move deprecated CLI options to otelcol-builder.yaml"

swiatekm-sumo:
This reverts commit 7b02da720c7746915815516d367a109525d58acb.

This doesn't actually work, and the version ends up as literally `$(VERSION)$(FIPS_SUFFIX)`. Need to find a different way to set these variables.

### DIFF
--- a/otelcolbuilder/.otelcol-builder.yaml
+++ b/otelcolbuilder/.otelcol-builder.yaml
@@ -10,11 +10,6 @@ dist:
   # the path to write the output (sources and binary).
   output_path: ./cmd
 
-  # These options were formerly CLI arguments of otelcol-builder
-  # They are being set in Makefile.
-  go: $(GO)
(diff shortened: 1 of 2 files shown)


sumo-drosiek (approved):

sumo-drosiek:
maybe we should use `${VERSION}` instead of `$(VERSION)` ?

swiatekm-sumo:
> maybe we should use `${VERSION}` instead of `$(VERSION)` ?

There's probably other ways of doing this, for now I'm just reverting things to a known working state.

sumo-drosiek:
> > maybe we should use `${VERSION}` instead of `$(VERSION)` ?
> 
> There's probably other ways of doing this, for now I'm just reverting things to a known working state.

doesn't work as well

PR #953, opened 2023-02-09
feat(scripts/install.sh): add information after uninstallation

sumo-drosiek:
Adds`Uninstallation completed` after uninstallation process

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -671,7 +671,6 @@ function uninstall() {
     fi
 
     echo "Uninstallation completed"
-    exit 0
 }
 
 function escape_sed() {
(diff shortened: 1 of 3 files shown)


swiatekm-sumo (approved):

scripts/install.sh
```diff
@@ -670,6 +670,7 @@ function uninstall() {
         fi
     fi
 
+    echo "Uninstallation completed"
     exit 0
```
swiatekm-sumo, line 674: I think this exit is redundant. We exit after returning from this function.

sumo-drosiek, line 674:
```suggestion
    echo "Uninstallation completed"
```

sumo-drosiek: Agree, removed

PR #955, opened 2023-02-09
docs(changelog): add information about adding syslog exporter

kkujawa-sumo:


aboguszewski-sumo (approved):
Are we ready to fill info about this component in other docs? (for example in the table in `README.md`)

PR #956, opened 2023-02-09
fix(syslogexporter) remove drop_invalid_messages parameter from example config

kkujawa-sumo:
`drop_invalid_messages` parameter is not supported by syslogexporter


swiatekm-sumo:
I don't think you need a changelog entry for this.

sumo-drosiek (approved):

PR #957, opened 2023-02-09
feat(syslogexporter): add support for endpoint set to IP address

kkujawa-sumo:


swiatekm-sumo (approved):
LGTM. I think you should skip the changelog entry though, or just add this PR link to the entry about adding the exporter.

PR #960, opened 2023-02-09
feat(scripts/install): use sed instead of xargs

sumo-drosiek:


aiwaniuk (approved):
Thanks!

swiatekm-sumo (approved):

PR #963, opened 2023-02-09
chore(syslogexporter): README updates for the syslog exporter

rnishtala-sumo:
README updates for the syslog exporter

### DIFF
--- a/pkg/exporter/syslogexporter/README.md
+++ b/pkg/exporter/syslogexporter/README.md
@@ -1,29 +1,46 @@
 # Syslog Exporter
 
-Export/Forward syslog messages to a syslog server
-
 **Stability level**: Experimental
 
-This exporter can forward syslog messages to a third party [syslog server](https://www.rsyslog.com/)
+## About The Exporter
 
-This exporter/forwarder also supports sending syslog messages to the [Cloud Syslog](https://help.sumologic.com/docs/send-data/hosted-collectors/cloud-syslog-source/).
+The syslog exporter/forwarder supports sending messages to a remote syslog server.
 
-Configuration is specified via the yaml in the following structure:
+- This exporter/forwarder can forward syslog messages to a third party [syslog server](https://www.rsyslog.com/) using RFC5424 and RFC3164
+- It also supports sending syslog messages to the [Cloud Syslog Source](https://help.sumologic.com/docs/send-data/hosted-collectors/cloud-syslog-source/) configured on a Sumo Logic hosted collector using the RFC5424 format
+- The syslog exporter must be used with the [syslog_parser](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/operators/syslog_parser.md)
+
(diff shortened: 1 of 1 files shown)


pkg/exporter/syslogexporter/README.md
```diff
@@ -1,29 +1,46 @@
...
 
-Configuration is specified via the yaml in the following structure:
+Please refer to the yaml below to configure the syslog exporter:
 
 ```yaml
 extensions:
   file_storage/syslog:
-    directory: /tmp/otc
+    directory: /tmp/otc # Please ensure this directory exists
     timeout: 10s
```
swiatekm-sumo, line 34: I don't think we need the storage configuration here. It's not really directly related to the exporter.

pkg/exporter/syslogexporter/README.md
```diff
@@ -1,29 +1,46 @@
 # Syslog Exporter
 
-Export/Forward syslog messages to a syslog server
-
 **Stability level**: Experimental
 
-This exporter can forward syslog messages to a third party [syslog server](https://www.rsyslog.com/)
+## About The Exporter
+
+The syslog exporter/forwarder supports sending messages to a remote syslog server.
```
swiatekm-sumo, line 7: I'd just avoid calling it a `forwarder` anywhere, I think that's more confusing than helpful.

pkg/exporter/syslogexporter/README.md
```diff
@@ -1,29 +1,46 @@
...
+
+- This exporter/forwarder can forward syslog messages to a third party [syslog server](https://www.rsyslog.com/) using RFC5424 and RFC3164
+- It also supports sending syslog messages to the [Cloud Syslog Source](https://help.sumologic.com/docs/send-data/hosted-collectors/cloud-syslog-source/) configured on a Sumo Logic hosted collector using the RFC5424 format
+- The syslog exporter must be used with the [syslog_parser](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/operators/syslog_parser.md)
```
swiatekm-sumo, line 11: Are we planning to document what the actual requirements are later, along with having configurable defaults for the missing fields?

rnishtala-sumo: As of now, the `syslog_parser` is the only identified dependency/requirement that needs to be configured in the receiver. Which missing fields are you referring to?

pkg/exporter/syslogexporter/README.md
rnishtala-sumo: Yes it is optional, so should we remove all references to the storage config from the readme?

pkg/exporter/syslogexporter/README.md
swiatekm-sumo: What I mean is that the exporter will work even if the syslog processor is not present, it'll simply use default values for properties. I think we should document this fact, rather than just saying it "must be used" with the processor.

PR #964, opened 2023-02-10
feat(scripts/install):do not fail if unable to get available versions

sumo-drosiek:
This ignores curl errors as we fallback to different methods of getting latest version

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1165,7 +1165,8 @@ INSTALLED_VERSION="$(get_installed_version)"
 echo -e "Installed version:\t${INSTALLED_VERSION:-none}"
 
 echo -e "Getting versions..."
-VERSIONS="$(get_versions)"
+# Get versions, but ignore errors are we fallback to other methods later
+VERSIONS="$(get_versions || echo "")"
 
 # Use user's version if set, otherwise get latest version from API (or website)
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -403,7 +403,7 @@ function get_versions() {
     | sed 'N;N;s/.*true.*//' \
     | grep -o 'v.*"' \
     | sort -r \
-    | sed 's/^v//;s/"$//'
+    | sed 's/^v//;s/"$//' || echo ""
```
aboguszewski-sumo, line 406: wouldn't `|| true` also work?
https://stackoverflow.com/a/36131305

sumo-drosiek: not entirely sure if it will work for all environments. We have mixed `true` and `echo ""` across the script so decided to use `echo` way.

aboguszewski-sumo: > not entirely sure if it will work for all environments

I love bash.
If we already have that mixed then ok, let's leave it, but I think we should make a standard for it at some point

aboguszewski-sumo (approved):

sumo-drosiek:
I moved silencing error outside of function. It is more logical with comment

swiatekm-sumo (approved):

PR #969, opened 2023-02-13
feat(sumologicextension): deprecate install_token in favor of installation_token

sumo-drosiek:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- feat(sumologicextension): deprecate `install_token` [#969]
+
+[#969]: https://github.com/SumoLogic/sumologic-otel-collector/pull/969
+[unreleased]: https://github.com/SumoLogic/sumologic-otel-collector/compare/v0.71.0-sumo-0...main
+
 ## [v0.71.0-sumo-0]
 
 ### Released 2023-02-09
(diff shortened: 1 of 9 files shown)


astencel-sumo:
Should we use a feature gate for this breaking change? This is recommended by upstream here: https://github.com/open-telemetry/opentelemetry-collector/blob/v0.71.0/CONTRIBUTING.md#breaking-changes.

> When deprecating a feature affecting end-users, consider first deprecating the feature in one version, then hiding it behind a [feature flag](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/service/featuregate/) in a subsequent version, and eventually removing it after yet another version.

sumo-drosiek:
> Should we use a feature gate for this breaking change? This is recommended by upstream here: https://github.com/open-telemetry/opentelemetry-collector/blob/v0.71.0/CONTRIBUTING.md#breaking-changes.
> 
> > When deprecating a feature affecting end-users, consider first deprecating the feature in one version, then hiding it behind a [feature flag](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/service/featuregate/) in a subsequent version, and eventually removing it after yet another version.

Technically this change itself is not a breaking change. Removing `install_token` will be a breaking change, and we can do it with `feature_flag`

I don't want to do a rename as a breaking change, because there will be problem with synchronising this change with UI, install script and documentation changes.

astencel-sumo:
We decided to not use a feature flag. Changing the config option `install_token` to `installation_token` is probably even easier than adding a feature flag CLI option to the binary invocation, so it doesn't seem to be worth it.

astencel-sumo (commented):
So you're adding a new configuration option `installation_token` but not updating the docs? :thinking:

astencel-sumo (approved):

PR #971, opened 2023-02-13
feat: rename install_token to installation token

sumo-drosiek:
This is continuation of #969

ToDo:

- support for windows both install_token and installation_token
  - [x] As I understand, we are not comparing installation tokens (previous and current installation) right now, so skipping this point
  - [x] Supporting to run in poershell with both `INSTALLTOKEN` and `INSTALLATIONTOKEN`
- update scripts_tests
  - [x] updated
  - [x] tested

### DIFF
--- a/pkg/scripts_test/config.go
+++ b/pkg/scripts_test/config.go
@@ -18,6 +18,7 @@ type extensions struct {
 
 type sumologicExtension struct {
 	InstallToken string            `yaml:"install_token"`
+	InstallationToken string       `yaml:"installation_token"`
 	Tags         map[string]string `yaml:"collector_fields"`
 	APIBaseURL   string            `yaml:"api_base_url"`
 }
--- a/pkg/scripts_test/config.go
+++ b/pkg/scripts_test/config.go
@@ -17,10 +17,10 @@ type extensions struct {
 }
 
 type sumologicExtension struct {
-	InstallToken string            `yaml:"install_token"`
-	InstallationToken string       `yaml:"installation_token"`
-	Tags         map[string]string `yaml:"collector_fields"`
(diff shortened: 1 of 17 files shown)


pkg/scripts_test/config.go
```diff
@@ -17,9 +17,10 @@ type extensions struct {
 }
 
 type sumologicExtension struct {
-	InstallToken string            `yaml:"install_token"`
-	Tags         map[string]string `yaml:"collector_fields"`
-	APIBaseURL   string            `yaml:"api_base_url"`
+	InstallToken      string            `yaml:"install_token,omitempty"`
```
rnishtala-sumo, line 20: should we remove the install_token entry?

scripts/install.sh
```diff
@@ -766,8 +766,16 @@ function get_user_config() {
...
+    # extract installation_token and strip quotes
+    # fallback to deprecated install_token
+    grep -m 1 installation_token "${file}" \
+        | sed 's/.*installation_token:[[:blank:]]*//' \
+        | sed 's/[[:blank:]]*$//' \
+        | sed 's/^"//' \
+        | sed "s/^'//" \
+        | sed 's/"$//' \
+        | sed "s/'\$//" \
+    || grep -m 1 install_token "${file}" \
```
rnishtala-sumo, line 778: why do we need two `greps` here and not just `grep` for `installation token` in the file?

sumo-drosiek: Backward compatibility. We want to support previous installations

pkg/scripts_test/config.go
sumo-drosiek: I need it to test backward compatibility

rnishtala-sumo (approved):

PR #972, opened 2023-02-13
refactor(scripts/install): do not expose token in sed command

sumo-drosiek:
Do not expose token in sed command as it can be saw on processes list

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -887,6 +887,7 @@ ${indentation}sumologic:/" "${file}"
 
 # write installation token to user configuration file
 function write_installation_token() {
+    set -x
     local token
     readonly token="${1}"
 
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -898,12 +899,13 @@ function write_installation_token() {
 
     # ToDo: ensure we override only sumologic `install_token`
     if grep "install_token" "${file}" > /dev/null; then
-        sed -i.bak -e "s/install_token:.*$/install_token: $(escape_sed "${token}")/" "${file}"
+        echo "s/install_token:.*$/install_token: $(escape_sed "${token}")/" | sed -i.bak -f - "${file}"
```
swiatekm-sumo, line 902: Can you add a comment explaining why this is done the way it is?

scripts/install.sh
```diff
@@ -898,12 +899,13 @@ function write_installation_token() {
...
     else
         # write installation token on the top of sumologic: extension
-        sed -i.bak -e "s/sumologic:/sumologic:\\
-\\${ext_indentation}install_token: $(escape_sed "${token}")/" "${file}"
+        echo "s/sumologic:/sumologic:\\
```
sumo-drosiek, line 905:
```suggestion
        Do not expose token in sed command as it can be saw on processes list
        echo "s/sumologic:/sumologic:\\
```

scripts/install.sh
```diff
@@ -898,12 +899,15 @@ function write_installation_token() {
...
     # ToDo: ensure we override only sumologic `install_token`
     if grep "install_token" "${file}" > /dev/null; then
-        sed -i.bak -e "s/install_token:.*$/install_token: $(escape_sed "${token}")/" "${file}"
+        # Do not expose token in sed command as it can be saw on processes list
+        echo "s/install_token:.*$/install_token: $(escape_sed "${token}")/" | sed -i.bak -f - "${file}"
     else
         # write installation token on the top of sumologic: extension
-        sed -i.bak -e "s/sumologic:/sumologic:\\
-\\${ext_indentation}install_token: $(escape_sed "${token}")/" "${file}"
+        Do not expose token in sed command as it can be saw on processes list
```
sumo-drosiek, line 906:
```suggestion
        # Do not expose token in sed command as it can be saw on processes list
```

swiatekm-sumo (approved):

aiwaniuk (approved):
Thanks!

PR #974, opened 2023-02-13
feat: trigger remote packaging workflow

amdprophet:
Adds support to Dev & Release builds to trigger package builds in https://github.com/SumoLogic/sumologic-otel-collector-packaging.


swiatekm-sumo (commented):
Two general requests:
* Can you do the `get_productversion.sh` -> `get_version.sh` rename in git so we can see the actual diff here?
* Can we first make the change in dev, see how it works, and enable it for release in a separate PR?

amdprophet:
I can't seem to get Git to detect it as a rename which appears to be due to the size of the diff. I've separated the rename into a separate commit and the diff can be seen by viewing the changes for the first commit. Alternatively, I can open a separate PR to handle the rename.

ci/get_productversion.sh
```diff
@@ -1,77 +1,190 @@
...
+        exit 1
+    fi
 
-# Validation
-if [[ $major_version -gt 255 ]]; then
-    echo "Major version cannot be greater than 255"
-    exit 1
-fi
+    if [[ $major_version -gt 255 ]]; then
+        echo "Major version cannot be greater than 255"
```
swiatekm-sumo: Where does this requirement come from? Windows?

amdprophet: I linked to the documentation in a comment further down: https://learn.microsoft.com/en-us/windows/win32/msi/productversion.

amdprophet:
I can see if I can find a way to have it wait on the remote workflow. I should be able to present a URL in the job output for the trigger.

amdprophet:
@swiatekm-sumo we can't wait on the workflow because of a chicken and egg problem. The binary artifacts produced by the workflow in this repository aren't available until the workflow has completed. I've pushed a change to output a link to the remote workflow.

Screenshot 2023-02-15 at 12 19 07

swiatekm-sumo (approved):

PR #976, opened 2023-02-14
feat(scripts/install.sh): show link to changelog instead of full changelog

sumo-drosiek:
Make output more readable by replacing changelog with link to it

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -534,47 +534,12 @@ function ask_to_continue() {
 
 }
 
-# Get changelog for specific version
-# Only version description and breaking changes are taken
-function get_changelog() {
-    local version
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -534,47 +534,12 @@ function ask_to_continue() {
 
 }
 
-# Get changelog for specific version
-# Only version description and breaking changes are taken
+# Get changelog for specific branch
 function get_changelog() {
```
swiatekm-sumo, line 538: Maybe `show_changelog_link` now instead?

sumo-drosiek: changed to `print_changelog_link`

swiatekm-sumo (approved):

PR #977, opened 2023-02-14
feat(scripts/install.sh): print info about breaking changes

sumo-drosiek:
```
Installed version:      0.47.0-sumo-0
Getting versions...
Version to install:     0.71.0-sumo-0

The following versions contain breaking changes: v0.49.0-sumo-0, v0.51.0-sumo-0, v0.52.0-sumo-0, v0.66.0-sumo-0, v0.71.0-sumo-0! Please make sure to read the linked Changelog file.
Changelog:              https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.71.0-sumo-0/CHANGELOG.md

Downloading:            https://github.com/SumoLogic/sumologic-otel-collector/releases/download/v0.71.0-sumo-0/otelcol-sumo-0.71.0-sumo-0-linux_amd64
```

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -534,12 +534,43 @@ function ask_to_continue() {
 
 }
 
+# Print information about breaking changes
+function print_breaking_changes() {
+    local versions
+    readonly versions="${1}"
+
+    local changelog
+    readonly changelog="$(echo -e "$(curl --retry 5 --connect-timeout 5 --max-time 30 --retry-delay 0 --retry-max-time 150 -sS https://raw.githubusercontent.com/SumoLogic/sumologic-otel-collector/main/CHANGELOG.md)")"
+
+    local is_breaking_change
+    local message
+    message=""
+
+    for version in ${versions}; do
+        # Print changelog for every version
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -534,12 +534,43 @@ function ask_to_continue() {
...
+        echo "The following versions contain breaking changes: ${message}! Please make sure to read the linked Changelog file."
+    fi
+}
+
 # Print changelog link for specific branch
 function print_changelog_link() {
     local branch
     readonly branch="${1}"
 
     echo -e "Changelog:\t\thttps://github.com/SumoLogic/sumologic-otel-collector/blob/${branch}/CHANGELOG.md"
```
astencel-sumo, line 572: The only nit is that I think the link to the CHANGELOG.md should point to `main` - it might include fixes to Changelog done after release, and this is what we use when looking for breaking changes anyway.

astencel-sumo (approved):

scripts/install.sh
sumo-drosiek: It makes sense, refactored due to that change

PR #981, opened 2023-02-15
docs(sumologicextension): add example of time_zone usage and fix column names that one should refer to

perk-sumo:


sumo-drosiek (approved):
I would go with `UTC`, but LGTM

kkujawa-sumo (approved):

PR #982, opened 2023-02-15
fix: do not expose installation token

sumo-drosiek:
Do not expose installation token on stdout

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -873,7 +873,6 @@ ${indentation}sumologic:/" "${file}"
 
 # write installation token to user configuration file
 function write_installation_token() {
-    set -x
     local token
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -873,7 +873,6 @@ ${indentation}sumologic:/" "${file}"
 
 # write installation token to user configuration file
 function write_installation_token() {
-    set -x
```
astencel-sumo, line 876: Why was this even here? :thinking:

astencel-sumo (approved):

scripts/install.sh
sumo-drosiek: debug

PR #986, opened 2023-02-15
fix(syslogexporter): validate all inputs at once and add display all error messages

rnishtala-sumo:
fix(syslogexporter): validate all inputs at once and add display all error messages

### DIFF
--- a/pkg/exporter/syslogexporter/config.go
+++ b/pkg/exporter/syslogexporter/config.go
@@ -15,8 +15,10 @@
 package syslogexporter
 
 import (
-	"fmt"
+	"errors"
 	"strings"
+
+	"go.uber.org/multierr"
 
 	"github.com/THREATINT/go-net"
 
@@ -25,10 +27,10 @@ import (
 )
 
 var (
-	unsupportedPort     = "unsupported port: port is required, must be in the range 1-65535"
-	invalidEndpoint     = "invalid endpoint: endpoint is required, must be a valid FQDN or IP address"
-	unsupportedProtocol = "unsupported protocol: protocol is required, only tcp/udp supported"
-	unsupportedFormat   = "unsupported format: Only rfc5424 and rfc3164 supported"
+	unsupportedPort     = errors.New("unsupported port: port is required, must be in the range 1-65535")
+	invalidEndpoint     = errors.New("invalid endpoint: endpoint is required, must be a valid FQDN or IP address")
+	unsupportedProtocol = errors.New("unsupported protocol: protocol is required, only tcp/udp supported")
+	unsupportedFormat   = errors.New("unsupported format: Only rfc5424 and rfc3164 supported")
 )
 
 // Config defines configuration for Syslog exporter.
@@ -52,7 +54,7 @@ type Config struct {
 
 // Validate the configuration for errors. This is required by component.Config.
 func (cfg *Config) Validate() error {
-	invalidFields := []string{}
+	invalidFields := []error{}
 	if cfg.Port < 1 || cfg.Port > 65525 {
 		invalidFields = append(invalidFields, unsupportedPort)
 	}
@@ -73,7 +75,7 @@ func (cfg *Config) Validate() error {
 	}
 
 	if len(invalidFields) > 0 {
-		return fmt.Errorf("%s", strings.Join(invalidFields, "; "))
+		return multierr.Combine(invalidFields...)
 	}
 
 	return nil
--- a/pkg/exporter/syslogexporter/config.go
+++ b/pkg/exporter/syslogexporter/config.go
@@ -25,10 +25,10 @@ import (
 )
 
 var (
-	unsupportedPort     = "Unsupported port: port is required, must be in the range 1-65535"
-	invalidEndpoint     = "Invalid endpoint: endpoint is required, must be a valid FQDN or IP address"
-	unsupportedProtocol = "Unsupported protocol: protocol is required, only tcp/udp supported"
-	unsupportedFormat   = "Unsupported format: Only rfc5424 and rfc3164 supported"
+	unsupportedPort     = "unsupported port: port is required, must be in the range 1-65535"
+	invalidEndpoint     = "invalid endpoint: endpoint is required, must be a valid FQDN or IP address"
+	unsupportedProtocol = "unsupported protocol: protocol is required, only tcp/udp supported"
+	unsupportedFormat   = "unsupported format: Only rfc5424 and rfc3164 supported"
 )
 
 // Config defines configuration for Syslog exporter.
@@ -73,7 +73,7 @@ func (cfg *Config) Validate() error {
 	}
 
 	if len(invalidFields) > 0 {
-		return fmt.Errorf("%s", strings.Join(invalidFields, "\n"))
+		return fmt.Errorf("%s", strings.Join(invalidFields, "; "))
 	}
 
 	return nil
--- a/pkg/exporter/syslogexporter/config.go
+++ b/pkg/exporter/syslogexporter/config.go
@@ -15,7 +15,7 @@
 package syslogexporter
 
 import (
-	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/THREATINT/go-net"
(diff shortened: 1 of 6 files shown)


pkg/exporter/syslogexporter/config.go
```diff
@@ -24,10 +24,10 @@ import (
...
 
 var (
-	unsupportedPort     = errors.New("Unsupported port: port is required, must be in the range 1-65535")
-	invalidEndpoint     = errors.New("Invalid endpoint: endpoint is required, must be a valid FQDN or IP address")
-	unsupportedProtocol = errors.New("Unsupported protocol: protocol is required, only tcp/udp supported")
-	unsupportedFormat   = errors.New("Unsupported format: Only rfc5424 and rfc3164 supported")
+	unsupportedPort     = "Unsupported port: port is required, must be in the range 1-65535"
+	invalidEndpoint     = "Invalid endpoint: endpoint is required, must be a valid FQDN or IP address"
+	unsupportedProtocol = "Unsupported protocol: protocol is required, only tcp/udp supported"
+	unsupportedFormat   = "Unsupported format: Only rfc5424 and rfc3164 supported"
```
sumo-drosiek, line 30: By convention errors are lower case :shrug:

```suggestion
	unsupportedPort     = "unsupported port: port is required, must be in the range 1-65535"
	invalidEndpoint     = "invalid endpoint: endpoint is required, must be a valid FQDN or IP address"
	unsupportedProtocol = "unsupported protocol: protocol is required, only tcp/udp supported"
	unsupportedFormat   = "unsupported format: Only rfc5424 and rfc3164 supported"
```

pkg/exporter/syslogexporter/config.go
```diff
@@ -54,23 +54,28 @@ type Config struct {
...
 	switch cfg.Format {
 	case formatRFC3164Str:
 	case formatRFC5424Str:
 	default:
-		return unsupportedFormat
+		invalidFields = append(invalidFields, unsupportedFormat)
+	}
+
+	if len(invalidFields) > 0 {
+		return fmt.Errorf("%s", strings.Join(invalidFields, "\n"))
```
sumo-drosiek, line 78: Please use `,` as separator
```suggestion
		return fmt.Errorf("%s", strings.Join(invalidFields, ","))
```

sumo-drosiek (commented):
I don't remember if it is used in Otelcol, but there is go function to join multiple errors into one: https://pkg.go.dev/errors#Join

pkg/exporter/syslogexporter/exporter.go
```diff
@@ -121,32 +125,47 @@ func (se *syslogexporter) getTimestamp(record plog.LogRecord) time.Time {
...
-	s, err := Connect(se.logger, se.config, se.tlsConfig, se.hostname, se.pid, se.app)
-	if err != nil {
-		return fmt.Errorf("error connecting to syslog server: %s", err)
+	var errs []error
+	s, errConn := Connect(se.logger, se.config, se.tlsConfig, se.hostname, se.pid, se.app)
+	if errConn != nil {
+		errs = append(errs, errConn)
+		errStr := fmt.Sprintf("Unable to send syslog messages: %s", strings.Join(errorListToStringSlice(errs), ", "))
+		return errors.New(errStr)
 	}
```
sumo-drosiek, line 134: We should return `consumererror` which contains unsent logs. Please see the following exmaples in sumoplogicexporter:

- https://github.com/SumoLogic/sumologic-otel-collector/blob/dc1c3185755b21b2252a8f7d01a5a596e1134614/pkg/exporter/sumologicexporter/exporter.go#L180-L182
- https://github.com/SumoLogic/sumologic-otel-collector/blob/dc1c3185755b21b2252a8f7d01a5a596e1134614/pkg/exporter/sumologicexporter/exporter.go#L250-L253

kkujawa-sumo: but in syslog exporter log messages may contain token which is required by sumologic ;/

sumo-drosiek: Good point, but I don't think that returned logs are printed anywhere

rnishtala-sumo: so what should we do here. This was a deliberate choice after discussing with Kasia. This commit focuses more on error handling.

sumo-drosiek: it should return: `return consumererror.NewLogs(multierr.Combine(errs...), ld)`

where `ld` contains unsent logs

sumo-drosiek: and errs is a list of errors to print for customer

rnishtala-sumo: pushed changes to support `consumererror` with dropped logs

kkujawa-sumo: @rnishtala-sumo Could you check if returned logs are printed, see: https://github.com/SumoLogic/sumologic-otel-collector/pull/986#discussion_r1108339762?

rnishtala-sumo: it doesn't look like it
```2023-02-16T11:29:33.070-0500	error	exporterhelper/queued_retry.go:303	Exporting failed. Dropping data. Try enabling sending_queue to survive temporary failures.	{"kind": "exporter", "data_type": "logs", "name": "syslog", "dropped_items": 1}```

**Comment by rnishtala-sumo** [comment]
looks like we hit a limit in one of the tests, otherwise the PR checks seem fine
```<?xml version="1.0" encoding="utf-8"?><Error><Code>ServerBusy</Code><Message>Egress is over the account limit.```

File: pkg/exporter/syslogexporter/config.go [hunk 4]
```diff
@@ -52,23 +52,28 @@ type Config struct {
...
 	switch cfg.Format {
 	case formatRFC3164Str:
 	case formatRFC5424Str:
 	default:
-		return unsupportedFormat
+		invalidFields = append(invalidFields, unsupportedFormat)
+	}
+
+	if len(invalidFields) > 0 {
+		return fmt.Errorf("%s", strings.Join(invalidFields, "; "))
```
sumo-drosiek, line 76: There is `errors.Join` function which could do the trick, but for some reason it is not used in otelcol code at all :confused: Anyway after few days of thinking about it I think we should use it

sumo-drosiek: I know why. They are using `multierr.Combine` :D

sumo-drosiek: Please use `invalidFields := []error` and then `return multierr.Combine(invalidFields...)`

pkg/exporter/syslogexporter/exporter.go
```diff
@@ -102,32 +106,71 @@ func (se *syslogexporter) getTimestamp(record plog.LogRecord) time.Time {
...
+		}
+	}
+	if len(dropped) > 0 {
+		ld = plog.NewLogs()
+		for i := range dropped {
+			rls := ld.ResourceLogs().AppendEmpty()
+			logRecords := rls.ScopeLogs().AppendEmpty().LogRecords().AppendEmpty()
+			dropped[i].resource.MoveTo(rls.Resource())
+			for j := 0; j < len(dropped[i].records); j++ {
+				dropped[i].records[j].MoveTo(logRecords)
```
sumo-drosiek, line 135: Please add unit tests for this

rnishtala-sumo: do we need a test syslog server in our tests (mock) to write tests for the `pushLogsData`?

pkg/exporter/syslogexporter/config.go
rnishtala-sumo: I made this change

sumo-drosiek (approved):
LGTM, we can add unit tests in separate PR.

Leaving decision to @rnishtala-sumo

pkg/exporter/syslogexporter/exporter.go
rnishtala-sumo: will add the test for this in a new PR

PR #991, opened 2023-02-16
linux.yaml changes for onboarding flow. Since linux yaml in sumologic…

sumoanema:
… otel repo is out of sync with apps repo linux example yaml and app


swiatekm-sumo (approved):
:+1:

PR #1005, opened 2023-02-20
chore(syslogexporter): Add tests for the syslog exporter

rnishtala-sumo:
chore(syslogexporter): Add tests for the syslog exporter


pkg/exporter/syslogexporter/exporter_test.go
```diff
@@ -24,3 +76,92 @@ func TestInitExporter(t *testing.T) {
...
+	}
+	t.Cleanup(func() { testServer.Close() })
+	hostPort := testServer.Addr().String()
+	cfg.Endpoint, port, err = net.SplitHostPort(hostPort)
+	if err != nil {
+		t.Fatalf("could not parse port: %s", err)
+	}
+	if userPort[0].isIncorrect {
+		port = "112"
+	}
```
sumo-drosiek, line 98: Why do we need it? Why `112`?

sumo-drosiek: ok, I understand, but I don't really like passing one boolean as list of structures :thinking:

sumo-drosiek: I would rather have a two functions:
- one to get server
- one to build exporter

We can left `func prepareExporterTest(t *testing.T, cfg *Config)` as using the above functions, but for testing incorrect port I would go with the following flow (two steps in test case):

- create server
- build exporter with invalid port

---

Alternatively we can check if `cfg.Port` exists and if yes, we do not override it with server port, but it can be a little magical. We should include some comments in that case

pkg/exporter/syslogexporter/exporter_test.go
```diff
@@ -24,3 +76,92 @@ func TestInitExporter(t *testing.T) {
...
+		logs := LogRecordsToLogs(buffer)
+		err := test.exp.pushLogsData(context.Background(), logs)
+		if err != nil {
+			fmt.Printf("could not send message: %s", err)
+		}
+	}()
+	err := test.srv.SetDeadline(time.Now().Add(time.Second * 1))
+	if err != nil {
+		t.Fatalf("could not accept connection: %s", err)
+	}
```
sumo-drosiek, line 134: Please use [require.NoError](https://pkg.go.dev/github.com/stretchr/testify/require#NoError) for error checking in tests
```suggestion
	require.NoError(t, err, "cannot set deadline")
```

pkg/exporter/syslogexporter/exporter_test.go
```diff
@@ -1,14 +1,66 @@
...
+	isIncorrect bool
+}
+
+func exampleLog() plog.LogRecord {
+	buffer := plog.NewLogRecord()
+	buffer.Body().SetStr("<165>1 2003-08-24T05:14:15-07:00 192.0.2.1 myproc 8710 - - It's time to make the do-nuts.")
+	timestamp := "2003-08-24T05:14:15-07:00"
+	t, err := time.Parse(time.RFC3339, timestamp)
+	if err != nil {
+		fmt.Print(err)
```
kkujawa-sumo, line 39: please use one style in tests for capturing errors in tests, now you use either `fmt.Print` or `t.Fatal`/ `t.Logf`

pkg/exporter/syslogexporter/exporter_test.go
```diff
@@ -24,3 +76,92 @@ func TestInitExporter(t *testing.T) {
...
+	err := test.srv.SetDeadline(time.Now().Add(time.Second * 1))
+	if err != nil {
+		t.Fatalf("could not accept connection: %s", err)
+	}
+	conn, err := test.srv.AcceptTCP()
+	if err != nil {
+		t.Logf("could not accept connection: %s", err)
+		t.Log(conn)
+	}
+	assert.Contains(t, consumerErr.Error(), "dial tcp 127.0.0.1:112: connect")
```
sumo-drosiek, line 166: [assert.ErrorContains()](https://pkg.go.dev/github.com/stretchr/testify/assert#ErrorContains) would be better here

pkg/exporter/syslogexporter/exporter_test.go
swiatekm-sumo: I think with `testify`, this is idiomatically done with `require.NoError`

pkg/exporter/syslogexporter/exporter_test.go
rnishtala-sumo: Used the first method to split the workflow in two steps.

pkg/exporter/syslogexporter/exporter_test.go
```diff
@@ -1,14 +1,59 @@
...
+		}
+	}
+	return buffer
+}
+
+func LogRecordsToLogs(record plog.LogRecord) plog.Logs {
+	logs := plog.NewLogs()
+	logsSlice := logs.ResourceLogs().AppendEmpty().ScopeLogs().AppendEmpty().LogRecords()
+	tgt := logsSlice.AppendEmpty()
+	record.CopyTo(tgt)
```
sumo-drosiek, line 53: What `tgt` stands for?
I would use `ls`

pkg/exporter/syslogexporter/exporter_test.go
```diff
@@ -24,3 +69,104 @@ func TestInitExporter(t *testing.T) {
...
+	if invalidExporter {
+		exp, err := buildInvalidExporter(t, testServer, cfg)
+		require.NoError(t, err, "Error building exporter")
+		require.NotNil(t, exp)
+		return &exporterTest{
+			srv: testServer,
+			exp: exp,
+		}
+	}
+	exp, err := buildValidExporter(t, testServer, cfg)
```
sumo-drosiek, line 122:
```suggestion
        var exp syslogexporter
        var err error
	if invalidExporter {
		exp, err := buildInvalidExporter(t, testServer, cfg)
	} else {
		exp, err := buildValidExporter(t, testServer, cfg)
	}
```

pkg/exporter/syslogexporter/exporter_test.go
```diff
@@ -24,3 +69,104 @@ func TestInitExporter(t *testing.T) {
...
+	buffer := exampleLog(t)
+	logs := LogRecordsToLogs(buffer)
+	consumerErr := test.exp.pushLogsData(context.Background(), logs)
+	err := test.srv.SetDeadline(time.Now().Add(time.Second * 1))
+	require.NoError(t, err, "cannot set deadline")
+	conn, err := test.srv.AcceptTCP()
+	require.ErrorContains(t, err, "i/o timeout")
+	require.Nil(t, conn)
+	assert.ErrorContains(t, consumerErr, "dial tcp 127.0.0.1:112: connect")
+}
```
sumo-drosiek, line 172: We should check if `consumerErr` contains unsent logs

pkg/exporter/syslogexporter/exporter_test.go
rnishtala-sumo: I tried this approach, but looks like exp is `nil` because of different scopes (inside and outside the if clause)

pkg/exporter/syslogexporter/exporter_test.go
rnishtala-sumo: it looks like one cannot return failed logs from the documentation `NewLogs` just returns an `error`

pkg/exporter/syslogexporter/exporter_test.go
rnishtala-sumo: this change has been made, I think we need `=` instead of `:=`

sumo-drosiek (approved):

PR #1006, opened 2023-02-21
fix(scripts/install): ignore errors from semanage installation

sumo-drosiek:


swiatekm-sumo:
Shouldn't we just error in this case, or do something else? We need the `semanage` to set the right SELinux ACLs.

sumo-drosiek:
We print warning in below section:
https://github.com/SumoLogic/sumologic-otel-collector/blob/b00946fbd89394269b7366a6ef27b1ccdf8eae2e/scripts/install.sh#L1332-L1334

swiatekm-sumo (approved):

sumo-drosiek:
@portertech we effectively do the same like for missing `setfacl` command. We print information that some features won;t work, but we do not fail

PR #1017, opened 2023-02-23
feat(sumologicexporter): use otlp url suffixes

swiatekm-sumo:
If the format is otlp, sumologic expoter should behave the same way as otlphttp exporter when it comes to
adding the per-signal url suffixes.

### DIFF
--- a/pkg/exporter/sumologicexporter/exporter.go
+++ b/pkg/exporter/sumologicexporter/exporter.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"path"
+	"strings"
 	"sync"
 
 	"go.opentelemetry.io/collector/component"
@@ -419,7 +421,19 @@ func (se *sumologicexporter) configure(ctx context.Context) error {
 		se.setDataURLs(logsUrl.String(), metricsUrl.String(), tracesUrl.String())
 
 	} else if httpSettings.Endpoint != "" {
-		se.setDataURLs(httpSettings.Endpoint, httpSettings.Endpoint, httpSettings.Endpoint)
+		logsUrl, err := getSignalURL(se.config, httpSettings.Endpoint, component.DataTypeLogs)
+		if err != nil {
+			return err
+		}
+		metricsUrl, err := getSignalURL(se.config, httpSettings.Endpoint, component.DataTypeMetrics)
+		if err != nil {
+			return err
+		}
+		tracesUrl, err := getSignalURL(se.config, httpSettings.Endpoint, component.DataTypeTraces)
+		if err != nil {
+			return err
+		}
+		se.setDataURLs(logsUrl, metricsUrl, tracesUrl)
 
 		// Clean authenticator if set to sumologic.
 		// Setting to null in configuration doesn't work, so we have to force it that way.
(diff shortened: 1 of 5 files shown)


pkg/exporter/sumologicexporter/exporter_test.go
```diff
@@ -796,3 +808,72 @@ func TestSendEmptyTraces(t *testing.T) {
...
+			description: "no change if log format not otlp",
+			signalType:  component.DataTypeLogs,
+			cfg:         &Config{LogFormat: TextFormat},
+			endpointUrl: "http://localhost",
+			expected:    "http://localhost",
+		},
+		{
+			description: "no change if metric format not otlp",
+			signalType:  component.DataTypeMetrics,
+			cfg:         &Config{MetricFormat: PrometheusFormat},
```
rnishtala-sumo, line 831: should we have one test for otlp format, if I understand that would change the Endpoint to
`http://localhost/v1/metrics`

pkg/exporter/sumologicexporter/sender.go
```diff
@@ -324,19 +324,16 @@ func (s *sender) handleReceiverResponse(resp *http.Response) error {
 
 func (s *sender) createRequest(ctx context.Context, pipeline PipelineType, data io.Reader) (*http.Request, error) {
 	var url string
-	if s.config.HTTPClientSettings.Endpoint == "" {
```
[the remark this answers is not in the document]
swiatekm-sumo: It's the reverse. Even if the endpoint is not empty, we still use the signal-specific urls. This actually doesn't change anything, because these values were the same; that is, if endpoint was set, the signal-specific urls were equal to it.

pkg/exporter/sumologicexporter/exporter_test.go
swiatekm-sumo: I could add similar tests for logs and metrics. I originally thought they were redundant, but maybe that's fine if it makes the behaviour clearer.

pkg/exporter/sumologicexporter/exporter.go
```diff
@@ -470,3 +485,33 @@ func (se *sumologicexporter) shutdown(context.Context) error {
 func (se *sumologicexporter) dropRoutingAttribute(attr pcommon.Map) {
 	attr.Remove(se.config.DropRoutingAttribute)
 }
+
+// get the destination url for a given signal type
+// this mostly adds signal-specific suffixes if the format is otlp
+func getSignalURL(oCfg *Config, endpointUrl string, signal component.DataType) (string, error) {
+	signalUrlSuffix := fmt.Sprintf("/v1/%s", signal)
```
sumo-drosiek, line 492: nit: I would move it just before usage

pkg/exporter/sumologicexporter/exporter_test.go
```diff
@@ -796,3 +809,72 @@ func TestSendEmptyTraces(t *testing.T) {
...
+			signalType:  component.DataTypeTraces,
+			endpointUrl: "http://localhost/v1/traces",
+			expected:    "http://localhost/v1/traces",
+		},
+		{
+			description: "no change if suffix already present",
+			signalType:  component.DataTypeTraces,
+			endpointUrl: "http://localhost/v1/traces",
+			expected:    "http://localhost/v1/traces",
+		},
```
sumo-drosiek, line 853:
```suggestion
		{
			description: "no change if suffix already present",
			signalType:  component.DataTypeTraces,
			endpointUrl: "http://localhost/v1/traces",
			expected:    "http://localhost/v1/traces",
		},
```

sumo-drosiek: Duplication

sumo-drosiek:
Does it work correctly with extension? Asking just in case. Do not remember how the injection actually works?

swiatekm-sumo:
> Does it work correctly with extension? Asking just in case. Do not remember how the injection actually works?

The code getting urls from the extension is in a different conditional branch, and we don't touch it. This is just for an `endpoint` set in the config.

rnishtala-sumo (approved):

sumo-drosiek (approved):

PR #1020, opened 2023-02-24
fix(sumologicexporter): fix links to upgrade docs

astencel-sumo:
We were trying to be helpful by helping the user migrate off unsupported properties, but links that point to nowhere do not help much.
This changes the links to point straight to the upgrade docs instead of release notes.
I also point to the `main` branch of the upgrade docs, because there have been, and there might be more, changes/improvements to these docs.

### DIFF
--- a/pkg/exporter/sumologicexporter/config.go
+++ b/pkg/exporter/sumologicexporter/config.go
@@ -114,26 +114,26 @@ func CreateDefaultHTTPClientSettings() confighttp.HTTPClientSettings {
 func (cfg *Config) Validate() error {
 
 	if len(cfg.MetadataAttributes) > 0 {
-		return fmt.Errorf(`metadata_attributes is not supported anymore.
-Please consult the changelog at https://github.com/SumoLogic/sumologic-otel-collector/releases/tag/v0.49.0-sumo-0`,
+		return fmt.Errorf(`the property metadata_attributes was removed in v0.49.0-sumo-0.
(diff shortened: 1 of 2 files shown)


pkg/exporter/sumologicexporter/config.go
```diff
@@ -114,26 +114,26 @@ func CreateDefaultHTTPClientSettings() confighttp.HTTPClientSettings {
 func (cfg *Config) Validate() error {
 
 	if len(cfg.MetadataAttributes) > 0 {
-		return fmt.Errorf(`metadata_attributes is not supported anymore.
-Please consult the changelog at https://github.com/SumoLogic/sumologic-otel-collector/releases/tag/v0.49.0-sumo-0`,
+		return fmt.Errorf(`the property metadata_attributes was removed in v0.49.0-sumo-0.
+See upgrade guide at https://github.com/SumoLogic/sumologic-otel-collector/blob/main/docs/upgrading.md#sumo-logic-exporter-metadata-handling`,
 		)
 	}
```
sumo-drosiek, line 120: Can't we just remove it? It was `23` version back

sumo-drosiek (approved):

pkg/exporter/sumologicexporter/config.go
astencel-sumo: Yeah we probably should get rid of these warnings at one time or another :smile:

PR #1023, opened 2023-02-27
fix(install-script): clears curl cache before attempting download to prevent installation errors

rnishtala-sumo:
fix(install-script): clears curl cache before attempting download to prevent installation errors


sumo-drosiek (changes requested):
Not sure is it the correct solution. As this is just like discarding cache. In my opinion we should check if the cached file is correct (eg. correct binary, or trying to run it (but it sounds like security problem, to run unverified file, ideally we could check correctness of checksum, but we do not provide them right now)

sumo-drosiek:
As per internal conversation:

Apparently issue occurs only if someone is using cache, and it will be resolved by using checksums to verify if cached file is correct

PR #1026, opened 2023-02-28
feat: add scripts/install.ps1

amdprophet:
Implements an install script in PowerShell for Windows that can detect OS type, CPU architecture, installed otelcol-sumo version, latest otelcol-sumo version, and can accept the following params:
* `-Version <version>`
* `-InstallationToken <installation_token>` (defaults to the `SUMOLOGIC_INSTALLATION_TOKEN` envvar)
* `-Tags <tags_as_hashtable>` (e.g. `-Tags @{ tag1 = "foo"; tag2 = "bar" }`)

The following command can be used to download & install after replacing `replaceme` with your installation token:

```powershell
Set-ExecutionPolicy RemoteSigned -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; $uri = "https://raw.githubusercontent.com/SumoLogic/sumologic-otel-collector/feat/install-ps1/scripts/install.ps1"; $path="${env:TEMP}\install.ps1"; (New-Object System.Net.WebClient).DownloadFile($uri, $path); . $path -InstallationToken replaceme -Tags @{tag3 = "foo"; tag4 = "bar"}
```

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -1,0 +1,463 @@
+using assembly System.Net.Http
+using namespace System.Net.Http
+
+param (
+    # Version is used to override the version of otelcol-sumo to install.
+    [string] $Version,
+
+    # InstallationToken is used to pass a Sumo Logic installation token to
+    # this script. The default value is set to the value of the 
+    # SUMOLOGIC_INSTALLATION_TOKEN environment variable.
+    [string] $InstallationToken = $env:SUMOLOGIC_INSTALLATION_TOKEN,
+
+    # Tags is used to specify a list of tags for the collector. Specified via a
+    # hash table.
+    # e.g. -Tags @{ tag1 = "foo" ; tag2 = "bar" }
+    [Hashtable] $Tags
+)
+
+##
+# Security tweaks
+#
+# This script requires TLS v1.2 or newer. Due to some versions of Windows not
+# using TLS v1.2 or TLS v1.3 by default we must detect if it is enabled and
+# attempt to enable it if not:
+#
+# 1. Determine if enabled security protocols contain an allowed security
+#    protocol. If yes then do nothing.
+# 2. Find which security protocols from the list of allowed protocols are
+#    supported by the system. If none, return an error.
+# 3. Enable the found security protocols.
+##
+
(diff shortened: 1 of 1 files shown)


scripts/install.ps1
```diff
@@ -0,0 +1,463 @@
...
+
+    Write-Host "Getting installed version..."
+    $installedVersion = Get-InstalledVersion
+    $installedVersionStr = "none"
+    if ($installedVersion -ne $null) {
+        $installedVersionStr = $installedVersion
+    }
+    Write-Host "Installed version:`t${installedVersionStr}"
+
+    # Get versions, but ignore errors are we fallback to other methods later
```
sumo-drosiek, line 397:
```suggestion
    # Get versions, but ignore errors as we fallback to other methods later
```

scripts/install.ps1
```diff
@@ -0,0 +1,463 @@
...
+    $msiLanguage = "en-US"
+    $msiFileName = "otelcol-sumo_${productVersion}_${msiLanguage}.${archName}.msi"
+    $msiUri = "https://github.com/SumoLogic/sumologic-otel-collector/releases/download/"
+    $msiUri += "v${Version}/${msiFileName}"
+    $msiPath = "${env:TEMP}\${msiFileName}"
+    Get-BinaryFromUri $msiUri -Path $msiPath -HttpClient $httpClient
+
+    # Install MSI
+    [string[]] $msiProperties = @()
+    $msiProperties += "INSTALLATIONTOKEN=${InstallationToken}"
```
sumo-drosiek, line 451: Does it work for `v0.72.0` release? As I changed the name recently I would expect it to work from `v0.73.0` :thinking:

amdprophet: Yes, it worked when I was testing it and it looks like it was included in `v0.72.0` https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.72.0-sumo-0/packaging/msi/wix/package.wxs

sumo-drosiek: eh, seems that I unnecessary added support for `INSTALLTOKEN` again :roll_eyes:

PR #1028, opened 2023-03-01
docs(migration): update information about local file source

sumo-drosiek:


docs/migration.md
```diff
@@ -418,11 +404,86 @@ exporters:
...
+      ## cleanup timestamp_field
+      - type: remove
+        field: attributes.timestamp_field
+    ## Marks to remove logs older than Sat Dec 31 2022 23:00:00 GMT+0000
+    transform/mark older:
+      log_statements:
+      - context: log
+        statements:
+        # 1672527600000000000 ns is equal to Sat Dec 31 2022 23:00:00 GMT+0000
+        - set(attributes["to_remove"], "true") where time_unix_nano < 1672527600000000000
```
rnishtala-sumo, line 460: is this going to add a new attribute to the log record attributes map?
https://pkg.go.dev/go.opentelemetry.io/collector/pdata/plog#LogRecord.Attributes

docs/migration.md
```diff
@@ -485,34 +546,51 @@ exporters:
...
+  extensions:
+    sumologic:
+      collector_fields:
+        cloud.availability_zone: zone-1
+        k8s.cluster.name: my-cluster
+  ```
+
+- For fields, which should be set for specific data, you should use [Transform Processor][transformprocessor].
+
+  For example, the following snippet configures two fields, `cloud.availability_zone` and `k8s.cluster.name`:
```
rnishtala-sumo, line 566: I think the example below isn't setting those fields. Also, does Transform processor add attributes to the log record?

docs/migration.md
```diff
@@ -418,11 +404,86 @@ exporters:
...
-The OpenTelemetry Collector doesn't have a substitute for this Installed Collector option.
-It supports two options, starting at the beginning or end of a file.
-Starting at the `beginning` will read the entire file every time it's started.
-Starting at the `end` will read only logs appended to file after it's started.
-This is configurable with the `start_at` option.
+The OpenTelemetry Collector substitution for this Installed Collector option is complex,
+as the following steps needs to be performed:
+
+- reading logs using [Filelog Receiver][filelogreceiver]
+- parsing timestamp from the logs using [Logs Transform Processor][logstransformprocessor]
```
astencel-sumo, line 411: Isn't it possible to do all this in one step using the Filter processor with [OTTL](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/filterprocessor#ottl)? The specific example of only logs from 2023 would be something like:

```yaml
processors:
  filter:
    logs:
      log_record:
      - 'Int(Substring(body, 0, 4)) >= 2023'
```

Even if we want the condition to be more complex, I believe we should be able to do away without adding additional attributes and removing in a separate step?

docs/migration.md
```diff
@@ -418,11 +404,86 @@ exporters:
...
+      - tmp/logs.log
+      # - ...
+      ## Adds file path log.file.path attribute, which will be used further in pipeline
+      include_file_path: true
+      ## We would like to read from beginning, as we can choose only between end and beginning
+      start_at: beginning
+      # ...
+  processors:
+    ## The following processor parses out the timestamp from `tmp/logs.log`
+    logstransform/extract timestamp:
```
swiatekm-sumo, line 441: Wouldn't this parsing be easier to do in filelog receiver itself?

docs/migration.md
```diff
@@ -418,11 +404,86 @@ exporters:
...
+      log_statements:
+      - context: log
+        statements:
+        # 1672527600000000000 ns is equal to Sat Dec 31 2022 23:00:00 GMT+0000
+        - set(attributes["to_remove"], "true") where time_unix_nano < 1672527600000000000
+    ## Remove logs which has been marked to be removed
+    filter/remove older:
+      logs:
+        exclude:
+          match_type: strict
```
swiatekm-sumo, line 465: You can use ottl directly in filterprocessor now: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/filterprocessor/README.md#ottl

docs/migration.md
sumo-drosiek: damn, I was scrooling README and stopped on `expr` :facepalm:. We definitely should use OTTL in filterprocessor if it support its. I used simple example, but can easily imagine equivalent:
`let's take logs from 2022-12-04 13:13:15`

docs/migration.md
```diff
@@ -210,7 +210,8 @@ exporters:
 
 ### Fields
 
-Fields in the Opentelemetry Collector can be added with the [resourceprocessor][resourceprocessor].
+Fields in the Opentelemetry Collector can be added with the `collector_fields` property of [Sumo Logic Extension][sumologicextension].
```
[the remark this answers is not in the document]
sumo-drosiek: Yes, this is equivalent of Installed Collector Collector configuration

docs/migration.md
sumo-drosiek: It was more logical for me to put it separetely when I wasn't aware that filterprocessor can into OTTL

docs/migration.md
sumo-drosiek: refactored

docs/migration.md
sumo-drosiek: fixed that with changing `context` to `resource`

docs/migration.md
```diff
@@ -418,11 +404,69 @@ exporters:
 
 ##### Collection should begin
 
-The OpenTelemetry Collector doesn't have a substitute for this Installed Collector option.
-It supports two options, starting at the beginning or end of a file.
-Starting at the `beginning` will read the entire file every time it's started.
-Starting at the `end` will read only logs appended to file after it's started.
-This is configurable with the `start_at` option.
+The OpenTelemetry Collector substitution for this Installed Collector option require manual timestamp parsing.
+Then you can use [Filter Processor][filterprocessor] to filter out logs before specific date.
```
swiatekm-sumo, line 408:
```suggestion
The OpenTelemetry Collector substitution for this Installed Collector option requires manual timestamp parsing.
Then you can use [Filter Processor][filterprocessor] to filter out logs before a specific date.
```

swiatekm-sumo (approved):

PR #1033, opened 2023-03-01
docs(syslog): migration doc edits for syslog source

rnishtala-sumo:
migration doc edits for syslog source

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -864,11 +864,14 @@ Remote File Source is not supported by the OpenTelemetry Collector.
 ### Syslog Source
 
 The equivalent of the Syslog Source is a combination of
-[the tcplog][tcplogreceiver] or [the udplog][udplogreceiver] receivers
-and [the sumologicsyslog processor][sumologicsyslog].
+[the TCP][tcplogreceiver] or [the UDP][udplogreceiver] receivers
+and [the Sumo logic Syslog Processor][sumologicsyslog].
 
-__Note: The OpenTelemetry Collector provides also [Syslog Receiver][syslogreceiver].
+__Note: The OpenTelemetry Collector also provides the [Syslog Receiver][syslogreceiver].
 See [this document](comparison.md#syslog) for details.__
+
+__The syslog messages could also be sent to sumologic using the [syslog exporter][syslogexporter] with the
+[syslog parser][syslogparser]__
 
 #### Overall example
 
@@ -909,18 +912,26 @@ processors:
   ## There is no substitute for `Description` in current project phase.
   ## It is recommended to use comments for that purpose, like this one.
   ## sumologic_syslog/<source group name>:
-  ## <source group name> can be substitute of Installed Collector `Name`.
+  ## <source group name> can be substitute of Installed Collector `Name`
   sumologic_syslog/syslog source:
-
-  ## The following configuration will add two fields to every record
-  resource/syslog source:
-    attributes:
-    - key: cloud.availability_zone
-      value: zone-1
-      action: insert
-    - key: k8s.cluster.name
-      value: my-cluster
-      action: insert
+  ## re-associates record attributes to a resource to make them available to be used by Source Processor
+  groupbyattrs:
+    keys:
+      - net.peer.name
+      - facility
+  ## Leave the timestamp parsing to the Sumo Logic backend
+  transform/clear_logs_timestamp:
+    log_statements:
+      - context: log
+        statements:
+          - set(time_unix_nano, 0)
+  ## The following configuration will add two fields to every resource
+  transform/syslog source:
+    log_statements:
+      - context: resource
+        statements:
+          - set(attributes["cloud.availability_zone"], "zone-1")
+          - set(attributes["k8s.cluster.name"], "my-cluster")
   source/syslog source:
     ## Installed Collector substitute for `Source Category`.
     source_category: example category
@@ -928,25 +939,23 @@ processors:
     source_name: "%{facility}"
     ## Set Source Host to `net.peer.name`
     source_host: "%{net.peer.name}"
+
 exporters:
   sumologic/syslog:
-    ## clear_logs_timestamp is by default set to True.
-    ## If it's set to true, it works like `Enable Timestamp Parsing`,
-    ## and `Time Zone` is going to be taken from `extensions` section.
-    ## There is no possibility to configure several time zones in one exporter.
-    ## clear_logs_timestamp sets to true also behaves like
-    ## `Timestamp Format` would be set to `Automatically detect the format`
-    ## in terms of Installed Collector configuration.
-    clear_logs_timestamp: true
+
 service:
   extensions:
   - sumologic
   pipelines:
     logs/syslog source:
       receivers:
-      - filelog/syslog source
+      - tcplog/first receiver
+      - udplog/first receiver
       processors:
-      - resource/syslog source
+      - transform/clear_logs_timestamp
+      - sumologic_syslog/syslog source
+      - groupbyattrs
+      - transform/syslog source
       - source/syslog source
       exporters:
       - sumologic/syslog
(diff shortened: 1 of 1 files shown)


docs/migration.md
```diff
@@ -842,33 +843,17 @@ extensions:
...
-    ## rel: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.72.0/receiver/udplogreceiver#configuration
-    add_attributes: true
+  ## Use a syslog receiver
+  syslog:
+    tcp:
+      listen_address: "0.0.0.0:54526"
+    protocol: rfc5424
+    operators:
+      - type: syslog_parser
+        protocol: rfc5424
```
sumo-drosiek, line 853: We have been using tcplog, because it doesn't verify correctness of syslog format. Like Installed Collector. Installed Collector's syslog is more like tcp/udp log receiver that syslogreceiver

sumo-drosiek (commented):
I would migrate to tcp/udp log receiver, as it won't raise an errors for non-rfc data. Like Installed Collecto

docs/migration.md
sumo-drosiek: The issue here is that we want to behave as close as possible to Installed Collector, so I wouldn't try to follow the rfc in this document. We don't want to change behavior for customers.

To be honest, Syslog source in Installed Collector should be named TCP Source, but it is what it is :shrug:

sumo-drosiek: Also we are using sumologic exporter for all of data in this document. If we use `syslog_parser` the customers will get different data that now, as now they get raw syslog lines

rnishtala-sumo: so I tested this and it looks like the pipeline will not work without the `syslog_parser` operator at the moment

sumo-drosiek: btw, we should fix syslog exporter to no send
`%!d(<nil>) 1970-01-01T00:00:00Z %!s(<nil>) %!s(<nil>) %!s(<nil>) %!s(<nil>) - %!s(<nil>)`

docs/migration.md
```diff
@@ -901,54 +898,47 @@ processors:
...
+    # NOTE: this limit does not apply to data sent in otlp format,
+    #   to limit size of otlp requests, please use the batch processor:
+    #   https://github.com/open-telemetry/opentelemetry-collector/tree/v0.71.0/processor/batchprocessor
+    # default = 1_048_576 (1MB)
+    max_request_body_size: <max_request_body_size>
+
+    # format to use when sending logs to Sumo Logic, default = otlp,
+    # NOTE: only `otlp` is supported when used with sumologicextension
+    log_format: {json, text, otlp}
+
```
sumo-drosiek, line 920: I would avoid adding too much configuration options for customers, epecially for migration from installed collector, as they should use defaults and shouldn't play with log format, request size, compression or endpoint by themselves

```suggestion
  sumologic:

```

docs/migration.md
```diff
@@ -879,13 +882,7 @@ receivers:
...
 
 processors:
-  ## There is no substitute for `Description` in current project phase.
-  ## It is recommended to use comments for that purpose, like this one.
-  ## sumologic_syslog/<source group name>:
-  ## <source group name> can be substitute of Installed Collector `Name`.
-  sumologic_syslog/syslog source:
-
-  ## The following configuration will add two fields to every record
+  ## The following configuration will add two fields to every resource
```
sumo-drosiek, line 885: I would leave this processor, as it ensures backward compatibility with syslog source
https://github.com/SumoLogic/sumologic-otel-collector/tree/main/pkg/processor/sumologicsyslogprocessor

Please ensure that final configuration of OTC produces the same result as Installed Collector

rnishtala-sumo: I tested this with the `syslog exporter`, I can double check it with the `sumologic exporter` again

rnishtala-sumo: The OT collector registered successfully using this endpoint on long: https://long-open-events.sumologic.net

I saw the log message here:
https://stagdata.long.sumologic.net/ui/#/search/create?id=1ML0NOm6P7QbVhNPtWxBeunSJUhINTnfEapIjTWi

The only abnormality I noticed was that the `source name` and `source host` were `undefined`

sumo-drosiek: We need to fix it. `facility` is probably due to missing `sumologicsyslogprocessor`

sumo-drosiek: `net.peer.name` is probably due to `tcp`/`udp` configuration

docs/migration.md
```diff
@@ -851,6 +854,7 @@ Below is an example of an OpenTelemetry configuration for a Syslog Source.
 extensions:
   sumologic:
     installation_token: <installation_token>
+    api_base_url: <api_base_url>
```
sumo-drosiek, line 857: We don't need it for customers. Default `api_base_url` works fine for all production deployments

docs/migration.md
rnishtala-sumo: used `'groupbyattr` to fix this.

docs/migration.md
```diff
@@ -861,10 +864,14 @@ processors:
   ## There is no substitute for `Description` in current project phase.
   ## It is recommended to use comments for that purpose, like this one.
   ## sumologic_syslog/<source group name>:
-  ## <source group name> can be substitute of Installed Collector `Name`.
+  ## <source group name> can be substitute of Installed Collector `Name`
   sumologic_syslog/syslog source:
-
-  ## The following configuration will add two fields to every record
+  ## re-associates record attributes to a resource to make them available to be used source attributes
```
sumo-drosiek, line 869:
```suggestion
  ## re-associates record attributes to a resource to make them available to be used by Source Processor
```

docs/migration.md
```diff
@@ -880,24 +887,20 @@ processors:
...
   sumologic/syslog:
-    ## clear_logs_timestamp is by default set to True.
-    ## If it's set to true, it works like `Enable Timestamp Parsing`,
-    ## and `Time Zone` is going to be taken from `extensions` section.
-    ## There is no possibility to configure several time zones in one exporter.
-    ## clear_logs_timestamp sets to true also behaves like
-    ## `Timestamp Format` would be set to `Automatically detect the format`
-    ## in terms of Installed Collector configuration.
-    clear_logs_timestamp: true
+
```
sumo-drosiek, line 893: We should use transform processor in order to set timestamp to zero

docs/migration.md
```diff
@@ -2125,12 +2112,7 @@ Remote File Source is not supported by the OpenTelemetry Collector.
 ### Syslog Source (Syslog)
 
 The equivalent of the Syslog Source is a combination of
-[the tcplog][tcplogreceiver] or [the udplog][udplogreceiver] receivers
-and [the sumologicsyslog processor][sumologicsyslog].
-More useful information can be found in [Syslog Source for Cloud Based Management](#syslog-source).
-
-__Note: The OpenTelemetry Collector provides also [Syslog Receiver][syslogreceiver].
-See [this document](comparison.md#syslog) for details.__
+[the tcplog][tcplogreceiver] or [the udplog][udplogreceiver] receivers and [the sumologicsyslog processor][sumologicsyslog].
```
sumo-drosiek, line 2115:
```suggestion
[the Tcp][tcplogreceiver] or [the UDP][udplogreceiver] receivers and [the 
Sumo Logic Syslog Processor][sumologicsyslog].
```

sumo-drosiek (approved):
Left minor comments

docs/migration.md
```diff
@@ -861,10 +864,14 @@ processors:
...
+  ## <source group name> can be substitute of Installed Collector `Name`
   sumologic_syslog/syslog source:
-
-  ## The following configuration will add two fields to every record
+  ## re-associates record attributes to a resource to make them available to be used source attributes
+  groupbyattrs:
+    keys:
+      - net.peer.name
+      - facility
+  ## The following configuration will add two fields to every resource
```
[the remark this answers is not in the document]
rnishtala-sumo: how do we add fields/attributes in ottl?

docs/migration.md
```diff
@@ -909,73 +912,61 @@ processors:
...
-    clear_logs_timestamp: true
+
 service:
   extensions:
   - sumologic
   pipelines:
     logs/syslog source:
       receivers:
-      - filelog/syslog source
+      - tcplog/first receiver
```
sumo-drosiek, line 952:
```suggestion
      - tcplog/first receiver
      - udplog/first receiver
```

lets use both of them if we defined two

sumo-drosiek (approved):

PR #1036, opened 2023-03-02
feat(scripts/install.sh): expose timeout configuration

sumo-drosiek:
```
Supported arguments:
  -i, --installation-token <token>      Installation token. It has precedence over 'SUMOLOGIC_INSTALLATION_TOKEN' env variable.
  -k, --skip-installation-token              Skips requirement for installation token.
                                        This option do not disable default configuration creation.
  -t, --tag <key=value>                 Sets tag for collector. This argument can be use multiple times. One per tag.
  -w, --download-only                   Download new binary only and skip configuration part.

  -u, --uninstall                       Removes Sumo Logic Distribution for OpenTelemetry Collector from the system and
                                        disable Systemd service eventually.
                                        Use with '--purge' to remove all configurations as well.
  -p, --purge                           It has to be used with '--uninstall'.
                                        It removes all Sumo Logic Distribution for OpenTelemetry Collector related configuration and data.

  -a, --api <url>                       Api URL
  -d, --skip-systemd                    Do not install systemd unit.
  -s, --skip-config                     Do not create default configuration.
  -v, --version <version>               Version of Sumo Logic Distribution for OpenTelemetry Collector to install, e.g. 0.57.2-sumo-1.
                                        By default it gets latest version.
  -f, --fips                            Install the FIPS 140-2 compliant binary on Linux.
  -H, --install-hostmetrics             Install the hostmetrics configuration to collect host metrics.
  -m, --download-timeout <timeout>      Timeout in seconds after which download will fail. Default is 180.
  -y, --yes                             Disable confirmation asks.

  -h, --help                            Prints this help and usage.

Supported env variables:
  SUMOLOGIC_INSTALLATION_TOKEN=<token>       Installation token.'
```

### DIFF
--- a/pkg/scripts_test/common.go
+++ b/pkg/scripts_test/common.go
@@ -4,6 +4,7 @@ package sumologic_scripts_tests
 
 import (
 	"fmt"
+	"os"
 	"runtime"
 	"testing"
 
@@ -38,6 +39,11 @@ func tearDown(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func cleanCache(t *testing.T) {
+	err := os.RemoveAll(cacheDirectory)
+	require.NoError(t, err)
+}
+
 func runTest(t *testing.T, spec *testSpec) {
 	ch := check{
 		test:                t,
@@ -62,6 +68,12 @@ func runTest(t *testing.T, spec *testSpec) {
 	}
 
 	ch.code, ch.output, ch.errorOutput, ch.err = runScript(ch)
+
+	// Remove cache in case of curl issue
+	if ch.code == 28 {
+		cleanCache(t)
+	}
+
 	checkRun(ch)
 
 	for _, c := range commonPostChecks {
--- a/pkg/scripts_test/common.go
+++ b/pkg/scripts_test/common.go
@@ -70,7 +70,7 @@ func runTest(t *testing.T, spec *testSpec) {
 	ch.code, ch.output, ch.errorOutput, ch.err = runScript(ch)
 
 	// Remove cache in case of curl issue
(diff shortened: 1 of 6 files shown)


scripts/install.sh
```diff
@@ -140,6 +142,7 @@ Supported arguments:
                                         By default it gets latest version.
   -${ARG_SHORT_FIPS}, --${ARG_LONG_FIPS}                            Install the FIPS 140-2 compliant binary on Linux.
   -${ARG_SHORT_INSTALL_HOSTMETRICS}, --${ARG_LONG_INSTALL_HOSTMETRICS}             Install the hostmetrics configuration to collect host metrics.
+  -${ARG_SHORT_TIMEOUT}, --${ARG_LONG_TIMEOUT} <timeout>      Timeout in seconds after which download will fail. Default is ${CURL_MAX_TIME}.
```
rnishtala-sumo, line 145: so is this an option until we have the asset checksum verification?

pkg/scripts_test/common.go
```diff
@@ -62,6 +68,12 @@ func runTest(t *testing.T, spec *testSpec) {
 	}
 
 	ch.code, ch.output, ch.errorOutput, ch.err = runScript(ch)
+
+	// Remove cache in case of curl issue
+	if ch.code == 28 {
```
rnishtala-sumo, line 73: should we define a constant for this code?

scripts/install.sh
sumo-drosiek: No, this is required for slow connection and is not related to cache issue

rnishtala-sumo (approved):

PR #1040, opened 2023-03-03
fix(windows): do not expose healthcheck on 0.0.0.0 in default configu…

sumo-drosiek:
…ration


aiwaniuk:
do we need to include in windows this as well https://github.com/SumoLogic/sumologic-otel-collector/pull/966/files?

swiatekm-sumo (approved):

PR #1053, opened 2023-03-09
docs(migration): extract source category and fields to common section

sumo-drosiek:


aboguszewski-sumo (approved):
Nice. We should do the same with source host.

PR #1055, opened 2023-03-10
feat(sumologicschemaprocessor): add docker stats metrics mapping

aboguszewski-sumo:


sumo-drosiek:
Could we add mapping to sumologicschema as part of this PR, or write some config to do this for customers? Should be pretty easy task and customers won't have to migrate their dashboards right now

aboguszewski-sumo:
> Could we add mapping to sumologicschema as part of this PR

We could, but is it worth it? I mean, for other sources there would be probably a similar situation, so do we want to add all that there?

sumo-drosiek:
> We could, but is it worth it? I mean, for other sources there would be probably a similar situation, so do we want to add all that there?

We should ensure that switching from Installed Collector to OTC will be as seamless as possible

aboguszewski-sumo:
@sumo-drosiek I added the subprocessor, please review.

pkg/processor/sumologicschemaprocessor/config.go
```diff
@@ -36,6 +37,7 @@ const (
 	defaultAddCloudNamespace           = true
 	defaultTranslateAttributes         = true
 	defaultTranslateTelegrafAttributes = true
+	defaultTranlateDockerMetrics       = true
```
aboguszewski-sumo, line 40: This causes this change to be a breaking one, how careful are we going to be with that?

sumo-drosiek: Lets make it optional and discuss next week

sumo-drosiek: IMHO, it should be enable by default and it is rather lower severity

sumo-drosiek (approved):
Awesome!!!

pkg/processor/sumologicschemaprocessor/config.go
aboguszewski-sumo: I set it to "false". I'll add this topic to Monday meeting's agenda

PR #1057, opened 2023-03-10
fix(sumologicextension)!: use fqdn before os.Hostname

swiatekm-sumo:
Use FQDN before hostname for the default collector name and metadata. The value is taken from the `go-fqdn` package, which is also what resource detection processor does.


perk-sumo (approved):
LGTM, thank you!

sumo-drosiek (approved):

PR #1081, opened 2023-03-20
add migrating dockerstats metadata

aboguszewski-sumo:


docs/migration.md
```diff
@@ -1522,6 +1523,27 @@ processors:
...
+in the same way as for translating metric names, by using the following config:
+
+```yaml
+processors:
+  sumologicschema/dockerstats:
+    translate_docker_metrics: true
+```
+
+In addition, there is some additional metadata sent by the OpenTelemetry Collector. Full list of it can be seen [here]
+[dockerstatsmetrics].
```
aboguszewski-sumo, line 1546: Should we describe that in detail?

sumo-drosiek: I don't think so

sumo-drosiek (approved):

PR #1083, opened 2023-03-21
docs: remove duplicates of help.sumologic.com

sumo-drosiek:
Update documentation after moving huge part of it to help.sumologic.com


swiatekm-sumo (approved):

swiatekm-sumo:
The changes look good, but we should also establish some process for updating the documentation under help.sumologic.com. Otherwise we'll diverge relatively quickly.

PR #1100, opened 2023-04-11
chore(scripts/install.sh): use associative arrays for tags

swiatekm-sumo:
Handle tags internally in the install script in an associative array, and only encode to yaml when writing to config files.

Currently we handle tags by passing them around as a yaml-encoded string. This is simple and convenient if that's all we need to do with them. What I'd like to do is also add them to data pipelines via the resource processor, where I need to generate a different representation.

I use some slightly edgy bash features, like associative arrays and reference variables. Would be happy to hear some feedback as to where I should add comments explaining what's going on.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -71,7 +71,7 @@ fi
 set -u
 
 API_BASE_URL=""
-FIELDS=""
+declare -A FIELDS=()
 VERSION=""
 FIPS=false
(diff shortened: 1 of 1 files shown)


swiatekm-sumo:
Turns out darwin has bash 3.2 by default, so no associative arrays for me. :(

PR #1112, opened 2023-04-28
chore: upgrade otel core to 0.76.1

swiatekm-sumo:
The contrib release in this case is 0.76.3. I've added a note to the release doc to sort this out during the next upgrade.


aboguszewski-sumo (approved):
Changelog entry is missing, otherwise LGTM

swiatekm-sumo:
The plugins check failing is expected, as the core and contrib versions are different.

PR #1116, opened 2023-04-28
fix: missing components

swiatekm-sumo:
Some components were accidentally removed in a previous change, adding them back.


aboguszewski-sumo (approved):
Works: https://github.com/SumoLogic/sumologic-otel-collector/actions/runs/4831221797/jobs/8608374253?pr=1115

PR #1127, opened 2023-05-12
feat(scripts): update install.sh to use packages on macOS

amdprophet:
Updates `scripts/install.sh` to install otelcol-sumo using packages when run on macOS. I forked the packaging repository to create an example release for testing. Set `PACKAGE_GITHUB_ORG="sensu"` inside `install.sh` for testing installation from a GitHub Release, or use `--binary-branch main` to test installing from GitHub Action artifacts.

These changes should be broken up into separate functions and the install script should be refactored to account for platform differences but @portertech wants to tackle that separately from this initial implementation.


sumo-drosiek (commented):
Are `psutil` and `launchcl` available on all mac platforms?

It would be nice to have e2e tests for this PR, as it is quite significant

amdprophet:
Yes, both plutil & launchctl exist for all macOS platforms and have been there for 10+ years. E2E tests are a great idea; I'll fix up the broken test(s) and start to add coverage today.

portertech:
Fixed an uninstallation bug https://github.com/SumoLogic/sumologic-otel-collector-packaging/pull/10

portertech:
@amdprophet @sumo-drosiek @perk-sumo We cannot merge this PR until the UI team is ready for the rollout across deploys. The UI team is putting a dynamic property in place to pin the MacOS installation script release (i.e. 0.79.0 instead of latest). Once this is done, we can then merge this PR and cut a collector release. The UI team can then flip a feature flag on the Staging deployment and QE can run through the E2E test suites (and make corrections etc). After E2E tests pass, the UI team can do a slow rollout across deploys (days).

@amdprophet in the meantime, please get builds passing and do further cleanup. I don't expect a green light until late next week 😓

pkg/scripts_test/check_darwin.go
```diff
@@ -29,9 +29,14 @@ func checkConfigFilesOwnershipAndPermissions(ownerName string, ownerGroup string
...
-					if filepath.Dir(path) != confDPath || path == userConfigPath {
+					if path == configPath {
+						// /etc/otelcol-sumo/sumologic.yaml
 						permissions = configPathFilePermissions
+					} else if path == userConfigPath {
+						// /etc/otelcol-sumo/conf.d/common.yaml
+						permissions = commonConfigPathFilePermissions
 					} else {
+						// /etc/otelcol-sumo/conf.d/*
 						permissions = confDPathFilePermissions
```
sumo-drosiek: I think we may use `switch` here

scripts/install.sh
```diff
@@ -723,11 +723,11 @@ function setup_config_darwin() {
     write_sumologic_extension "${COMMON_CONFIG_PATH}" "${INDENTATION}"
 
     # fill in api base url
-    if [[ -n "${API_BASE_URL}" && -z "${USER_API_URL}" ]]; then
+    if [[ -n "${API_BASE_URL}" ]]; then
```
sumo-drosiek, line 726: Do we need this change? If yes, can we linux with it?

sumo-drosiek (approved):
two minor things

scripts/install.sh
amdprophet: This is necessary on macOS as `common.yaml` is replaced on upgrades. I haven't been able to find a way around this. I'll update the logic for Linux to match.

PR #1129, opened 2023-05-15
ci: run tests before building binaries

swiatekm-sumo:
The main reason to do this is to cache the test dependencies and results. Github Actions caches are immutable, so we want to run tests first to ensure their results are cached.

We also run tests on main before dev builds for the same reason.

### DIFF
--- a/.github/workflows/dev_builds.yml
+++ b/.github/workflows/dev_builds.yml
@@ -44,10 +44,163 @@ jobs:
 
       - name: Run golangci-lint
         run: make golint
+  test:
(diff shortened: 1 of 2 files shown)


sumo-drosiek (approved):

swiatekm-sumo:
Holding for now, this might not be the best way to achieve better caching in the CI.

swiatekm-sumo:
Closing in favor of #1131

swiatekm-sumo:
Closing in favor of #1131

PR #1130, opened 2023-05-16
test: run install script tests separately

swiatekm-sumo:
Install script tests require root and modify the environment. For convenience, they shouldn't run with the rest of package tests. Building the tests as root also causes intermittent permission issues with the mod cache.


sumo-drosiek (approved):
Looks good to me

PR #1133, opened 2023-05-17
ci: test and build in one job

swiatekm-sumo:


swiatekm-sumo:
This was done in #1172

PR #1137, opened 2023-05-22
chore: update Go to 1.19.9 for local development

astencel-sumo:


sumo-drosiek (dismissed):

vagrant/provision.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-export GO_VERSION=1.19
+export GO_VERSION="1.19.9"
```
sumo-drosiek, line 3: Does it make sense?

sumo-drosiek: 1.19 ensures we are always the latest, and with `1.19.9` we may forget to update it

astencel-sumo: Without it, v1.19 is downloaded, which is in fact v1.19.0.

PR #1145, opened 2023-05-31
fix: do not send html special characters as unicode

sumo-drosiek:
By default we are encoding htlm characters like `<`, `>`, `&`. It is unexpected for our customers

log:
```
{"log": "<<<<>>>>>"}
```

output to Sumo Logic
```
2023-05-31T06:01:09.357Z DEBUG [receiver_mock::router] log => {"log":"{\"log\": \"\u003c\u003c\u003c\u003c\u003e\u003e\u003e\u003e\u003e\"}","log.file.path_resolved":"/home/drosiek/projects/sumologic-otel-collector/tmp/logs/unicode.json","timestamp":1685512869155}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - chore: update OT to v0.78.0 [#1142]
 - feat(receiver/sqlquery): add experimental logs support [#1144]
 
+### Fixed
+
+- fix: do not send html special characters as unicode [#1145]
+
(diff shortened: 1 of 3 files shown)


sumo-drosiek:
The log in SumoLogic UI looks the same (`<`), but it doesn't work with search as expected (require using `\u003c`)

aboguszewski-sumo (approved):

astencel-sumo:
This issue only affected the `json` log formatting, is this correct?

swiatekm-sumo (approved):

PR #1150, opened 2023-06-01
docs: revert release v0.78.0-sumo-0

astencel-sumo:
The release failed due to a MacOS signing problem. We'll retry the release after the issue is fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,15 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-<!--
 ## [Unreleased]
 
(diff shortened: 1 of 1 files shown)


swiatekm-sumo:
Shouldn't we mark the release as failed instead and add `v0.78.0-sumo-1` instead? Like we did with 0.70.0 in #935 ?

astencel-sumo:
> Shouldn't we mark the release as failed instead and add `v0.78.0-sumo-1` instead? Like we did with 0.70.0 in #935 ?

I'm not sure what's the value in preserving this failed release `v0.78.0-sumo-0`. It wasn't really created. I would just remove the tag and create it again later.

In fact, I already removed the tag yesterday, so there is no `v0.78.0-sumo-0` release.

swiatekm-sumo:
In my opinion we should avoid messing with existing git tags as much as possible. For example, if someone pulls one of our components as a Go module using the tag, then that gets preserved in the Go proxy's package cache, leading to very confusing behaviour after we add that tag again on a different commit.

Since the tag is already gone though, we may as well proceed, so I'm approving this change.

swiatekm-sumo (approved):

PR #1152, opened 2023-06-02
fix(sumologicextension): check credentials dir at start

### LINKED ISSUE #1147: Do not register collector in case cred cannot be stored on disk
Right now OTC registers itself and then tries to store creds on disk.
In case the creds path is read-only OTC errors with:
```"mkdir /etc/otel/.sumologic-otel-collector: read-only file system"
open exporter_sumologic__logs: permission denied```

... and then re-register again - which leads to many registered but otherwise unusable collectors in the UI.

OTC should first check if creds path is writable and register only if that's the case.

### PR DESCRIPTION
**Pull request by swiatekm-sumo**
Fixes #1147. We simply try to ensure the credentials directory exists and has the right permissions when the credentials store is instantiated.

### DIFF
--- a/pkg/extension/sumologicextension/credentials/credentialsstore_localfs_test.go
+++ b/pkg/extension/sumologicextension/credentials/credentialsstore_localfs_test.go
@@ -18,6 +18,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pkg/extension/sumologicextension/credentials/credentialsstore_localfs_test.go [hunk 1]
```diff
@@ -74,3 +74,17 @@ func TestCredentialsStoreLocalFs(t *testing.T) {
...
+func TestCredentialsStoreEnsureWritable(t *testing.T) {
+	dir := filepath.Join(t.TempDir(), "store")
+	expectedFileMode := fs.FileMode(0700)
+	err := os.Mkdir(dir, 0400)
+	require.NoError(t, err)
+
+	_, err = NewLocalFsStore(WithCredentialsDirectory(dir), WithLogger(zap.NewNop()))
+	require.NoError(t, err)
+
+	stat, err := os.Stat(dir)
```
rnishtala-sumo, line 87: doesn't `ensureDir` already create the directory with the right permissions? why do we have `os.Stat` here again...

rnishtala-sumo (approved):

PR #1153, opened 2023-06-05
fix(sumoextension): skip credentials dir check if registration is forced

swiatekm-sumo:
It's feasible that someone might want to run the collector without a writable credentials directory. In that case, we should continue running if `force_registration: true` is set, instead of failing at start time.

This PR also correctly identifies #1152 as a breaking change in the changelog.


sumo-drosiek (approved):
LGTM

PR #1173, opened 2023-06-22
chore: test otelcol configs from example directory

dmolenda-sumo:
- move otelcol configs to separate folder in examples
- test config_logging.yaml and logs_json/config.yaml in collector_config_test.go (only these 2 because they don't contain sumologic extension which want to connect with sumo during startup)
- fix bug in logs_json/config.yaml


examples/README.md
```diff
@@ -1,12 +1,5 @@
 # Examples
 
 - [ansible](ansible) - example Ansible playbook to install Sumo Logic Distribution for OpenTelemetry Collector
-- [config_cpu_load_metrics.yaml](config_cpu_load_metrics.yaml) - example configuration to collect CPU load metrics using [Host Metrics Receiver][hostmetricsreceiver]
 - [puppet](puppet) - example manifest and module to install Sumo Logic Distribution for OpenTelemetry Collector with Puppet
 - [chef](chef) - example cookbook to install Sumo Logic Distribution for OpenTelemetry Collector with Chef
```
swiatekm-sumo, line 5: We should link to `./otelcolconfigs` her as well.

examples/otelcolconfigs/logs_json/config.yaml
```diff
@@ -6,12 +6,12 @@ exporters:
 receivers:
   filelog:
     include:
-    - /sumologic/examples/logs_json/example.log
+    - /sumologic/examples/otelcolconfigs/logs_json/example.log
```
swiatekm-sumo, line 9: This path is relative to otelcol's working directory, and it won't pick up the file this way unless we're in the vagrant VM. Maybe assume we're running from the project root here? So it'd be `example/otelcolconfigs/logs_json/example.log`.

swiatekm-sumo (approved):
LGTM

PR #1176, opened 2023-06-23
fix(build): fix dev Dockerfile

swiatekm-sumo:
The dev Dockerfile uses the official Go image as the base for the final stage, and copies journalctl and dependencies from a debian image. This is consistent with the production image, but that uses scratch, while the golang image also uses Debian as base. As a result, if the debian versions diverge, it's possible to get cryptic errors about glibc versions, as has recently happened in our CI. See for example https://github.com/SumoLogic/sumologic-otel-collector/actions/runs/5355217286/jobs/9713524369.

Instead, skip the intermediate step and install systemd in the final image directly.


sumo-drosiek:
With that, we are going to introduce the following differences between dev and prod images:

- different `journalctl` binary
- different process of building docker image

Are we fine with it?

swiatekm-sumo:
> With that, we are going to introduce the following differences between dev and prod images:
> 
>     * different `journalctl` binary
> 
>     * different process of building docker image
> 
> 
> Are we fine with it?

I don't think it's that big a deal, but if we want to stay with the current setup, we should probably use a Debian image and manually install Go instead of using the Go image directly. 

Why do we even want the dev image to be different from the prod one? Additional tooling? In that case it should be named differently (debug), and we should publish it for prod as well. WDYT?

sumo-drosiek:
> Why do we even want the dev image to be different from the prod one? Additional tooling? In that case it should be named differently (debug), and we should publish it for prod as well. WDYT?

Not sure. I would expect prod image to build. Dev-builds should be as close as possible to production ones

.github/workflows/dev_builds.yml
```diff
@@ -153,7 +153,7 @@ jobs:
         if: matrix.arch_os == 'linux_amd64'
         run: |
           cp otelcol-sumo-fips-${{ matrix.arch_os }} otelcol-sumo
-          make build-push-container-multiplatform-dev \
+          make build-push-container-multiplatform \
             BUILD_TAG=${{ steps.extract_tag.outputs.tag }}-fips \
             PLATFORM=${{ matrix.arch_os }} \
             LATEST_TAG_FIPS_SUFFIX="-fips"
```
sumo-drosiek, line 159: Shouldn't we set `REPO_URL` to dev one?

```
REPO_URL="$(OPENSOURCE_REPO_URL_DEV)"
```

Makefile
```diff
@@ -248,35 +248,26 @@ build-container-local:
...
-		REPO_URL="$(OPENSOURCE_REPO_URL_DEV)" \
-		DOCKERFILE="Dockerfile_dev" \
-		PLATFORM="$(PLATFORM)" \
-		./ci/build-push-multiplatform.sh $(PUSH)
-
 .PHONY: build-container-multiplatform-dev
-build-container-multiplatform-dev:
-	$(MAKE) _build-container-multiplatform-dev PUSH=
+build-container-multiplatform-dev: REPO_URL="$(OPENSOURCE_REPO_URL)"
+build-container-multiplatform-dev: build-container-multiplatform
```
sumo-drosiek, line 260: Does it work as expected :open_mouth: ?

swiatekm-sumo: No, but it's pretty close. I set `REPO_URL` to `OPENSOURCE_REPO_URL` by default, and then used the former in the release targets.

sumo-drosiek: so it works as
```
build-container-multiplatform-dev: REPO_URL="$(OPENSOURCE_REPO_URL)" build-container-multiplatform
```
?

swiatekm-sumo: Works more or less like:
```
build-container-multiplatform-dev:  
  make build-container-multiplatform REPO_URL="$(OPENSOURCE_REPO_URL)"
```

except it doesn't start a new shell, and you can do more stuff in the variable assignment.

sumo-drosiek (approved):

PR #1184, opened 2023-07-04
chore(packaging): wix 4.0.1

amdprophet:
Bumps the WiX toolset version from 4.0.0-preview.1 to 4.0.1.

In order to get this to build I had to update some of our custom WixUI code:
* The UI element's `Id` attribute now requires the platform that the UI is built for.
* The `Print` control for the EULA had no action and did nothing. I've added the proper line to fix it.

I found a few other recommendations and one other bug:
* The `Next` control for the `CustomizeDlg` page was going to `VerifyReadyDlg` instead of `ConfigurationDlg`. This prevented the dialog that asks for the `InstallToken` and `Tags` from appearing unless the  `Back` control was clicked on the `VerifyReadyDlg` page.
* Removed the `PlatformProgramFilesFolder` variable and logic to determine which Program Files directory to use depending on the platform. This is now handled by a new `ProgramFiles6432Folder` built-in to WiX 4.
* Both `INSTALLFOLDER` and `APPDATAFOLDER` directory elements have had the manufacturer name (Sumo Logic) removed from their names. They've been wrapped in a new directory elements that represent the manufacturer name.
* The name of the `OtelcolBinary` value for x86 now uses `386` instead of `arm64`. This will only matter in the event that we ever produce a `windows/386` distribution of the collector.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- chore(packaging): wix 4.0.1 [#1184]
+
+[#1184]: https://github.com/SumoLogic/sumologic-otel-collector/pull/1184
+
 ## [v0.80.0-sumo-0]
 
 ### Released 2023-06-30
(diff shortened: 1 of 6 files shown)


portertech:
@amdprophet any change in the Program Files directory logic itself? This going to break collector updates in some specific cases? Also, is this a typo? "value for x86 now uses 386 instead of **_arm64_**."

amdprophet:
@portertech no change in behaviour, it's just how the folders are represented within WiX itself. Not a typo, there's branch logic for x86 in `variables.wsi` that was looking for an arm64 binary -- I know we don't have an x86 binary but it's fixed now in the event that we ever have to release an x86 installer.

sumo-drosiek (approved):

PR #1194, opened 2023-07-12
fix: mute more errors when collecting host metrics

astencel-sumo:
Use two more configuration properties
to prevent unneeded error logs from the collector.


swiatekm-sumo (approved):
Please make sure this is also changed in packaging and the internal content repository.

sumo-drosiek (approved):

PR #1195, opened 2023-07-12
feat: add single quotes around configuration keys

sumo-drosiek:
The main motivation of this PR is to handle the following case (`_` as separator for numbers):

`--tag numeric-tag=1_024` which should be saved as:

```
numeric-tag: '1_024' 
```

but was saved as:

```
numeric-tag: 1024 
```


astencel-sumo (approved):
:see_no_evil: 

At least there are some tests, thank you for this :pray:

sumo-drosiek:
> At least there are some tests, thank you for this pray

This is only code I;m sure about :sweat_smile:

PR #1197, opened 2023-07-18
feat(sumologicexporter): add option to decompose OTLP Histograms

swiatekm-sumo:
Sumo doesn't currently support OTLP Histograms. In the interim, we can at least make sure sending a histogram via OTLP and Prometheus gives the same result. We already have the logic for the Prometheus format, and this PR also implements it for OTLP. This comes down to decomposing the Histogram into individual sum, count and bucket metrics.

The behaviour is configurable and disabled by default.

### DIFF
--- a/pkg/exporter/sumologicexporter/otlp.go
+++ b/pkg/exporter/sumologicexporter/otlp.go
@@ -1,0 +1,160 @@
+// Copyright 2023, OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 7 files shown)


pkg/exporter/sumologicexporter/otlp.go
```diff
@@ -0,0 +1,156 @@
...
+// This exists because Sumo doesn't support OTLP histograms yet, and has the same semantics as the conversion to Prometheus format in prometheus_formatter.go
+func DecomposeHistograms(md pmetric.Metrics) pmetric.Metrics {
+	// short circuit and do nothing if no Histograms are present
+	foundHistogram := false
+	for i := 0; i < md.ResourceMetrics().Len(); i++ {
+		resourceMetric := md.ResourceMetrics().At(i)
+		for j := 0; j < resourceMetric.ScopeMetrics().Len(); j++ {
+			scopeMetric := resourceMetric.ScopeMetrics().At(j)
+			for k := 0; k < scopeMetric.Metrics().Len(); k++ {
+				foundHistogram = foundHistogram || scopeMetric.Metrics().At(k).Type() == pmetric.MetricTypeHistogram
```
sumo-drosiek, line 34: Shouldn't we break out of the loops?

swiatekm-sumo: Yeah, we could, not really sure which way is cleaner.

swiatekm-sumo: Done.

sumo-drosiek (approved):

Issue #875, opened 2022-12-19
Remove remaining data-modifying options from sumologic exporter


swiatekm-sumo:
Exporters shouldn't modify data, as this messes with fan-out in pipelines and leads to race conditions with multiple exporters. We have a few remaining options in the Sumo exporters that we should either get rid of or move to the sumo schema processor.

- [x] `clear_logs_timestamp` - move to sumo schema processor, or just use transform processor?
- [x] `routing_attribute_to_drop` - drop, routing processor now does this
- [x] `json_logs.add_timestamp` and `json_logs.flatten_body` shouldn't modify the otlp record

swiatekm-sumo:
See https://github.com/SumoLogic/sumologic-kubernetes-collection/pull/2933 for examples on how these features can be reproduced using the transform processor.

echlebek:
@swiatekm-sumo necessary to reproduce these features elsewhere before removal? Are there other issues filed for this?

echlebek:
Thanks, it seems pretty straightforward!

sumo-drosiek:
https://github.com/SumoLogic/sumologic-otel-collector/pull/1201
https://github.com/SumoLogic/sumologic-otel-collector/pull/1372 
https://github.com/SumoLogic/sumologic-otel-collector/pull/1371

swiatekm-sumo:
#1373 and #1378 should be closed before we close this issue.

PR #1201, opened 2023-07-25
chore: remove routing_attributes_to_drop

echlebek:
The upstream issue that necessitated this configuration was closed in 2022, and special treatment is no longer required.

### DIFF
--- a/.changelog/1201.breaking.txt
+++ b/.changelog/1201.breaking.txt
@@ -1,0 +1,3 @@
+chore: remove routing_attributes_to_drop
+
+This option is duplication of `drop_resource_routing_attribute` of routingprocessor
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -4,7 +4,7 @@
   - [Upgrading to v0.90.0-sumo-1](#upgrading-to-v0900-sumo-1)
     - [Change configuration for `syslogexporter`](#change-configuration-for-syslogexporter)
     - [`sumologic` exporter: deprecate `clear_logs_timestamp`](#sumologic-exporter-deprecate-clear_logs_timestamp)
-    - [Removing `routing_attributes_to_drop` from `sumologic` exporter](#removing-routing_attributes_to_drop-from-sumologic-exporter)
+    - [`sumologic` exporter: remove `routing_attributes_to_drop`](#sumologic-exporter-remove-routing_attributes_to_drop)
   - [Upgrading to v0.89.0-sumo-0](#upgrading-to-v0890-sumo-0)
     - [`remoteobserver` processor: renamed to `remotetap` processor](#remoteobserver-processor-renamed-to-remotetap-processor)
     - [`sumologic` exporter: changed default `timeout` from `5s` to `30s`](#sumologic-exporter-changed-default-timeout-from-5s-to-30s)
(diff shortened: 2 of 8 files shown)


sumo-drosiek (changes requested):
We should deprecate it before removing it. Not sure if the deprecation process is described, but this is more less how it looks like:

1. add deprecation information in readme, log deprecation warning in case argument is used and also add migration document
   https://github.com/SumoLogic/sumologic-otel-collector/pull/685/files
   Create issue with task for removal and add milestone with target tag version
1. remove feature except config and deprecation message (which need to be changed to removed attribute message), so customers have clear information what to do (in opposite to error which only indicates that the configuration doesn't exist anymore):
   https://github.com/SumoLogic/sumologic-otel-collector/pull/686/files

echlebek:
@sumo-drosiek Understood, such a strategy for deprecation makes complete sense. However, I'm a little confused, since #875 didn't mention deprecation, and when I found the attribute, it appeared to have already been deprecated to some extent, as evidenced by the comment here: https://github.com/SumoLogic/sumologic-otel-collector/blob/main/pkg/exporter/sumologicexporter/config.go#L60

As for deprecation warnings, the deprecation warning you describe appears to have been removed since that PR landed, so I guess that these are also being cleaned up over time. Or, has the deprecation warning logic itself been moved?

I'm left wondering about what the actual stage of the maintenance lifecycle this attribute is in. Perhaps at this stage, only step #2 of your suggested workflow needs to be executed? Or perhaps the attribute was not actually deprecated as the comment describes.

I'm also wondering if this particular configuration option should really be carried forward, since it appears that it was only introduced as a workaround in the first place and doesn't seem to be mentioned in the README.

Thanks for any insight you can provide!

sumo-drosiek:
> @sumo-drosiek Understood, such a strategy for deprecation makes complete sense. However, I'm a little confused, since #875 didn't mention deprecation, and when I found the attribute, it appeared to have already been deprecated to some extent, as evidenced by the comment here: https://github.com/SumoLogic/sumologic-otel-collector/blob/main/pkg/exporter/sumologicexporter/config.go#L60

Hmm, it should be added to the issue. Removing configuration option is a breaking change and we should be informative about it and provide migration document for such changes. It seems that this change has been deprecated, but only in the comment which you pointed out.
   
> As for deprecation warnings, the deprecation warning you describe appears to have been removed since that PR landed, so I guess that these are also being cleaned up over time. Or, has the deprecation warning logic itself been moved?

Yes, the deprecation warnings can be cleaned after some time. We didn't formalize it, but we keep them for few versions as it is more informative as default error from Otelcol.

> I'm left wondering about what the actual stage of the maintenance lifecycle this attribute is in. Perhaps at this stage, only step 2 of your suggested workflow needs to be executed? Or perhaps the attribute was not actually deprecated as the comment describes.
> I'm also wondering if this particular configuration option should really be carried forward, since it appears that it was only introduced as a workaround in the first place and doesn't seem to be mentioned in the README.

Maybe you're right, as this configuration has been added specifically for kubernetes collection. I would definitely go with deprecation document and maybe warning message, just in case

sumo-drosiek (approved):
@echlebek seems that this needs to be merged as we want to stop modyfing data in exporter. It was used only in helm chart so I think we can safely proceed with it. I can take over the PR to add remaining changes. WDYT?

echlebek:
@sumo-drosiek I had forgotten about this! Please feel free to take it over. :bow:

swiatekm-sumo (approved):

PR #1217, opened 2023-08-08
feat(monitoringjobreceiver): Add Component Outline

c-kruse:
Closes https://sumologic.atlassian.net/browse/SUMO-226995

This is the unimplemented monitoring jobs receiver component, and is not yet included in the sumo distro.

Intended for review of configuration and basic readme contents.


swiatekm-sumo (commented):
Overall this looks reasonable. One thing I'd like to be clarified more is the format of the output data. Are we just emitting logs where the body is the output string from the command, without any additional attributes?

c-kruse:
> Overall this looks reasonable. One thing I'd like to be clarified more is the format of the output data. Are we just emitting logs where the body is the output string from the command, without any additional attributes?

@swiatekm-sumo Thank you for the read - apologies for the large followup CL. That got me thinking and I've made a few changes I hope will make it clear what data I think monitoring jobs should produce.
- multiline support: makes the default newline delimited log event behavior clear, keeps basic feature parity w/ sumo's scirpt sources and supports the more sensu-like user story. I am not sure how useful this is without it.
- add `attributes` and `resource` configuration to let users specify log and resource attributes.
- include `command.name` and `command.output.stream` attributes. Can't say I've got a clear idea of what best practice would be here. I'd say I feel medium strong about labeling the output stream so as to not crossing stderr with stdout and less strong about including command name. @portertech  may have opinions? It'd be sweet to include attributes for command exit code, maybe execution time, but I'd be at least a little afraid of buffering all of the output.

swiatekm-sumo:
If we want to let the user do some basic processing on the raw data before converting it to Otel data structures, we may want to consider adopting stanza: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/operators/README.md#what-operators-are-available. There's a bunch of receivers in contrib which have the similar problem (most notably the [filelog receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver)), and they solve it using stanza operators. Then we don't need to expose lots of specific configuration, as users will be able to add whatever attributes they want, do multiline processing, regex parsing, etc.

c-kruse:
Starting anew with stanza too!

PR #1219, opened 2023-08-10
feat(scripts): add linux deb & rpm package support to install script

amdprophet:
**WIP:** This is still a WIP but is nearly complete. There are a few test failures for deb packages that should be fixed shortly.

Adds support for Linux deb & rpm packages to the install script. A new `--use-native-packaging` flag has been added to use package installation instead of the current binary installation method.

https://github.com/SumoLogic/sumologic-otel-collector-packaging/pull/28 contains fixes for packages that are required for the package installation via `install.sh` to function. A new package release is required after that PR is merged.

### DIFF
--- a/pkg/scripts_test/check.go
+++ b/pkg/scripts_test/check.go
@@ -18,6 +18,7 @@ import (
 type check struct {
 	test                *testing.T
 	installOptions      installOptions
+	installType         installType
 	code                int
 	err                 error
 	expectedInstallCode int
@@ -55,6 +56,44 @@ func checkRun(c check) {
 	require.Equal(c.test, c.expectedInstallCode, c.code, "unexpected installation script error code")
 }
 
+func checkConfigFilesOwnershipAndPermissions(c check) {
+	ownerName := getConfigFilesOwner(c)
+	groupName := getConfigFilesGroup(c)
+	etcPathGlob := filepath.Join(etcPath, "*")
+	etcPathNestedGlob := filepath.Join(etcPath, "*", "*")
+
+	for _, glob := range []string{etcPathGlob, etcPathNestedGlob} {
+		paths, err := filepath.Glob(glob)
+		require.NoError(c.test, err)
+		for _, path := range paths {
+			var permissions uint32
+			info, err := os.Stat(path)
+			require.NoError(c.test, err)
+			if info.IsDir() {
+				permissions = configPathDirPermissions
+			} else {
+				switch path {
+				case configPath:
+					// /etc/otelcol-sumo/sumologic.yaml
+					permissions = configPathFilePermissions
+				case userConfigPath:
+					// /etc/otelcol-sumo/conf.d/common.yaml
+					permissions = commonConfigPathFilePermissions
+				case tokenEnvFilePath:
+					// /etc/otelcol-sumo/env/token.env
+					permissions = commonConfigPathFilePermissions
+				default:
+					// /etc/otelcol-sumo/conf.d/**/
+					permissions = confDPathFilePermissions
+				}
+			}
+			PathHasPermissions(c.test, path, permissions)
+			PathHasOwner(c.test, configPath, ownerName, groupName)
+		}
+	}
+	PathHasPermissions(c.test, configPath, configPathFilePermissions)
+}
+
 func checkConfigCreated(c check) {
 	require.FileExists(c.test, configPath, "configuration has not been created properly")
 }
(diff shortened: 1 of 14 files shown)


pkg/scripts_test/install_unix_test.go
```diff
@@ -7,21 +7,58 @@ import (
...
+		checkConfigNotCreated,
+		checkUserConfigNotCreated,
+		checkUserNotExists,
+		checkGroupNotExists,
+		checkUserConfigNotCreated,
+		checkSystemdConfigNotCreated,
+		checkDeprecatedSystemdConfigNotCreated,
+	}
+
 	for _, spec := range []testSpec{
```
swiatekm-sumo, line 21: Is there a simple way to avoid all this duplication? For example, take the testspec slice, run all the tests as they are before this change, and then run them again manually enabling native packaging?

pkg/scripts_test/check.go
```diff
@@ -56,7 +56,13 @@ func checkRun(c check) {
 }
 
 func checkConfigDirectoryOwnershipAndPermissions(c check) {
-	PathHasOwner(c.test, etcPath, rootUser, rootGroup)
+	ownerUser := systemUser
+	ownerGroup := systemGroup
+	if !c.installOptions.useNativePackaging {
+		ownerUser = rootUser
+		ownerGroup = rootGroup
```
swiatekm-sumo, line 63: Why is this different? Do the native packages not set the same owner as the install script?

swiatekm-sumo (commented):
What are these Python scripts and docker-compose config for? Since they're so big, I'd prefer to add them in a separate PR, with some minimal documentation on how to use them, maybe some additional tooling.

pkg/scripts_test/check.go
amdprophet: When `SYSTEMD_DISABLED=true`, the install script doesn't add the otelcol-sumo user and thus the files are owned by root. The packages only support systemd and thus will always make use of the otelcol-sumo user.

amdprophet:
> What are these Python scripts and docker-compose config for? Since they're so big, I'd prefer to add them in a separate PR, with some minimal documentation on how to use them, maybe some additional tooling.

The docker-compose config is a way to run both the tests against both deb & rpm packages on platforms that don't support it (e.g. macOS). It was quite helpful developing against it. I plan to add a job to test RPMs in CI as well.

Systemd requires PID 1 which doesn't easily work with containers. The Python scripts are a hack to simulate systemd & journalctl for the purpose of testing.

I agree that the Dockerfiles, config & Python scripts don't belong in this PR. I'll open a separate PR for that work and add some documentation.

pkg/scripts_test/install_unix_test.go
amdprophet: The main issue is that the PreActions, PreChecks, PostChecks, etc. differ between binary & package installs in most if not all of the specs. Reducing duplication may not be worth the time investment as we're planning on removing binary installation support from the install script.

swiatekm-sumo: Intuitively, the install script should give the same results whether using native packaging or not, at least with systemd enabled. What is the source of the differences?

amdprophet: The majority of the difference is in the pre-actions & pre-checks.

I avoided using much of the config mocking used in the tests to prevent package installation conflicts around existing config files. This isn't a problem when installing under an interactive tty as the installer can prompt how to handle conflicts. To work around this issue I wrote a helper to install a previous version of the package (defined by the const `previousVersion`) in some of the preActions of specs. This should work with binary installations but versions differ between binaries and packages to allow for out-of-band changes to packaging code.

The binary install uses the wrong path for the systemd unit file which has been fixed in the packages. I had assumed, perhaps incorrectly, that we wouldn't want to change paths of any files for binary installations. We can change the path to match the packages. We'll need logic to handle the removal or migration of the unit file at the old path.

amdprophet:
I'm currently refactoring the specs & checks to remove a lot of the duplication. In doing so I discovered that I broke FIPS installation, hostmetrics installation with Linux packages only, and a few other install paths.

The ownership of config files differs between, not only binary and package installs but, non-systemd binary and systemd binary installs. This is why some of the specs check for `rootUser` as an owner and others check for systemUser. I'd like to drop non-systemd support in the install script to further simplify it. Is this something that we can do? @swiatekm-sumo @portertech

pkg/scripts_test/install_unix_test.go
swiatekm-sumo: Allright, I'm fine with that. Can we then move the native test cases to a separate test? There's a lot of them and I'd like to introduce some basic organization here, to make changes easier in the future.

portertech:
@amdprophet once we're "all-in" with native packages, we can look to deprecate/remove non-service managed installations.

sumo-drosiek:
@amdprophet What is the status of this change?

echlebek:
@amdprophet can this be closed, since we are moving to the packaging repo?

amdprophet:
We reimplemented this work in https://github.com/sumologic/sumologic-otel-collector-packaging.

PR #1225, opened 2023-08-21
feat(sumologicextension)!: remove support for install_token

sumo-drosiek:
Remove support for `install_token` which has been deprecated long time ago


swiatekm-sumo (approved):

FlyingRedDonkey:
Hey this update broke my shit, thanks

astencel-sumo:
Sorry to hear that @FlyingRedDonkey. We understand it's not ideal to introduce breaking changes, and yet we still choose to make them sometimes, to keep the product comprehensible and in good shape, or we are forced by other factors.

The [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector) itself is in a `v0.x.y` versioning scheme, which by definition of the [semantic versioning](https://semver.org/) means breaking changes may happen.

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

Same goes for the Sumo Logic distribution of the collector.

We try as much as possible to keep the number of breaking changes low and we don't take the decision to introduce ones lightly. The deprecation of `install_token` in favor of `installation_token` in the Sumo Logic extension has been introduced in [v0.72.0-sumo-0](https://github.com/SumoLogic/sumologic-otel-collector/blob/main/CHANGELOG.md#v0720-sumo-0) in February 2023, which is over 6 months ago. We also explicitly call out the breaking changes in the [changelog](https://github.com/SumoLogic/sumologic-otel-collector/blob/main/CHANGELOG.md) and in the [release notes](https://github.com/SumoLogic/sumologic-otel-collector/releases), and also add [upgrade instructions](https://github.com/SumoLogic/sumologic-otel-collector/blob/main/docs/upgrading.md) whenever possible.

If you have any thoughts on what we could or should be doing differently, please let us know by opening a new issue. We'd love to get more feedback.

PR #1226, opened 2023-08-22
Add monitorinjob receiver component outline

c-kruse:
Includes configuration and documentation for the new monitoringjob
receiver.  Also includes the necessary plumbing to build and start a
stanza pipeline based off the receiver's output configuration with an
exposed custom input to that pipeline - a pattern that is unique to this
receiver.

Signed-off-by: Christian Kruse <ctkruse99@gmail.com>


pkg/receiver/jobreceiver/runner.go
```diff
@@ -0,0 +1,33 @@
...
+	return &stubRunner{Consumer: out}, nil
+}
+
+// stubRunner is a stub implementation.
+type stubRunner struct {
+	Consumer consumer.Interface
+}
+
+func (r *stubRunner) Start(operator.Persister) error {
+	var buf bytes.Buffer
```
c-kruse, line 25: This is where we'd start our main loop. Scheduling and executing commands.

PR #1227, opened 2023-08-22
Add monitoringjob receiver component outline

c-kruse:
Includes configuration and documentation for the new monitoringjob receiver.  Also includes the necessary plumbing to build and start a stanza pipeline based off the receiver's output configuration with an exposed custom input to that pipeline - a pattern that is unique to this receiver.


pkg/receiver/jobreceiver/builder/config.go
```diff
@@ -0,0 +1,69 @@
+package builder
```
c-kruse, line 1: This package is basically the glue between Stanza's receiver adapter and our code.

NewOperatorBuilder gives stanza a way to build an input operator to use, hooks up one of our output handlers (event or log entries) to that input operator, and hooks up our main loop to that output handler through the consumer interface.

pkg/receiver/jobreceiver/runner.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/SumoLogic/sumologic-otel-collector/pkg/receiver/jobreceiver/builder"
+	"github.com/SumoLogic/sumologic-otel-collector/pkg/receiver/jobreceiver/output/consumer"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/operator"
+	"go.uber.org/zap"
+)
+
+// Build returns the job runner, the process responsible for scheduling and
+// running commands and piping their output to the output consumer.
+func (c Config) Build(logger *zap.SugaredLogger, out consumer.Interface) (builder.JobRunner, error) {
+	return &stubRunner{Consumer: out}, nil
```
c-kruse, line 16: This is where we'd set up and return a real implementation that schedules & executes the jobs.

c-kruse:
@astencel-sumo working on sharing my findings with the folks upstream.

I ended up buying in to the stanza adapter way of running the receiver for now - starting with adapter.NewFactory. I feel it may be less readable than if we called receiver.WithLogs ourself, but it certainly spared us a good bit of low-ish level stanza setup which is what pushed me this way. If we do end up deciding to switch things up I feel okay about our configuration's stability.

c-kruse:
@rnishtala-sumo link to the epic
https://sumologic.atlassian.net/browse/SUMO-224185

pkg/receiver/jobreceiver/README.md
```diff
@@ -0,0 +1,130 @@
...
+monitoringjob:
+  schedule:
+    interval: 1h
+  exec:
+    command: check_ntp_time
+    arguments:
+        - "-H"
+        - time.nist.gov
+    timeout: 8s
+  output:
```
rnishtala-sumo, line 123: is there a reason that we need the output config in the receiver? Why not use the stanza operator in the receiver config like so:
```
receiver:
  monitoringjob:
    schedule:
       interval: 1h
    operators:
      - type: regex_parser
           ....
```

You may be able to use the [regex parser](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/operators/regex_parser.md), to parse/ include specific fields from the incoming log/event.

pkg/receiver/jobreceiver/README.md
```diff
@@ -0,0 +1,130 @@
...
+    type: event
+    event:
+      include_command_name: false
+      include_command_status: true
+      include_command_duration: true
+      max_body_size: '32kib'
```
rnishtala-sumo, line 129: Removing the stanza stuff from the receiver may simplify receiver configuration.

pkg/receiver/jobreceiver/README.md
c-kruse: @rnishtala-sumo I don't have particularly strong feelings on _where_ the stanza config lives. Maybe this will explain the choice to put it under `output`.

Take a look at output `type`. We want the receiver to run these jobs, but the output can be handled in two radically different ways. The default (and what I suspect the original intent in design was) type is event. Where the full output gets bundled up into a single log entry and emitted with a full context of the execution. The other option, `log_entires`, is meant to cover some popular existing sumo use cases (Script Sources) where the command output is streamed line by line.

We could probably do away with a nested _output_, but I thought it made more sense given the two output types.

pkg/receiver/jobreceiver/README.md
c-kruse: At the moment I'd tend to agree. Long term, though I think we'll regret not having it.

Take a look at the Sumo Logic Script Source docs for example - there's a good bit of configuration for parsing the raw log output client side. I think we are trying to head off needing to come back to add features like this piecemeal.

pkg/receiver/jobreceiver/README.md
c-kruse: Recapping a conversation had out of band.

I set up the output using the existing upstream conventions in similar receivers like `filelogreceiver` including multi line processing, max log entry size. It is intended to be protective over collector resources.

An interesting experiment may be splitting individual log records in a "streaming" processor as Raj has suggested.

pkg/receiver/jobreceiver/README.md
```diff
@@ -0,0 +1,131 @@
...
+### Runtime Asset Configuration
+
+//todo(ck)
+
+## Additional Features
+
+### Time Parameters
+
+Time parameters are specified as sets of decimal numbers followed by a
+unit suffix. e.g. `60s`, `45m`, `2h30m40s`.
```
swiatekm-sumo, line 104: We can link to the source of truth here: https://pkg.go.dev/time#ParseDuration

pkg/receiver/jobreceiver/output/config.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	// Load non-dynamic parts like normal
+	if err := component.Unmarshal(c); err != nil {
+		return err
+	}
+
+	// Load non-dynamic parts like normal
+	if err := component.Unmarshal(c); err != nil {
+		return err
+	}
```
swiatekm-sumo, line 62: This second unmarshal seems unnecessary.

swiatekm-sumo (approved):
Looks good to me overall. Left some minor comments.

PR #1232, opened 2023-08-28
chore: switch to Node.js version of Markdownlint

astencel-sumo:
- Removes the need to have Ruby installed to run `make markdownlint`
- Now you need to have Node.js to run `make markdownlint` (use `make install-markdownlint` to install the linter)
- If you don't have Node.js, use `make markdownlint-docker` to run the containerized version with Docker


.github/workflows/pull_requests.yml
```diff
@@ -24,13 +24,13 @@ jobs:
         with:
           files: |
             **/*.md
-      - uses: ruby/setup-ruby@v1
+      - uses: actions/setup-node@v3
```
c-kruse, line 27: I suspect the semantics of ruby/setup-ruby aren't valid for actions/setup-node.

https://github.com/actions/setup-node#usage

I am a github actions novice but I suspect we want something like this.
```
- uses: actions/setup-node@v3
  if: steps.changed-files.outputs.any_changed == 'true'
  with:
    node-version: 18
    cache: 'npm'
```

c-kruse (commented):
Thank you @astencel-sumo! Other than nailing down caching in CI I am content to call this a quality of life improvement.

.github/workflows/pull_requests.yml
astencel-sumo: Oh, right, missed that 🙈

c-kruse (approved):

PR #1235, opened 2023-09-05
chore: update otelcol core to v0.84.0

astencel-sumo:


aboguszewski-sumo (commented):
Add a changelog entry and fix failing CI please (rerunning should be enough, but who knows)

swiatekm-sumo (approved):

PR #1240, opened 2023-09-07
feat(jobreceiver): Implement the command package

c-kruse:
Implements `pkg/receiver/jobreceiver/command/command`

I had a hard time with this one. I started by copying the work @portertech did in the PoC that had been adapted from sensu, but I ran into several issues.

First. The sensu command package had lots of opinionated behavior that I felt was not appropriate to copy over. Things like timeout behavior (hard coding exit code 2 and writing a blurb to stderr). More difficult to suss out was the bespoke timeout, process group and cleanup logic that all seemed geared towards the assumption of a shell environment (which is no longer obligated.)

Second. Our usage pattern here is different than Sensu's. Sensu's command interface buffers the entirety of command output before returning it to the caller. Here we need to stream that output to the pipeline to accommodate the log_entries output format. Handling this inner process communication is actually quite nuanced. Go's os/exec package takes a reasonable approach to handling this with exec.Cmd.Std[out|err]Pipe but it wasn't particularly compatible with sensu's command interface or the way it was doing timeouts.

In the end I decided to lean towards default behavior and accept that we may end up needing to re-learn some of lessons in the future.


echlebek (commented):
It looks good to me but I suppose we should also wait for review from someone who has been on the OSC team a bit longer. I found a few doc nits.

pkg/receiver/jobreceiver/command/command_test.go
```diff
@@ -0,0 +1,183 @@
...
+		})
+		_, err := childCommand.Run()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "fork error: %v", err)
+			os.Exit(3)
+		}
+	}
+}
+
+//nolint:all
```
swiatekm-sumo, line 67: Why do we need these tags?

pkg/receiver/jobreceiver/command/command_test.go
```diff
@@ -0,0 +1,183 @@
...
+func eventualOutput(i *Execution) <-chan string {
+	out := make(chan string, 1)
+	stdout, _ := i.Stdout()
+	stderr, _ := i.Stderr()
+	go func() {
+		var buf syncBuffer
+		io.Copy(&buf, stdout)
+		io.Copy(&buf, stderr)
```
swiatekm-sumo, line 161: Why do we need synchronization here?

pkg/receiver/jobreceiver/command/exec_linux.go
```diff
@@ -0,0 +1,15 @@
+package command
```
swiatekm-sumo, line 1: For the platform specific `setOptions`, can we pick the right one at runtime instead of using build tags? Doesn't look like we use any platform-specific symbols.

swiatekm-sumo (commented):
Looks good overall, had some questions about tests and the platform-specific bits.

pkg/receiver/jobreceiver/command/command_test.go
c-kruse: Good catch! We don't - just an artifact of one revision of this.

pkg/receiver/jobreceiver/command/exec_linux.go
c-kruse: Afraid the syscall.SysProcAttr structure is os specific.

`for GOOS in "linux" "darwin" "windows"; go doc -short syscall.SysProcAttr`

swiatekm-sumo (approved):

PR #1249, opened 2023-09-12
fix(install.sh): ensure selinux records are created

amdprophet:
A bug currently exists where selinux records are sometimes not set for the otelcol-sumo binary. This will occur if the systemd unit file exists when installing but the otelcol-sumo binary does not. For example, attempting to install after an uninstallation without purging configs.

I've removed the logic that checks if the systemd unit file already exists. This fixes the bug described and will also ensure that users are always using the correct systemd unit file for the version of the collector they're installing.


c-kruse (approved):
Worked for me! (Fedora Linux 38)

Issue #1242, opened 2023-09-08
[k8sprocessor] Owner metadata should have a grace period after deletion


swiatekm-sumo:
We keep Pod metadata around after the Pod itself is deleted, to account for telemetry data that might be processed after the deletion takes place. However, we do not do this for owner metadata, whereas we should for the same reason. As is, if a Deployment is deleted and we process data from a Pod belonging to that Deployment afterwards, we won't add the `deployment` attribute.

I believe this was introduced in #710, where we started calculating owner metadata dynamically instead of keeping it alongside Pod metadata.

c-kruse:
@swiatekm-sumo  if you don't mind I'd give this a shot.

Reading through, the most apparent solution would be to reapply that same deferred deletion queue pattern from the pod identity cache to the Owner cache. Probably with some twist to account for the multiple cached collections on that structure. Does that all track?

swiatekm-sumo:
Yeah, I think the same solution will work modulo different types.

PR #1256, opened 2023-09-21
fix(k8sprocessor): owner meta deletion grace period

c-kruse:
Adds a deferred deletion queue to the owner metadata cache, allowing time for telemetry data to be processed with correct metadata after the owning resource(s) have been deleted.

Closes https://github.com/SumoLogic/sumologic-otel-collector/issues/1242


pkg/processor/k8sprocessor/kube/owner.go
```diff
@@ -469,3 +489,37 @@ func (op *OwnerCache) GetOwners(pod *Pod) []*ObjectOwner {
...
+			op.deleteMu.Lock()
+			for i, d := range op.deleteQueue {
+				if d.ts.Add(gracePeriod).After(now) {
+					break
+				}
+				cutoff = i + 1
+			}
+			toDelete := op.deleteQueue[:cutoff]
+			op.deleteQueue = op.deleteQueue[cutoff:]
+			op.deleteMu.Unlock()
```
echlebek, line 511: I feel this is worth its own method, and then the Unlock can be deferred

echlebek (commented):
There seems to be a large amount of commented out code in one of the test files that this PR touches. I would suggest deleting it if possible.

c-kruse:
@swiatekm-sumo you cool with yanking the commented out benchmarks? They had drifted a bit even before this CL. Saw that you had been "using" them in earlier PR/issue discussion.

swiatekm-sumo (commented):
The overall approach looks good to me. I'd like:
1. Comments for the new functions
2. A test that actually measures the deletion delay
3. Maybe a manual E2E test in Kubernetes to make sure this is ok? I'm not sure this is necessary, but it's not too difficult to do using the Helm Chart integration tests. Let me know if you want to try it, but if not, I'm ok merging this change as-is.

c-kruse:
@swiatekm-sumo Updated godocs and tests with assertions on honoring the grace period.

Have dug into the E2E tests and may yet manually validate the change.

swiatekm-sumo (approved):

PR #1257, opened 2023-09-22
chore(ci): build fips binary w/glibc 2.26

amdprophet:
The `linux/amd64` FIPS binary is currently built using GLIBC 2.28. Some supported Linux distributions use an older version of GLIBC and the binary we produce is incompatible as a result. This change updates CI to build the binary using the `amazonlinux:2` container image which has GLIBC 2.26.

### DIFF
--- a/ci/build-fips-action/Dockerfile
+++ b/ci/build-fips-action/Dockerfile
@@ -6,6 +6,10 @@ ENV TARGETARCH=$TARGETARCH
 
 RUN yum groupinstall -y "Development Tools" && yum install -y curl git
 
+RUN curl -Lo go.tar.gz https://go.dev/dl/go1.20.5.linux-$TARGETARCH.tar.gz
+RUN tar -zxvf go.tar.gz -C /usr/local
+ENV PATH="/usr/local/go/bin:${PATH}"
+
 COPY entrypoint.sh /entrypoint.sh
 
 ENTRYPOINT ["/entrypoint.sh"]
--- a/ci/build-fips-action/Dockerfile
+++ b/ci/build-fips-action/Dockerfile
@@ -1,0 +1,11 @@
+FROM amazonlinux:2
+
+ARG TARGETARCH
+
+ENV TARGETARCH=$TARGETARCH
+
+RUN yum groupinstall -y "Development Tools" && yum install -y curl git
+
+COPY entrypoint.sh /entrypoint.sh
+
+ENTRYPOINT ["/entrypoint.sh"]
--- a/ci/build-fips-action/entrypoint.sh
+++ b/ci/build-fips-action/entrypoint.sh
@@ -2,13 +2,6 @@
 
 git config --global --add safe.directory /github/workspace
 
-# Install Go
-url="https://go.dev/dl/go${GO_VERSION}.linux-${TARGETARCH}.tar.gz"
-echo "Downloading ${url}"
-curl -Lo go.tar.gz "$url"
-tar -zxvf go.tar.gz -C /usr/local
-export PATH="/usr/local/go/bin:${PATH}"
-
 # Install builder
 cd otelcolbuilder || exit 1
 mkdir "${HOME}/bin"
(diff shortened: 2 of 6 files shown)


ci/build-fips-action/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM amazonlinux:2
+    MAINTAINER Justin Kolberg <amd.prophet@gmail.com>
```
sumo-drosiek, line 2: Do we need this indentation?

sumo-drosiek:
Could we add only Dockerfile like for other builds and add makefile target. W have makefile to be independent on the CI provider and to be able to easily run everything locally

ci/build-fips-action/entrypoint.sh
```diff
@@ -0,0 +1,19 @@
...
+echo "Downloading ${url}"
+curl -Lo go.tar.gz "$url"
+tar -zxvf go.tar.gz -C /usr/local
+export PATH="/usr/local/go/bin:${PATH}"
+
+# Install builder
+cd otelcolbuilder || exit 1
+mkdir "${HOME}/bin"
+export PATH="${HOME}/bin:${PATH}"
+make install-builder
```
swiatekm-sumo, line 16: These install commands seem like they'd be more at home in the Dockerfile.

amdprophet: I couldn't find a way of passing the Go version into the Dockerfile itself from GitHub Actions.

amdprophet:
> Could we add only Dockerfile like for other builds and add makefile target. W have makefile to be independent on the CI provider and to be able to easily run everything locally

I can try adding a makefile target to run this locally.

ci/build-fips-action/entrypoint.sh
amdprophet: I moved the Go installation steps to the Dockerfile and hardcoded the Go version for now. I was hoping to minimize the amount of places where we need to bump the Go version but it does cut down on runtimes.

I don't think it's possible to use anything except the `ci/build-fips-action` directory as the context when GitHub Actions builds the Docker image. I've left the builder installation in `entrypoint.sh` as a result.

otelcolbuilder/Makefile
```diff
@@ -134,6 +134,11 @@ build-debug: ensure-correct-builder-version
 	@$(MAKE) generate-sources
 	@$(MAKE) _gobuild_debug
 
+.PHONY: build-fips
+build-fips:
+	docker build --platform linux/amd64	-t otelcol-sumo-builder-fips -f ../ci/build-fips-action/Dockerfile ../ci/build-fips-action
```
sumo-drosiek, line 139:
```suggestion
	docker build --platform linux/amd64 -t otelcol-sumo-builder-fips -f ../ci/build-fips-action/Dockerfile ../ci/build-fips-action
```

ci/build-fips-action/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM amazonlinux:2
+
+ARG TARGETARCH
+
+ENV TARGETARCH=$TARGETARCH
+
+RUN yum groupinstall -y "Development Tools" && yum install -y curl git
+
+RUN curl -Lo go.tar.gz https://go.dev/dl/go1.20.5.linux-$TARGETARCH.tar.gz
```
portertech, line 9: Not happy about the hard-coded go version here, but I understand why it is. Going to be a bit of a pain when we update go.

portertech (approved):

otelcolbuilder/Makefile
```diff
@@ -134,6 +134,11 @@ build-debug: ensure-correct-builder-version
 	@$(MAKE) generate-sources
 	@$(MAKE) _gobuild_debug
 
+.PHONY: build-fips
```
portertech, line 137: I was able to successfully build locally.

portertech:
CI test run https://github.com/SumoLogic/sumologic-otel-collector/actions/runs/6331827931

PR #1258, opened 2023-09-22
feat(jobreceiver): Implement monitoring job receiver

c-kruse:
More of an integration PR. I've taken some care to organize this work into discrete commits so that they may be reviewed separately, but I'm not super confident that each could be rebased directly on main in a tidy way. Happy to hear the team's thoughts on how y'all like to review work.

## CLs

* Implement the `event` output handler 
* Implement the `log_entries` output handler
* Implement the receiver ~~(behind feature gate)~~
* ~~Include the receiver in the sumo distro~~

## Demo

Install otel collector per usual, then replace the binary with one built from this branch.
Add this configuration to the globbed config directory:
```
# /etc/otelcol-sumo/conf.d/monitoringjobs.yaml

receivers:
  monitoringjob/log:
    schedule:
      interval: 1m
    exec:
      command: /bin/sh
      timeout: 1m
      arguments:
        - '-c'
        - 'for i in {0..120}; do echo -n "warning: " && date -Iseconds --date "90 seconds ago" && sleep 1 && echo "multi line $(head -c 12 /dev/urandom | base64)"; done;'
    output:
      type: log_entries
      log_entries:
        include_command_name: true
        include_stream_name: true
        max_log_size: '16kb'
        encoding: 'utf-8'
        multiline:
          line_start_pattern: '\w+:.*'
      attributes:
        type: log
      resource:
        demo: '#1258'
      operators:
        - type: regex_parser
          parse_from: body
          regex: '^(?P<level>\w+):\s+(?P<ts>.*)'
        - type: severity_parser
          parse_from: attributes.level
        - type: time_parser
          parse_from: attributes.ts
          layout_type: gotime
          layout: '2006-01-02T15:04:05-07:00'
        - type: move
          from: attributes["command.name"]
          to: resource["_sourceName"]
 monitoringjob/event:
    schedule:
      interval: 10s
    exec:
      command: df
      arguments:
        - '-h'
    output:
      type: event
      event:
        include_command_name: true
        include_command_status: true
        include_command_duration: true
        max_body_size: 1024
      resource:
        demo: '#1258'
      operators:
        - type: move
          from: attributes["command.name"]
          to: resource["_sourceName"]


service:
  pipelines:
    logs/monitoringjob:
      receivers:
        - monitoringjob/event
        - monitoringjob/log
      processors:
        - memory_limiter
        - batch
        - resourcedetection/system
      exporters:
        - sumologic
```

See output in sumo:
image

image

### DIFF
--- a/pkg/receiver/jobreceiver/README.md
+++ b/pkg/receiver/jobreceiver/README.md
@@ -10,14 +10,6 @@ executable at defined intervals, and propagates the output from that process
 as log events. In addition, the monitoring job receiver simplifies the process
 of downloading runtime assets necessary to run a particular monitoring job.
 
-## Feature Gate
-
-This receiver is currently gated behind a [featuregate][featuregate]
-`receiver.monitoringjob.enabled`. To start the collector with monitoringjobs
-enabled use the `--feature-gates=receiver.monitoringjob.enabled` flag.
-
-[featuregate]: https://github.com/open-telemetry/opentelemetry-collector/tree/main/featuregate
-
 ## Configuration
 
 | Configuration | Default | Description
--- a/pkg/receiver/jobreceiver/README.md
+++ b/pkg/receiver/jobreceiver/README.md
@@ -10,6 +10,14 @@ executable at defined intervals, and propagates the output from that process
 as log events. In addition, the monitoring job receiver simplifies the process
 of downloading runtime assets necessary to run a particular monitoring job.
 
+## Feature Gate
+
+This receiver is currently gated behind a [featuregate][featuregate]
+`receiver.monitoringjob.enabled`. To start the collector with monitoringjobs
+enabled use the `--feature-gates=receiver.monitoringjob.enabled` flag.
+
+[featuregate]: https://github.com/open-telemetry/opentelemetry-collector/tree/main/featuregate
+
 ## Configuration
 
 | Configuration | Default | Description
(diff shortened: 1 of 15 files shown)


swiatekm-sumo:
Is there a reason you want to use a feature gate here? A user would need to explicitly opt-in to using the receiver anyway by including it in their configuration, so I'm not sure what the additional hurdle of a feature gate adds here.

I'd also recommend adding the receiver to the build (and also changing the stability to Alpha?) in a separate PR. The rest of these changes are fine together imo.

c-kruse:
> Is there a reason you want to use a feature gate here? A user would need to explicitly opt-in to using the receiver anyway by including it in their configuration, so I'm not sure what the additional hurdle of a feature gate adds here.

There is a reason I wanted an explicit opt-in. A larger hangup in this design was around mitigating a (rather involved) vector where a misconfiguration could make that assertion in that second sentence not totally true 😄. The use of a feature gate for that purpose is probably semantically misleading. Yanking the feature gate and the inclusion into the build for now.

swiatekm-sumo:
> > Is there a reason you want to use a feature gate here? A user would need to explicitly opt-in to using the receiver anyway by including it in their configuration, so I'm not sure what the additional hurdle of a feature gate adds here.
> 
> There is a reason I wanted an explicit opt-in. A larger hangup in this design was around mitigating a (rather involved) vector where a misconfiguration could make that assertion in that second sentence not totally true 😄. The use of a feature gate for that purpose is probably semantically misleading. Yanking the feature gate and the inclusion into the build for now.

If what we want to achieve is make the receiver available for testing, but not in release builds, we could only add it to dev builds to start with.

PR #1265, opened 2023-10-03
fix(extension/sumologic): send non-empty autodiscovery tag values

astencel-sumo:
Sumo backend apparently does not allow empty tag values, so let's set the `sumo.disco.*` tag values to a non-empty value.


pkg/extension/sumologicextension/extension.go
```diff
@@ -742,7 +742,7 @@ func discoverTags() (map[string]interface{}, error) {
 	}
 
 	for _, v := range pl {
-		t["sumo.disco."+v] = "" // We do not currently need a value, save bytes.
+		t["sumo.disco."+v] = 1 // Sumo does not allow empty tag values, let's set it to anything.
 	}
```
aboguszewski-sumo, line 746: If we want to save bytes, wouldn't it be better to set this to `bool` in order to use only one byte? (I don't know how marshalling later works, so I might be wrong).

aboguszewski-sumo: For reference: 
```go
package main

import (
	"fmt"
	"unsafe"
)

func main() {
	fmt.Printf("%d\n", unsafe.Sizeof(1))
	fmt.Printf("%d\n", unsafe.Sizeof(true))
}

```

swiatekm-sumo: For [protobuf](https://protobuf.dev/programming-guides/encoding/#structure), Boolean values are encoded as `VARINT`, same as int32, so this doesn't make any difference in terms of serialized size. 

For JSON, it's `true` vs `1.0`, so the int is actually shorter.

The boolean seems more intuitively appropriate here though. We're indicating the presence of a tag after all.

astencel-sumo: I would call this is a classic example of "premature optimization" :ok_hand: Let's not spend cycles on such a tiny little detail. Let's move on.

swiatekm-sumo: I don't think this matters for performance, but it does matter for ergonomic imo. We want to choose a value that will make the most sense when rendered by the Sumo UI and processed by any further logic. I don't know if we have any convention for tags without values, but we should use if it we do.

swiatekm-sumo (approved):

PR #1269, opened 2023-10-03
ci: add fips build & msi for windows/amd64

amdprophet:
Adds a new `windows/amd64` FIPS binary and a new corresponding MSI package. Test CI run: https://github.com/SumoLogic/sumologic-otel-collector/actions/runs/6397614173

I validated the new FIPS binary is installed by the new MSI package by both verifying the checksum of the installed `otelcol-sumo.exe` file and using `go tool nm .\otelcol-sumo.exe | Select-String vendor/github.com/microsoft/go-crypto-winnative/internal/bcrypt.GetFipsAlgorithmMode`.

### DIFF
--- a/otelcolbuilder/Makefile
+++ b/otelcolbuilder/Makefile
@@ -3,6 +3,7 @@ BUILDER_VERSION ?= 0.87.0
 FIPS_SUFFIX ?= ""
 BUILDER_REPO ?= github.com/open-telemetry/opentelemetry-collector
 GO ?= go
+GOTAGS ?=
 
 ifeq ($(OS),Windows_NT)
 	OS=windows
@@ -51,14 +52,20 @@ GOEXPERIMENT=""
 LDFLAGS="-s -w"
 # We want to use boringcrypto for fips setup
 else
+ifeq ($(shell go env GOOS),windows)
+GOEXPERIMENT=cngcrypto
+GOTAGS := $(GOTAGS) requirefips
+LDFLAGS="-w"
+else
 GOEXPERIMENT=boringcrypto
 LDFLAGS="-w"
+endif
 endif
 
 # We want to enforce go dns for all types of binaries. The FIPS binaries are built with CGO enabled and use
 # the CGO resolver. This has caused a DNS resolution error for .local domains in our K8s containers.
 # ref: https://pkg.go.dev/net#hdr-Name_Resolution
-GOFLAGS=-tags=netgo
+GOTAGS := $(GOTAGS) netgo
 
 .PHONY: _install-bin
 _install-bin:
@@ -103,7 +110,8 @@ _builder:
 .PHONY: _gobuild
 _gobuild:
 	(cd cmd && \
-		GOFLAGS=$(GOFLAGS) CGO_ENABLED=$(CGO_ENABLED) GOEXPERIMENT=$(GOEXPERIMENT) go build -v \
+		CGO_ENABLED=$(CGO_ENABLED) GOEXPERIMENT=$(GOEXPERIMENT) go build -v \
+		-tags "$(GOTAGS)" \
 		-ldflags=$(LDFLAGS) \
 		-trimpath \
 		-o ./$(BINARY_NAME)$(BUILDER_BIN_EXT) . \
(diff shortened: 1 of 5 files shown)


amdprophet:
The `release-builds` workflow still needs to be updated to build the new MSI. The `install.ps1` script will also need to be updated to support the new MSI. I was thinking that work should be done in a separate PR but I can add it to this PR if desired.

sumo-drosiek (approved):

swiatekm-sumo:
As far as I know, BoringCrypto is not FIPS certified for Windows.

amdprophet:
> As far as I know, BoringCrypto is not FIPS certified for Windows.

Their documentation does state:

```
It is important to note that an application built with Microsoft's Go toolchain and running in FIPS compatible mode is not FIPS compliant per-se
```

I'm a bit confused by this statement however https://github.com/microsoft/go/tree/microsoft/main/eng/doc/fips:

```
On Linux, the fork uses [OpenSSL](https://www.openssl.org/) through the golang-fips/openssl module in Go 1.21+ and the go-crypto-openssl module in earlier versions. On Windows, [CNG](https://docs.microsoft.com/en-us/windows/win32/seccng/about-cng), using go-crypto-winnative. Similar to BoringSSL, certain OpenSSL and CNG versions are FIPS 140-2 certified.
```

The Microsoft fork links against CNG and I'm fairly sure that I tested that startup failed when FIPS mode wasn't enabled in CNG (I'll re-verify this, I may have forced `requirefips` when testing). 

I may be wrong but it seems like the same or a similar issue exists on Linux. BoringCrypto is FIPS certified but from my understanding `GOEXPERIMENT=boringcrypto` does not guarantee that otelcol-sumo will always make use of it. Am I missing something and if so can we provide a Windows binary with the caveat that we can't guarantee FIPS compliancy?

swiatekm-sumo:
I actually missed you were using Microsoft's fork, my bad. I'm fine merging this as-is so long as we only produce dev builds  using the CNG toolchain. If we want to produce release builds and actually claim FIPS compliance for them, we need to document that properly in https://github.com/SumoLogic/sumologic-otel-collector/blob/main/docs/fips.md.

> I may be wrong but it seems like the same or a similar issue exists on Linux. BoringCrypto is FIPS certified but from my understanding GOEXPERIMENT=boringcrypto does not guarantee that otelcol-sumo will always make use of it. Am I missing something and if so can we provide a Windows binary with the caveat that we can't guarantee FIPS compliancy?

You can force the Linux build to run in FIPS-approved mode by importing https://go.googlesource.com/go/+/dev.boringcrypto/src/crypto/tls/fipsonly/fipsonly.go, which we do here: https://github.com/SumoLogic/sumologic-otel-collector/blob/main/otelcolbuilder/cmd/fips.go. Maybe something similar is available in the CNG build?

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - feat: add `debug` exporter [#1268]
 - feat: add `routing` connector [#1268]
+- feat(ci): add fips build & msi for windows/amd64 [#1269]
```
swiatekm-sumo, line 18: If we're only providing dev builds for now, let's skip the changelog entry to avoid confusing users.

otelcolbuilder/Makefile
```diff
@@ -51,9 +51,13 @@ GOEXPERIMENT=""
 LDFLAGS="-s -w"
 # We want to use boringcrypto for fips setup
 else
+ifeq ($(shell go env GOOS),windows)
+GOEXPERIMENT=cngcrypto
```
swiatekm-sumo, line 55: Don't you want to keep debug symbols here just as we do for the BoringCrypto build?

swiatekm-sumo (approved):

otelcolbuilder/Makefile
amdprophet: Good catch, I re-added them.

amdprophet:
@swiatekm-sumo I've added the `requirefips` build tag after doing some additional testing. The binary will panic when Windows CNG isn't running in FIPS mode or if `GOFIPS=0`.

PR #1273, opened 2023-10-05
feat(windowsadinvreceiver): Adding the Windows Active Directory inventory receiver to the collector

rnishtala-sumo:
The Windows Active Directory Inventory receiver collects inventory data from Active Directory. This includes information such as computer names, user names, email addresses, and location information

Uses this library to query data from windows ADSI
https://pkg.go.dev/github.com/go-adsi/adsi

Example configuration (consistent with AD attributes in [IC](https://help.sumologic.com/docs/send-data/installed-collectors/sources/windows-active-directory-inventory-source/))

```
receivers:
  activedirectoryinv:
    base_dn: "CN=Guest,CN=Users,DC=exampledomain,DC=com"
    attributes: [name, mail, department, manager, memberOf]
    poll_interval: 10h
```

Example output

```
2023-10-13T15:16:08.161Z        info    ResourceLog #0
Resource SchemaURL:
ScopeLogs #0
ScopeLogs SchemaURL:
InstrumentationScope
LogRecord #0
ObservedTimestamp: 2023-10-20 18:09:26.2811961 +0000 UTC
Timestamp: 2023-10-20 18:09:26.2811961 +0000 UTC
SeverityText:
SeverityNumber: Unspecified(0)
Body:  Str({"memberOf":["CN=Guests,CN=Builtin,DC=exampledomain,DC=com"],"name":["Guest"]})
```

The body of the above log records contain data from the Windows AD server.


pkg/receiver/activedirectoryinvreceiver/adinvreceiver.go
```diff
@@ -0,0 +1,173 @@
...
+	l.logger.Debug("Shutting down logs receiver")
+	close(l.doneChan)
+	l.wg.Wait()
+	return nil
+}
+
+// Start polling for Active Directory inventory records
+func (l *ADReceiver) startPolling(ctx context.Context) {
+	defer l.wg.Done()
+	t := time.NewTicker(l.config.PollInterval * time.Second)
```
astencel-sumo, line 99: This multiplying of the `PollInterval` by `time.Second` is misleading. The `PollInterval` is already a duration, so treat it like one, instead of treating it like the number of seconds.

pkg/receiver/activedirectoryinvreceiver/README.md
```diff
@@ -0,0 +1,52 @@
...
+  activedirectoryinv:
+    # Base DN
+    base_dn: "CN=Users,DC=exampledomain,DC=com"
+
+    # User attributes
+    attributes: [name, mail, department, manager, memberOf]
+
+    # The polling interval.
+    # default = 60
+    poll_interval: 60
```
astencel-sumo, line 22:
```suggestion
    # default = 60s
    poll_interval: 60s
```

pkg/receiver/activedirectoryinvreceiver/README.md
```diff
@@ -0,0 +1,52 @@
+# Windows Active Directory inventory Receiver
+
+**Stability level**: Alpha
+
+A Windows Active Directory Inventory receiver collects inventory data from Active Directory. This includes information such as computer names, user names, email addresses, and location information
```
astencel-sumo, line 5: Does this receiver only work on Windows? If so, the README should mention this. Also, the collector should fail fast if configured with this receiver on a non-Windows machine.

pkg/receiver/activedirectoryinvreceiver/factory.go
```diff
@@ -0,0 +1,58 @@
...
+func NewFactory() receiver.Factory {
+	return receiver.NewFactory(
+		typeStr,
+		CreateDefaultConfig,
+		receiver.WithLogs(createLogsReceiver, component.StabilityLevelAlpha),
+	)
+}
+
+// CreateDefaultConfig creates the default configuration for the receiver
+func CreateDefaultConfig() component.Config {
```
astencel-sumo, line 40: Are these defaults consistent with the defaults for the [Windows Active Directory Inventory Source](https://help.sumologic.com/docs/send-data/installed-collectors/sources/windows-active-directory-inventory-source/)?

rnishtala-sumo: After checking again, it looks like the following should be the defaults

- DN = ""
- Attributes = [name, mail, department, manager, memberOf]
- PollInterval = 24 hours

pkg/receiver/activedirectoryinvreceiver/README.md
rnishtala-sumo: yes only windows, I'll look into this one

pkg/receiver/activedirectoryinvreceiver/factory.go
```diff
@@ -0,0 +1,58 @@
...
+	return &ADConfig{
+		DN:           "CN=Guest,CN=Users,DC=exampledomain,DC=com",
```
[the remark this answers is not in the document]
rnishtala-sumo: Yeah, I think this should be an empty string by default, since it comes under optional arguments in the [docs](https://help.sumologic.com/docs/send-data/installed-collectors/sources/windows-active-directory-inventory-source/)

pkg/receiver/activedirectoryinvreceiver/adinvreceiver.go
rnishtala-sumo: changed the poll interval to accept a duration string like "24h".

pkg/receiver/activedirectoryinvreceiver/README.md
rnishtala-sumo: the poll interval now accepts "60s"

pkg/receiver/activedirectoryinvreceiver/factory.go
rnishtala-sumo: The Distinguished name (DN) shouldn't be empty and must have a user domain. So the default value is a placeholder string.

pkg/receiver/activedirectoryinvreceiver/factory.go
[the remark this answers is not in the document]
astencel-sumo: Yes let's make it empty by default if it's optional.

pkg/receiver/activedirectoryinvreceiver/factory.go
rnishtala-sumo: will make the default empty and add a note about the format of the DN in the readme

pkg/receiver/activedirectoryinvreceiver/adinvreceiver_test.go
```diff
@@ -0,0 +1,164 @@
...
+	sink := &consumertest.LogsSink{}
+	mockClient := &MockClient{}
+	mockRuntime := &MockRuntime{}
+	mockRuntime.On("SupportedOS").Return(false)
+
+	logsRcvr := newLogsReceiver(cfg, zap.NewNop(), mockClient, mockRuntime, sink)
+
+	err := logsRcvr.Start(context.Background(), componenttest.NewNopHost())
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "activedirectoryinv is only supported on Windows")
```
rnishtala-sumo, line 126: @astencel-sumo added a test for returning an error on an unsupported OS.

pkg/receiver/activedirectoryinvreceiver/config.go
```diff
@@ -0,0 +1,61 @@
...
+	"regexp"
+	"time"
+)
+
+// ADConfig defines configuration for Active Directory Inventory receiver.
+
+type ADConfig struct {
+	DN           string   `mapstructure:"base_dn"` // DN is the base distinguished name to search from
+	Attributes   []string `mapstructure:"attributes"`
+	PollInterval string   `mapstructure:"poll_interval"`
```
astencel-sumo, line 28: The type of `PollInterval` should be `time.Duration` and not `string`. See for example https://github.com/open-telemetry/opentelemetry-collector/blob/v0.88.0/exporter/exporterhelper/retry_sender.go#L28

pkg/receiver/activedirectoryinvreceiver/factory.go
```diff
@@ -0,0 +1,59 @@
...
+	"context"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/receiver"
+)
+
+const (
+	// The value of "type" key in configuration.
+	typeStr = "activedirectoryinv"
```
astencel-sumo, line 27: There's an upstream [Active Directory Domain Services receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.88.0/receiver/activedirectorydsreceiver/README.md) that has type name `active_directory_ds`. Perhaps we should also use underscores in our receiver's name: `active_directory_inv`? Or `active_directory_inventory`?

pkg/receiver/activedirectoryinvreceiver/config.go
```diff
@@ -0,0 +1,61 @@
...
+func isValidDuration(durationStr string) bool {
+	_, err := time.ParseDuration(durationStr)
+	return err == nil
+}
+
+// Validate validates all portions of the relevant config
+func (c *ADConfig) Validate() error {
+
+	// Define the regular expression pattern for a valid Base DN
+	pattern := `^((CN|OU)=[^,]+(,|$))*((DC=[^,]+),?)*$`
```
astencel-sumo, line 46: Looking at this regex, it looks like an empty `base_dn` is allowed. Shouldn't the receiver fail to start when `base_dn` is not specified? By the way, is there a source of this regex that we could document (even if only as a comment on this `pattern` variable)?

pkg/receiver/activedirectoryinvreceiver/factory.go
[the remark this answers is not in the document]
astencel-sumo: I think I'm getting confused on this one. If an empty `base_dn` makes the receiver to not function property, then the empty value should probably be forbidden, making `base_dn` a required property. Is this correct?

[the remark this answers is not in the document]
rnishtala-sumo: It wouldn't collect any data. The user would get an error message saying that the DN is invalid and the collector would keep polling according to the poll interval.

pkg/receiver/activedirectoryinvreceiver/factory.go
rnishtala-sumo: I'll go with `active_directory_inventory`

pkg/receiver/activedirectoryinvreceiver/config.go
rnishtala-sumo: I'll describe what the regex is doing in a comment and check for an empty DN

pkg/receiver/activedirectoryinvreceiver/factory.go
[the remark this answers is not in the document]
rnishtala-sumo: As per our discussion, won't let the collector start without a valid non empty DN.

pkg/receiver/activedirectoryinvreceiver/config.go
rnishtala-sumo: changed the poll interval to `time.Duration`.

rnishtala-sumo:
@astencel-sumo this last [commit](https://github.com/SumoLogic/sumologic-otel-collector/pull/1273/commits/ccf543c202db9b08a107a316b21d1a59ed236a78) has all the changes

pkg/receiver/activedirectoryinvreceiver/factory.go
```diff
@@ -0,0 +1,60 @@
...
+		typeStr,
+		CreateDefaultConfig,
+		receiver.WithLogs(createLogsReceiver, component.StabilityLevelAlpha),
+	)
+}
+
+// CreateDefaultConfig creates the default configuration for the receiver
+func CreateDefaultConfig() component.Config {
+	return &ADConfig{
+		DN:           "CN=Users,DC=exampledomain,DC=com",
```
astencel-sumo, line 43: The `base_dn` property should be empty by default and required, meaning if the user doesn't specify it, the receiver fails to start.

rnishtala-sumo: just pushed the change to make this empty. The receiver doesn't start with empty value.

rnishtala-sumo:
@astencel-sumo I think everything that we've discussed so far has been covered. I've also validated the format of the logs with the installed collector. Its just a json log.

astencel-sumo (approved):
Thank Raj, I'm approving this request, although I would like to see two more very minor changes.

1. The README should make it clear that the `base_dn` configuration property is required.
2. (related) When user does not pass a non-empty value for the `base_dn` property, the error message should be more descriptive:
    - the error message should mention the configuration property name `base_dn`
    - ideally the error message would say bo more specifc, saying explicitly "`base_dn` is required" instead of generic "Invalid value of `base_dn`".

pkg/receiver/activedirectoryinvreceiver/config.go
```diff
@@ -0,0 +1,65 @@
...
+// ADConfig defines configuration for Active Directory Inventory receiver.
+
+type ADConfig struct {
+	DN           string        `mapstructure:"base_dn"` // DN is the base distinguished name to search from
+	Attributes   []string      `mapstructure:"attributes"`
+	PollInterval time.Duration `mapstructure:"poll_interval"`
+}
+
+var (
+	errInvalidDN           = errors.New("Base DN is incorrect, it must be a valid distinguished name (CN=Guest,OU=Users,DC=example,DC=com)")
```
astencel-sumo, line 32: Error message should specify the configuration property name `base_dn` to make it easier for user to fix the issue.

pkg/receiver/activedirectoryinvreceiver/config.go
```diff
@@ -0,0 +1,65 @@
...
+	errInvalidPollInterval = errors.New("poll interval is incorrect, invalid duration")
```
astencel-sumo, line 33: Error message should specify the configuration property name `poll_interval` explicitly, to make it easier for users to fix the issue.

pkg/receiver/activedirectoryinvreceiver/config.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"errors"
+	"regexp"
+	"time"
+)
+
+// ADConfig defines configuration for Active Directory Inventory receiver.
+
+type ADConfig struct {
+	DN           string        `mapstructure:"base_dn"` // DN is the base distinguished name to search from
```
astencel-sumo, line 26: Ideally the name of the variable in the config struct would be as close as possible to the configuration property name.
```suggestion
	BaseDN           string        `mapstructure:"base_dn"` // DN is the base distinguished name to search from
```

pkg/receiver/activedirectoryinvreceiver/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+Supported pipeline types: logs
+
+## Configuration
+
+```yaml
+receivers:
+  active_directory_inv:
+    # Base DN
+    # default = ""
```
astencel-sumo, line 15: Should mention that this property is required (i.e. empty value is not valid).

PR #1274, opened 2023-10-05
OpAMP load/save effective configuration using remote_configuration_directory

portertech:
Updated the OpAMP agent implementation to reflect the [OpAMP Sequence Diagram](https://github.com/SumoLogic/sumologic-otel-collector/files/12835260/OpAMP.Sequence.Diagram-2.pdf).

- OpAMP agent effective configuration is ONLY derived from the contents of `remote_configuration_directory`
- The contents of `remote_configuration_directory` are completely managed by the agent. Received remote configuration files are persisted here (replacing any existing contents).
- OpAMP remote configuration is now optional via `accepts_remote_configuration` (defaults to `true`). OpAMP can potentially be used for other purposes.


pkg/extension/opampextension/README.md
```diff
@@ -68,7 +67,7 @@ option:
 
 Here is a list of valid values for the OpAMP `endpoint** configuration option:
 
-**Note:** As of Jan 2023, these endpoints are not yet available.
```
sumo-drosiek, line 71: Do we need to specify date? It requires update once per month and as we can easily miss it

pkg/extension/opampextension/config.go
```diff
@@ -34,6 +36,9 @@ type Config struct {
 	// RemoteConfigurationDirectory is where received OpAMP remote configuration
 	// is stored.
 	RemoteConfigurationDirectory string `mapstructure:"remote_configuration_directory"`
+
+	// AcceptsRemoteConfiguration indicates if the OpAMP agent will accept remote configuration.
+	AcceptsRemoteConfiguration bool `mapstructure:"accepts_remote_configuration"`
```
sumo-drosiek, line 41: Is there a scenario when someone does want to use `opamextension` but dos not want to accept remote configuration?

pkg/extension/opampextension/README.md
portertech: Good point, removing date, it's unnecessary.

pkg/extension/opampextension/config.go
portertech: Yes. OpAMP can be used for other capabilities such as heartbeats, health, and effective configuration reporting. Core collection wanted this capability to be configurable at the start, before we build out the other capabilities.

pkg/extension/opampextension/opamp_agent.go
```diff
@@ -309,41 +315,74 @@ func (o *opampAgent) createAgentDescription() error {
...
-	defer f.Close()
 
-	_, err = f.Write([]byte(o.effectiveConfig))
+	files, err := d.Readdir(0)
 	if err != nil {
 		return err
 	}
 
+	for _, f := range files {
+		os.Remove(filepath.Join(dir, f.Name()))
```
echlebek, line 368: Errors aren't checked here, and in the worst case, this could lead to the wrong configuration being present in the directory with no error reported, if the deleted file isn't present in the list of files to be created.

echlebek (changes requested):

echlebek (approved):
kronk

PR #1276, opened 2023-10-06
Declare globprovider module path correctly

echlebek:
The globprovider module path did not equal the globprovider package path. This made it unusable with go get, which would complain that the module path and package paths were not equal.

After changing the module path, I fixed up the otelcolbuilder Makefile in order to ensure that it will always import at the latest version correctly.


sumo-drosiek (approved):
LGTM

Issue #1277, opened 2023-10-09
[k8sprocessor] Handle resource deletion on DeletedFinalStateUnknown


c-kruse:
The k8sprocessor uses a cache to store cluster resources and to update its model on change. In a degraded state, this cache can become disconnected from the apiserver and can miss deletion watch events. The cache will eventually reconcile this and notify the processor of the deletion and the resource's last known state using the tombstone [cache.DeletedFinalStateUnknown](https://pkg.go.dev/k8s.io/client-go/tools/cache#DeletedFinalStateUnknown). The processor is mishandling this notification.

Uncovered in #1267.

It is unclear to me if this particular issue is a primary contributing factor to the memory consumption problems observed in 1267.

c-kruse:
FWIW, I can reliably reproduce this error for Pods, but I can't seem to hit the similar case for the (owning) resources. If it is possible, I'm pretty sure it'd cause a panic.

PR #1278, opened 2023-10-10
Handle DeletedFinalStateUnknown events form cache

c-kruse:
Closes https://github.com/SumoLogic/sumologic-otel-collector/issues/1277


pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -193,11 +193,28 @@ func (c *WatchClient) handlePodUpdate(old, new interface{}) {
...
+	switch obj := obj.(type) {
+	case *api_v1.Pod:
+		pod = obj
+	case cache.DeletedFinalStateUnknown:
+		prev, ok := obj.Obj.(*api_v1.Pod)
+		if !ok {
+			c.logger.Error(
+				"object received was DeletedFinalStateUnknown but did not contain api_v1.Pod",
+				zap.Any("received", obj),
+			)
```
sumo-drosiek, line 208: is it logged often? We would like to avoid extensive error logging if this may be expected behavior

astencel-sumo: If I understand correctly, this is more of a sanity check that shouldn't trigger an error in normal situation, as by design the `cache.DeletedFinalStateUnknown` event should contain the object it pertains to.

astencel-sumo (approved):
Thanks a lot for this fix @c-kruse! :clap:

pkg/processor/k8sprocessor/kube/client.go
sumo-drosiek: I needed to double check if this function should always get pod object and confirm that as observer is run per object type

sumo-drosiek (approved):

PR #1287, opened 2023-10-11
feat(opamp): Add OpAmp configuration provider

echlebek:
The OpAmp configuration provider reads a yaml configuration file, extracting an opamp extension field named `remote_configuration_directory`. The remote configuration directory is expected to contain one or more files with an extension of `yaml`. After the directory is retrieved, the glob provider is used to read the yaml files from it.

The OpAmp configuration provider will only activate when a user supplies the --opamp-config flag, but is not hooked up yet, as the opampprovider module must first be published.


pkg/configprovider/opampprovider/provider.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	data, err := ioutil.ReadFile(url.Path)
+	if err != nil {
+		return nil, err
+	}
+	if err := yaml.Unmarshal(data, &cfg); err != nil {
+		return nil, err
+	}
+	glob := p.GlobProvider
+	return glob.Retrieve(ctx, glob.Scheme()+":"+cfg.RemoteConfigurationDirectory+"/*.yaml", fn)
```
portertech, line 63: Need to support Windows paths, e.g. `\`

echlebek:
@portertech Ready for review again, PTAL

pkg/configprovider/opampprovider/provider.go
```diff
@@ -0,0 +1,108 @@
...
+type ConfigFragment struct {
+	Extensions struct {
+		OpAmp struct {
+			RemoteConfigurationDirectory string `yaml:"remote_configuration_directory"`
+		} `yaml:"opamp"`
+	} `yaml:"extensions"`
+}
+
+func (c ConfigFragment) ConfigDir() string {
+	return c.Extensions.OpAmp.RemoteConfigurationDirectory
```
portertech, line 45: @echlebek please add some validation for this. While it is a required extension field, I think it makes sense to also require it in this provider for consistent behaviour.

pkg/configprovider/opampprovider/provider.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (c ConfigFragment) Validate() error {
+	if c.ConfigDir() == "" {
+		return errors.New("remote_configuration_dir missing")
```
portertech, line 51: @echlebek how about "opamp extension remote_configuration_directory must be specified"? More words and the correct field name.

portertech (approved):

Issue #1295, opened 2023-10-23
[sumologicextension] FQDN hostname resolution hangs in github actions hosts


c-kruse:
The fqdn library appears to occasionally get hung up running the darwin amd64 tests.

c-kruse:
As noted by @rnishtala-sumo, the collector already has a facility for getting the collector's hostname, the resource detection processor.

I think there is value in finding a way to allow for some consistency between this extension and what is configured in resourcedetection. As I understand, when installed and configured by sumo it should be consistent https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.88.0-sumo-0/examples/sumologic.yaml#L81 (try to resolve fqdn then us os.hostname), but the resourcedetection processor is configurable in ways this extension is not.

PR #1296, opened 2023-10-23
Bypass FQDN resolution in darwin tests

c-kruse:
Adds a hack to work around issue https://github.com/SumoLogic/sumologic-otel-collector/issues/1295

### DIFF
--- a/pkg/extension/sumologicextension/extension.go
+++ b/pkg/extension/sumologicextension/extension.go
@@ -986,10 +986,13 @@ func addClientCredentials(req *http.Request, credentials accessCredentials) {
 	req.Header.Add("Authorization", authHeaderValue)
 }
 
+// TODO(ck): hostname allows the darwin tests to bypass fqdn.
+var hostname = fqdn.FqdnHostname
+
 // getHostname returns the host name consistently with the resource detection processor's defaults
 // TODO: try to dynamically extract this from the resource processor in the pipeline
 func getHostname(logger *zap.Logger) (string, error) {
-	fqdnHostname, err := fqdn.FqdnHostname()
+	fqdnHostname, err := hostname()
 	if err == nil {
 		return fqdnHostname, nil
 	}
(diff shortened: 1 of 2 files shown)


echlebek (approved):
dirty

pkg/extension/sumologicextension/extension.go
```diff
@@ -986,10 +986,13 @@ func addClientCredentials(req *http.Request, credentials accessCredentials) {
 	req.Header.Add("Authorization", authHeaderValue)
 }
 
+// TODO(ck): hostname allows the darwin tests to bypass fqdn.
+var hostname = fqdn.FqdnHostname
+
 // getHostname returns the host name consistently with the resource detection processor's defaults
 // TODO: try to dynamically extract this from the resource processor in the pipeline
```
rnishtala-sumo, line 993: is it possible to do what this comment says? i.e extract the hostname from the resource processor.

c-kruse: Good question. I'd be surprised if extensions had a way to get processor state like that. I also doubt that'd be a productive way to go at least as far as tests are concerned. (i.e. setting up an entire pipeline with the resource detection processor configured to not use os instead of dns.)

Side note, I wager that processor has the same issue (when configured with dns)

rnishtala-sumo: I guess that solution would assume that we'd be using the resource processor in the pipeline like below.

```
service:
  extensions:
    - sumologic
  pipelines:
    logs:
      receivers: 
        - tcplog
      processors:
        - resource
```

c-kruse: @rnishtala-sumo  you okay with kicking this can down the road? Maybe noting that option in https://github.com/SumoLogic/sumologic-otel-collector/issues/1295

rnishtala-sumo: yes thats fine

rnishtala-sumo (approved):

PR #1298, opened 2023-10-24
chore: use Towncrier for Changelog management

astencel-sumo:


.changelog/changelog_template.jinja
sumo-drosiek, line 1: Could you add links to upstream releases as well?

astencel-sumo: Hm we actually never included the upstream release links in the Changelog. Can we make it a separate PR?

sumo-drosiek: Oh, it's changelog. We may skip it then

sumo-drosiek (approved):

swiatekm-sumo (approved):
We don't actually need to use prettier here. In https://github.com/SumoLogic/sumologic-kubernetes-collection we're using it to format MD and yaml files anyway, so it makes sense, but in this repo we don't have a formatter for markdown at all, so maybe it's not worth it?

astencel-sumo:
> We don't actually need to use prettier here. In https://github.com/SumoLogic/sumologic-kubernetes-collection we're using it to format MD and yaml files anyway, so it makes sense, but in this repo we don't have a formatter for markdown at all, so maybe it's not worth it?

I only added it because Towncrier insists on adding two newlines at the end of its Changelog entry, which needs to be replaced with a single newline for the Mardown linter to pass. Prettier is the easiest way to do this.

PR #1299, opened 2023-10-24
chore: add make targets to install pre-commit

astencel-sumo:


Makefile
```diff
@@ -45,6 +45,14 @@ markdown-link-check:
 shellcheck:
 	./ci/shellcheck.sh
 
+.PHONY: install-pre-commit
+install-pre-commit:
+	python3 -m pip install pre-commit
```
sumo-drosiek, line 50: What about target to install python3 :hand_over_mouth:

sumo-drosiek (approved):

Makefile
astencel-sumo: We have to make some assumptions 😄 Our Makefile currentkly assumes that you have both Python 3 and Node.js installed.

PR #1306, opened 2023-10-26
fix(install): install policycoreutils on RHEL if missing

aboguszewski-sumo:
This dependency is necessary for using semanage, which is needed to do SELinux relabeling.


astencel-sumo (approved):

sumo-drosiek:
> command.go:200: ../../scripts/install.sh: line 348: REQUIRED_COMMANDS: unbound variable

:grimacing:

astencel-sumo (approved):

PR #1310, opened 2023-10-27
Integrate opampprovider into otelcolbuilder

echlebek:
When the --remote-config flag is passed, the otel collector will use it exclusively, bypassing the other configuration providers.

Closes #1287


portertech:
@echlebek I think we should add a few tests around this, via `pkg/scripts_test`.

rnishtala-sumo (commented):
LGTM, do we need doc (readme) changes/additions as part of this PR? or will that be another PR?

rnishtala-sumo (approved):

echlebek:
@rnishtala-sumo I'll file a follow-up PR for docs today

Issue #1309, opened 2023-10-27
[jobreceiver] IO issues with logs entires


c-kruse:
Ran into this a bit in development and now in testing.

We are using go's https://pkg.go.dev/os/exec#Cmd.StdoutPipe, which creates an os.Pipe. On linux that corresponds to a [named pipe](https://man7.org/linux/man-pages/man7/pipe.7.html), which can be especially sensitive to read latency and errors on read after all of the writer fds are closed even when the small read buffer is not empty (resulting in dropped data.) When sensu runs a command it sets exec.Cmd.Stdout to a buffer, and the exec package behind the scenes creates an os.Pipe and starts a goroutine that aggressively reads that pipe into the buffer.

To avoid problems we should add a similar buffer somewhere. The event output handler effectively replicates that buffer [here](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.88.0-sumo-0/pkg/receiver/jobreceiver/output/event/handler.go#L57-L60), but the log entries handler currently tries to process records [in the read loop](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.88.0-sumo-0/pkg/receiver/jobreceiver/output/logentries/handler.go#L33-L49). I hope some form of buffer may make this "good enough", but but because of the flawed nature of this form of IPC we can't really apply effective backpressure like we're accustomed to. I suspect this issue exists in the installed collector script sources as well, and just gets papered over with scala the libraries.

c-kruse:
Side note: I have my doubts around the reality of os pipes, but they might not be the root of the problem.
from Cmd.StdoutPipe:
> Wait will close the pipe after seeing the command exit, so most callers need not close the pipe themselves. It is thus incorrect to call Wait before all reads from the pipe have completed.

May have just goofed on this and need to reorganize how the command package works to accommodate this.

PR #1311, opened 2023-10-27
fix logging in monitoring job receiver

c-kruse:
Ran into a logging bug while observing what is likely a more important bug: #1309.

We should not use the `%w` placeholder unless constructing an error. The output is something like "%w{ fs.PathError {...} }", and doesn't give the reader the actual useful error message.

### DIFF
--- a/pkg/receiver/jobreceiver/output/logentries/handler.go
+++ b/pkg/receiver/jobreceiver/output/logentries/handler.go
@@ -49,7 +49,7 @@ func (h *handler) consume(ctx context.Context, in io.Reader, stream string) {
 		h.writer.Write(ctx, ent)
 	}
 	if err := scanner.Err(); err != nil {
-		h.logger.Errorf("error reading input stream %s: %s", stream, err)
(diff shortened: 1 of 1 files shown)


swiatekm-sumo (approved):

pkg/receiver/jobreceiver/output/logentries/handler.go
```diff
@@ -48,7 +49,7 @@ func (h *handler) consume(ctx context.Context, in io.Reader, stream string) {
 		h.writer.Write(ctx, ent)
 	}
 	if err := scanner.Err(); err != nil {
-		h.logger.Errorf("error reading input stream %s: %w", stream, err)
+		h.logger.Errorf("error reading input stream %s: %s", stream, err)
```
swiatekm-sumo, line 52: Since this is zap, I think the following would be idiomatic:

```suggestion
		h.logger.Errorf("error reading input stream", zap.String("stream", stream), zap.Error(err))
```

PR #1312, opened 2023-10-27
Add upload of installation script logs

ccressent:
As part of [OSC-174](https://sumologic.atlassian.net/browse/OSC-174) and to close [SUMO-227556](https://sumologic.atlassian.net/browse/SUMO-227556), we need to collect the installation script's actions and send that log to Sumologic for further processing.

This PR is the piece that makes the installation script log everything it does into a file and send that file to a Sumo endpoint whenever and for whatever reason the installation script exits.

Note that the log collection endpoint is still set to its staging version while we wait for the production version to be deployed.
[Example log data gathered during a CI run](https://stagdata.long.sumologic.net/ui/#/search/create?id=kwsaGPlMoo6vhspCyftvURRQx41rtOJMnR3SCXjz)

[OSC-174]: https://sumologic.atlassian.net/browse/OSC-174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SUMO-227556]: https://sumologic.atlassian.net/browse/SUMO-227556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccressent:
I'm just realizing: this probably needs a PowerShell equivalent for Windows installations...

scripts/install.sh
```diff
@@ -170,6 +173,12 @@ Supported env variables:
 EOF
 }
 
+function reporter {
+    echo "SUMOLOGIC_INSTALLATION_TOKEN=${SUMOLOGIC_INSTALLATION_TOKEN}" >> $INSTALLATION_LOGFILE
+    curl --silent --location -X POST --data-binary @"${INSTALLATION_LOGFILE}" "${INSTALLATION_LOGFILE_ENDPOINT}"
```
sumo-drosiek, line 178: Shouldn't we secure this endpoint somehow?

scripts/install.sh
```diff
@@ -130,6 +130,9 @@ SYSTEMD_DISABLED=false
 # alternative commands
 TAC="tac"
 
+INSTALLATION_LOGFILE="/tmp/sumologic-otel-collector_installation.log"
+INSTALLATION_LOGFILE_ENDPOINT='https://stag-open-events.sumologic.net/api/v1/collector/installation/logs'
```
sumo-drosiek, line 134: Should this endpoint be a stag endpoint?

scripts/install.sh
ccressent: Good point; the endpoint in question is handled by a different team, but I know there was a security review and after some changes it was deemed acceptable; see: [APPSEC-1878](https://sumologic.atlassian.net/browse/APPSEC-1878)

sumo-drosiek: I wonder if we should add option to disable this functionality. I imagine some customers may want to skip sending those information, or this endpoint may be not accessible within their network

ccressent: Totally. As it stands `curl` would timeout silently, but I agree with you.
How does `--no-installation-telemetry` sound?

I can also make the endpoint URL configurable with a flag, which should be useful for CI runs and stuff.

sumo-drosiek: Maybe `--disable-telemetry` :thinking: I was always not into naming :sweat_smile: 

> I can also make the endpoint URL configurable with a flag, which should be useful for CI runs and stuff.

It would be very nice addition. I like the idea :)

scripts/install.sh
```diff
@@ -170,6 +181,14 @@ Supported env variables:
 EOF
 }
 
+function reporter {
+    if ! $DISABLE_INSTALLATION_TELEMETRY; then
+        echo "SUMOLOGIC_INSTALLATION_TOKEN=${SUMOLOGIC_INSTALLATION_TOKEN}" >> $INSTALLATION_LOGFILE
```
sumo-drosiek, line 186: Could we use some hash to not store token directly inside the file and send it as source header? Right now we are going to get one line with the token and then multiple lines which will be hard to say which installation token are related

https://help.sumologic.com/docs/send-data/hosted-collectors/http-source/logs-metrics/upload-logs/#supported-http-headers

I think `X-Sumo-Name` would be the best choice :thinking:

ccressent: Not 100% sure if `X-Sumo-Name` is the best choice; what about using `X-Sumo-Fields` instead with a value of `installation-token-sha256=<value>`?

ccressent: Whether I use `X-Sumo-Name` or `X-Sumo-Fields` for the headers I send to the endpoint I was provided with, that data doesn't seem to make it into Sumo.

I had to remove compression already because it turned out the `Content-Type` header is not supported by the endpoint. I wonder if whatever they put in place just ignores headers. I'll get in touch with the team in charge of that endpoint.

ccressent: The team has confirmed that those headers are not supported; it's still unclear to me whether or not they'll work on supporting them.

Can you see a different way to accomplish the same idea?

sumo-drosiek: If the endpoint treats all the data like multiline log I think it's fine, Otherwise I would put hash on the beginning of every line, so we do not loose the context

ccressent: The hosted collector (`otel-installation-logs-collector`) is setup with `multilineProcessingEnabled: true`, so I think it's fine indeed.

ccressent:
I suggest we go with what we have so that we can start collecting and analyzing data. We can then alter what we collect as needed.

swiatekm-sumo:
@ccressent I've moved the install scripts to https://github.com/SumoLogic/sumologic-otel-collector-packaging/tree/main/install-script and we're planning to remove them from this repo in #1536. Can you recreate this PR in packaging?

ccressent:
Closing in favour of https://github.com/SumoLogic/sumologic-otel-collector-packaging/pull/64

PR #1321, opened 2023-11-06
Build linux_arm64 and amd64 fips static binaries

c-kruse:
Closes: https://github.com/SumoLogic/sumologic-otel-collector/issues/1325

**Background:**
Our linux_amd64 FIPS binary has a required CGO dependency (boringcrypto) and is currently built as a dynamically linked executable. By contrast, our non-fips linux builds are pure go and are statically linked executables. Because of this our build infrastructure is limited, largely focused on producing a single portable static binary per target. In order to work around the system dependency introduced by dynamically linking c libraries, we have resorted to building our FIPS binary on an amazonlinux:2 container, as it has the lowest glibc version for any distribution we know of that we have agreed is necessary to support.

**This change:**
* Adds a linux_arm64 FIPS build (cross compiled from the amd64 github hosted runners)
* Changes our FIPS build to produce a static executable - linked with the musl libc instead of glibc (which cannot be reliably statically linked) in order to remove the system dependency and increase portability.

**Future notes:**
Once a native aarch64 build environment becomes available, it would be preferable to switch to using the native toolchain from alpine linux rather than building the toolchain ourselves. Let's link an issue to get that cleaned up.


ci/build-fips-action/Dockerfile
```diff
@@ -1,14 +1,58 @@
...
+# golang alpine image ready to build sumologic-otel-collector includes native
+# and cross CC toolchains ready to build static executables. Hard coded for
+# x86_64 for github runners.
+FROM golang:1.20.10-alpine3.18 as FINAL
+
+RUN apk add alpine-sdk
+
+WORKDIR /tmp
+
+ADD x86_64/* /opt
```
swiatekm-sumo, line 53: I'm not a big fan of storing the built binaries in the repo. It looks like we have a reproducible build process for these, so why not just cache them in CI instead?

Also we should rename this folder to something less confusing. `toolchains`, maybe?

swiatekm-sumo (commented):
This looks like the right approach. There are two major missing pieces in this PR from my perspective:

1. Documentation. We should have a README file in ci/build-fips-action, explaining how the build process works and why it's the way it is. Your PR description would be a good start.
2. E2E tests. If you prefix a branch in this repo with `dev-build/`, it'll build the binaries and publish them, at which point it's possible to run E2E tests against them. We should verify we haven't broken anything with this change.

c-kruse:
@swiatekm-sumo Alrighty! Updated to build + cache the toolchains in CI, added some documentation, and hopefully simplified the whole thing a little bit.

This 100% needs ran through our E2E tests!

.github/workflows/workflow-build.yml
```diff
@@ -122,12 +123,31 @@ jobs:
...
+        if: ${{ inputs.fips && contains(inputs.arch_os, 'linux') && (! inputs.only-if-changed || steps.changed-files.outputs.any_changed == 'true') }}
+        run: |
+          CC=$(find /opt/toolchain/bin -type f -name "*-linux-musl-gcc")
+          test "$CC"
+          echo "Using toolchain: $CC"
+          make otelcol-sumo-${{inputs.arch_os}} \
+            FIPS_SUFFIX="-fips" \
+            CGO_ENABLED="1" \
+            CC="$CC" \
+            EXTRA_LDFLAGS="-linkmode external -extldflags '-static'"
```
swiatekm-sumo, line 149: Could we add this as a target to the `otelcolbuilder` Makefile? I'd like this build to be relatively easy to do locally, and currently we're a bit too coupled to Github Actions.

ci/build-fips-action/README.md
```diff
@@ -0,0 +1,50 @@
...
+This would be ideal. Unfortunately the combination of our limited build
+infrastructure (x86_64-only Github hosted runners) and the size and scale of
+our build executable (emulation with qemu takes ~1.5 hours!) means that we are
+forced to cross compile our aarch64 executables. Since alpine does not build a
+suitable cross toolchain, we have opted to build it ourself until we can
+provide a native build environment for our aarch64 builds.
+
+## Building
+
+This is a small warpper around
```
swiatekm-sumo, line 36:
```suggestion
This is a small wrapper around
```

swiatekm-sumo (commented):
Since we're basically getting rid of the custom Github Action, could we just move the whole toolchain directory to `otelcolbuilder`? Then it would be easy to have building the static FIPS binary be just another target in the Makefile there, and consequently allow for easy local builds.

If you'd rather we do that in a separate PR, I'm ok with that as well. @astencel-sumo can you have a look too?

c-kruse:
+1 to all of @swiatekm-sumo's suggestions. I had thought to move this all back into the otlecolbuilder makefile, just ran short on time in my last few days.

Taking a back seat on this one from here on out. :wave:

c-kruse:
@fguimond  closing this because it's getting stale. Obviously feel welcome to co-author and/or re-author anything in here!

PR #1332, opened 2023-11-14
Increasing the default timeout in the sumologic exporter

rnishtala-sumo:
Increasing the default timeout in the sumologic exporter

### DIFF
--- a/pkg/exporter/sumologicexporter/README.md
+++ b/pkg/exporter/sumologicexporter/README.md
@@ -44,7 +44,7 @@ exporters:
     trace_format: {otlp}
 
     # timeout is the timeout for every attempt to send data to the backend,
-    # maximum connection timeout is 55s, default = 5s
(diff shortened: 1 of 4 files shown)


pkg/exporter/sumologicexporter/README.md
```diff
@@ -44,7 +44,7 @@ exporters:
     trace_format: {otlp}
 
     # timeout is the timeout for every attempt to send data to the backend,
-    # maximum connection timeout is 55s, default = 5s
+    # maximum connection timeout is 55s, default = 30s
```
rnishtala-sumo, line 47: do we think this is too high for the default timeout?

swiatekm-sumo: We use this in the Helm Chart and it works fine there.

swiatekm-sumo (approved):

mat-rumian (approved):

PR #1334, opened 2023-11-17
feat(processor/sumologic): add upstream `sumologic` processor and deprecate `sumologic_schema` processor

astencel-sumo:


pkg/processor/sumologicschemaprocessor/README.md
```diff
@@ -1,13 +1,19 @@
 # Sumo Logic Schema Processor
 
-**Stability level**: Beta
+**Stability level**: Deprecated
```
aboguszewski-sumo, line 3: Isn't there a way to mark this in code? IMO we should print a warning if someone uses it.

astencel-sumo: Thanks, very good point, I forgot about this. Let me fix it.

astencel-sumo: Added in https://github.com/SumoLogic/sumologic-otel-collector/pull/1334/commits/cf9c1e80278b2ba160e81766c3ad303bdadee2c2.

aboguszewski-sumo (approved):

PR #1336, opened 2023-11-17
docs: mark deprecated components

astencel-sumo:


swiatekm-sumo (approved):
Looks good to me, stands out visually and I think it's also pretty obvious what the strikethrough and warning signs mean even if you don't read the legend.

Should we also add a point to OT core updates to check if any existing components are deprecated?

PR #1337, opened 2023-11-17
docs: add `active_directory_inv` receiver to component list

astencel-sumo:
We should probably have an automated check that all components from the builder manifest are included in README.


swiatekm-sumo:
> We should probably have an automated check that all components from the builder manifest are included in README.

An automated way to generate the README table from the manifest would be even better.

swiatekm-sumo (approved):

PR #1341, opened 2023-11-21
feat(exporter/syslog)!: replace syslog exporter with syslog exporter from opentelemetry-collector-contrib

kkujawa-sumo:
Fixes https://github.com/SumoLogic/sumologic-otel-collector/issues/1200

PR #1342, opened 2023-11-21
feat(exporter/syslog): deprecate syslog exporter in favor of upstream version of syslog exporter

kkujawa-sumo:
Related to #1341

### DIFF
--- a/.changelog/1342.changed.txt
+++ b/.changelog/1342.changed.txt
@@ -1,0 +1,1 @@
+feat(exporter/syslog): deprecate syslog exporter in favor of upstream version of syslog exporter
--- a/pkg/exporter/syslogexporter/README.md
+++ b/pkg/exporter/syslogexporter/README.md
@@ -1,6 +1,45 @@
 # Syslog Exporter
 
-**Stability level**: Alpha
+**Stability level**: Deprecated
+
+This exporter is deprecated in favor of the [syslog exporter][syslog_exporter_contrib] that lives in the [OpenTelemetry Collector Contrib][contrib_repo] repository.
+The functionality is the same but the configuration is slightly different.
+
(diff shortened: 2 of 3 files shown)


aboguszewski-sumo (commented):
Please also change it in the code:

https://github.com/SumoLogic/sumologic-otel-collector/blob/6ad4db6e4ad9d53d91bc3c58a4dea373003f9766/pkg/exporter/syslogexporter/factory.go#L29

swiatekm-sumo (approved):

aboguszewski-sumo:
Also a comment: maybe in the future we should try to name the components we're pushing into upstream differently than there. This way we won't have such breaking changes.

aboguszewski-sumo (approved):

swiatekm-sumo:
> Also a comment: maybe in the future we should try to name the components we're pushing into upstream differently than there. This way we won't have such breaking changes.

Would that have helped us avoid this deprecation?

kkujawa-sumo:
I think that @aboguszewski-sumo meant #1341

astencel-sumo (changes requested):
How can we suggest using contrib syslog exporter if we don't include it in our distro?

aboguszewski-sumo:
@astencel-sumo we will once https://github.com/SumoLogic/sumologic-otel-collector/pull/1341 is merged. But good point.

kkujawa-sumo:
>How can we suggest using contrib syslog exporter if we don't include it in our distro?

#1341 must be merged first.

astencel-sumo (approved):
So the plan is to replace the Sumo Logic `syslog` exporter with upstream Contrib `syslog` exporter in the next release, `v0.90.0-sumo-0`. This is a breaking change, but only a minor configuration change is needed to upgrade, as described in #1341.

PR #1343, opened 2023-11-21
Add remote management support to the install script

portertech:
Add `--remotely-managed` support to the installation script, for configuring and enabling the OpAMP extension for collector remote management (OTRM).

When `--remotely-managed` is specified, `sumologic.yaml` is manipulated to support OpAMP remote management. The installation script will not create and manage `conf.d/common.yaml`, instead, it will update the relevant fields in `sumologic.yaml`. The installation script will create a separate configuration directory for remote configuration, `opamp.d`, so not to destroy existing local configuration (even tho it won't be in effect).

I ran the following to test these changes on Linux VMs:

```
sudo curl -sL https://raw.githubusercontent.com/SumoLogic/sumologic-otel-collector/otrm/scripts/install.sh | SUMOLOGIC_INSTALLATION_TOKEN="YOURTOKEN" sudo -E bash -s -- --remotely-managed --tag "host.group=default" --tag "deployment.environment=default" --api https://stag-open-events.sumologic.net/ -c otrm
```


portertech:
While the installation script `--remotely-managed` flag is able to configure the OpAMP extension and enable it, I ran into a serious issue. 

We decided to make remote vs local configuration mutually exclusive beyond a single configuration file. This is partially enforced through the newly added `--remote-config` collector flag, which accepts a single YAML file path (e.g. `sumologic.yaml`). Other configuration flags/providers cannot be used at the same time. This new `--remotely-managed` installation script flag manipulates the `sumologic.yaml` base configuration file in order to work with `--remote-config`.

The installation script has shifted to manipulating a "common config" located at `conf.d/common.yaml`. This is where the base API URL, collector fields, etc are being configured. This configuration is still very relevant with remote management, however, it is not being applied as it's not present in the single `--remote-config sumologic.yaml`. I see a few different ways this can be address, however, all cause a further increase in installation script complexity (and could break things).

examples/sumologic.yaml
```diff
@@ -20,6 +20,13 @@ extensions:
   file_storage:
     directory: /var/lib/otelcol-sumo/file_storage
 
+  ## Configuration for OpAMP Extension
+  ## The OpAMP extension remotely manages collector configuration.
+  ## ref: https://github.com/SumoLogic/sumologic-otel-collector/tree/main/pkg/extension/opampextension
+  opamp:
+    remote_configuration_directory: /etc/otelcol-sumo/conf.d
```
portertech, line 27: Currently using the existing `conf.d` directory as the remote configuration directory. The OpAMP agent will overwrite its contents with received remote configuration. Alternatively, we could create a `remote.d` or `opamp.d` directory for this purpose. WDYT?

portertech:
OK, it wasn't as terrible as expected. I was able to tweak and leverage the existing config manipulation functions to do my bidding. Going to write more tests and give it a whirl with stag.

amdprophet (approved):
It may be worth modifying one of the existing tests to check that the `opamp.d` directory isn't created when the flag isn't passed.

What happens if a user specifies the hostmetrics flag? Does that config go into the `opamp.d` directory if the `--remotely-managed` flag is specified so that it can be controlled?

Other than that it looks great! 👍

scripts/install.sh
```diff
@@ -731,27 +774,42 @@ function setup_config() {
...
     # Finish setting permissions after we're done creating config files
     chmod -R 440 "${CONFIG_DIRECTORY}"/*  # all files only readable by the owner
     find "${CONFIG_DIRECTORY}/" -mindepth 1 -type d -exec chmod 550 {} \;  # directories also traversable
 }
 
 function setup_config_darwin() {
-    create_user_config_file "${COMMON_CONFIG_PATH}"
-    add_extension_to_config "${COMMON_CONFIG_PATH}"
-    write_sumologic_extension "${COMMON_CONFIG_PATH}" "${INDENTATION}"
+    config_path="${COMMON_CONFIG_PATH}"
```
sumo-drosiek, line 783: `config_path` should be `local` and `readonly`

scripts/install.sh
```diff
@@ -731,27 +774,42 @@ function setup_config() {
...
+
+    if [[ "${REMOTELY_MANAGED}" == "true" ]]; then
+        config_path="${CONFIG_PATH}"
+    fi
```
sumo-drosiek, line 787: `config_path` should be `local` and `readonly` afterwards

sumo-drosiek (commented):
Looks ok, but I would run in over `shellcheck` if you didn't do it yet

Second thought: Is OpAMP considered customer ready and global available? If not, maybe we should add some warning and disclaimer?

PR #1344, opened 2023-11-23
chore: update Go to 1.21

swiatekm-sumo:
I used more flexible version constraints where I could, so we automatically use the latest minor version.

### DIFF
--- a/.changelog/1344.changed.txt
+++ b/.changelog/1344.changed.txt
@@ -1,0 +1,1 @@
+chore: update Go to `v1.21`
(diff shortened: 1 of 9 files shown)


.changelog/1344.changed.txt
```diff
@@ -0,0 +1 @@
+chore: update Go to 1.21
```
astencel-sumo, line 1: These two changelog entries are different, I don't suppose you want to create two separate ones? 😅 
```suggestion
chore: update Go to `v1.21`
```

astencel-sumo (approved):

PR #1350, opened 2023-11-28
fix(sourceprocessor): fix templating for sourceCategoryPrefix

aboguszewski-sumo:
Fixes #1339 - please see the issue to see the reason why it was reopened.

It turns out that the previous fix was not enough, because the code was affected also in some other places. I also did a small refactor.


pkg/processor/sourceprocessor/source_category_filler.go
```diff
@@ -76,21 +75,30 @@ func (f *sourceCategoryFiller) fill(attributes *pcommon.Map) {
...
+	} else {
+		doesUseAnnotation = true
 	}
 	valueTemplate = prefix + valueTemplate
 
+	if doesUseAnnotation {
+		templateAttributes = extractTemplateAttributes(valueTemplate)
+	} else {
+		templateAttributes = f.templateAttributes
+	}
```
sumo-drosiek, line 100: Is it expected that we skip prefix here?

pkg/processor/sourceprocessor/source_category_filler.go
```diff
@@ -76,21 +75,30 @@ func (f *sourceCategoryFiller) fill(attributes *pcommon.Map) {
...
 
 	prefix := getAnnotationAttributeValue(f.annotationPrefix, sourceCategoryPrefixAnnotation, attributes)
 	if prefix == "" {
 		prefix = f.prefix
+	} else {
+		doesUseAnnotation = true
 	}
 	valueTemplate = prefix + valueTemplate
 
+	if doesUseAnnotation {
```
astencel-sumo, line 96: Why does it matter if we use annotations or not? Should prefix templating only work when using annotations?

pkg/processor/sourceprocessor/source_category_filler_test.go
```diff
@@ -53,14 +53,15 @@ func TestFillWithAnnotations(t *testing.T) {
 	attrs := pcommon.NewMap()
 	attrs.PutStr("k8s.namespace.name", "ns-1")
 	attrs.PutStr("k8s.pod.uid", "123asd")
+	attrs.PutStr("k8s.pod.name", "ABC")
 	attrs.PutStr("k8s.pod.annotation.sumologic.com/sourceCategory", "sc-from-annot-%{k8s.namespace.name}-%{k8s.pod.uid}")
-	attrs.PutStr("k8s.pod.annotation.sumologic.com/sourceCategoryPrefix", "%{k8s.pod.uid}-Prefix:")
+	attrs.PutStr("k8s.pod.annotation.sumologic.com/sourceCategoryPrefix", "%{k8s.pod.name}-%{k8s.pod.uid}-Prefix:")
```
astencel-sumo, line 58: Why is prefix templating tested only with pod annotations? What about `TestFill` and `TestFillWithContainerAnnotations`?

astencel-sumo (changes requested):
Let's get the questions answered before we merge this.

pkg/processor/sourceprocessor/source_category_filler.go
aboguszewski-sumo: `valueTemplate` already contains the prefix, so all templates in the prefix will be affected.

pkg/processor/sourceprocessor/source_category_filler.go
aboguszewski-sumo: This is a good question to the author of the processor.
As I understand it: if we use the annotations, then we override the standard template value or the prefix, and then we can't use the preprocessed template attributes array.

pkg/processor/sourceprocessor/source_category_filler_test.go
aboguszewski-sumo: While this shouldn't matter, it's a good point that it should have more tests. Fixed.

pkg/processor/sourceprocessor/source_category_filler.go
sumo-drosiek: I was asking about `f.templateAttributes`. Does it contain correct prefix as well?

aboguszewski-sumo: Yes, see line `41`

sumo-drosiek (approved):
LGTM
Waiting for @astencel-sumo

aboguszewski-sumo:
@astencel-sumo

pkg/processor/sourceprocessor/source_category_filler_test.go
```diff
@@ -131,4 +134,20 @@ func TestFillWithContainerAnnotations(t *testing.T) {
 
 		assertAttribute(t, attrs, "_sourceCategory", "THIRD_s-c!")
 	})
+
+	t.Run("set source category prefix", func(t *testing.T) {
```
astencel-sumo, line 138: I'm not sure what is the intent of this test? It does not enable container annotations (`cfg.ContainerAnnotations.Enabled = true`), so it only tests that the source category configured in the processor's settings is used.

We should rather have a test that defines what happens if you try to use templates in per-container annotations. I propose something like this:

```go
	t.Run("container annotations do not support templating", func(t *testing.T) {
		cfg := createDefaultConfig().(*Config)
		cfg.ContainerAnnotations.Enabled = true
		cfg.SourceCategory = "my-source-category"
		cfg.SourceCategoryPrefix = "my-prefix"

		attrs := pcommon.NewMap()
		attrs.PutStr("k8s.container.name", "container-name-1")
		attrs.PutStr("k8s.namespace.name", "ns-1")
		attrs.PutStr("k8s.pod.name", "ABC")
		attrs.PutStr("k8s.pod.annotation.sumologic.com/sourceCategory", "sc-from-annot-%{k8s.namespace.name}-%{k8s.pod.uid}")
		attrs.PutStr("k8s.pod.annotation.sumologic.com/sourceCategoryPrefix", "%{k8s.pod.name}-%{k8s.pod.uid}-Prefix:")
		attrs.PutStr("k8s.pod.annotation.sumologic.com/sourceCategoryReplaceDash", "#")
		attrs.PutStr("k8s.pod.annotation.sumologic.com/container-name-1.sourceCategory", "%{k8s.namespace.name}-%{k8s.pod.name}")
		attrs.PutStr("k8s.pod.annotation.sumologic.com/container-name-2.sourceCategory", "another/source-category")
		attrs.PutStr("k8s.container.name", "container-name-1")

		filler := newSourceCategoryFiller(cfg, zap.NewNop())
		filler.fill(&attrs)

		assertAttribute(t, attrs, "_sourceCategory", "%{k8s.namespace.name}-%{k8s.pod.name}")
	})
```

astencel-sumo (approved):
Ideally let's fix that test, other than that 🚀

aboguszewski-sumo:
@astencel-sumo I think that I added that test because there was no test that just checks the prefixing (I had some problems while implementing this and this test helped me).
I just added your test, I guess it should be fine.

PR #1358, opened 2023-12-01
feat(cascadingprocessor): add collector instances configuration

mat-rumian:
PR introducing `collector_instances` config option used by `cascadingfilter` to recalculate global and policy specific `spans_per_second` limits. This functionality simplifies `cascadingfilter` configuration in dynamically scaling environments where multiple instances sharing the same configuration file.


swiatekm-sumo (commented):
Any reason not to just make this a config option? If the user wants to read it from an env variable, they can do that directly in the yaml.

sumo-drosiek (approved):
:violin:

PR #1364, opened 2023-12-04
chore: update golang version to 1.21.4

sumo-drosiek:
Crucial change is

```
powershell -command "Set-ItemProperty -Path HKLM:\System\CurrentControlSet\Control\Lsa\FipsAlgorithmPolicy -Name Enabled -Value \$true"
```

which enables FIPS backend on Windows


swiatekm-sumo (approved):
LGTM, but I'd like @amdprophet to have a look as well as the author.

PR #1367, opened 2023-12-05
Updated install.sh write_installation_token to support installation_token

portertech:
Problem:

Installation with `--remotely-managed` and `--skip-systemd` results in a `install_token` and `installation_token` in `sumologic.yaml`.

For example:

```
extensions:
  ## Configuration for Sumo Logic Extension
  ## Manages registration, heartbeats and authentication to Sumo Logic
  ## ref: https://github.com/SumoLogic/sumologic-otel-collector/tree/main/pkg/extension/sumologicextension
  sumologic:
    api_base_url: https://stag-open-events.sumologic.net/
    install_token: U1VNT2*******dpYy5uZXQ=
    installation_token: ${SUMOLOGIC_INSTALLATION_TOKEN}
    collector_credentials_directory: /var/lib/otelcol-sumo/credentials
```

This pull-request supports updating `installation_token` in the configuration file. It also fixes a bug where `install_token` found its way into the sumologic exporter configuration. I intentionally avoided reworking the deprecated `install_token` and related tests, as I don't want to disrupt existing installations and their upgrade path.


sumo-drosiek:
> I intentionally avoided reworking the deprecated `install_token` and related tests, as I don't want to disrupt existing installations and their upgrade path.

install_token is deprecated from 9 months and not supported at all from 3 months: https://github.com/SumoLogic/sumologic-otel-collector/pull/1225

I think we should clean it up

echlebek:
Looks like the merge resulted in two different otel versions being used, 0.90.0 and 0.90.1

swiatekm-sumo:
> Looks like the merge resulted in two different otel versions being used, 0.90.0 and 0.90.1

This is actually how upstream is right now for `0.90.1`. @sumo-drosiek fixed this in #1364, but in the meantime we can merge in spite of this check not passing. It doesn't actually affect anything.

echlebek (changes requested):
InstallToken should be removed or documentation should be added clarifying how it differs from InstallationToken, IMO.

echlebek:
@sumo-drosiek I've taken a stab at cleaning up the deprecated InstallToken bits. PTAL, thanks!

swiatekm-sumo (approved):
LGTM, but I'd like @sumo-drosiek to take a look as well.

sumo-drosiek (approved):
LGTM

echlebek (approved):

PR #1370, opened 2023-12-06
fix(install): fail install if Otelcol is not running on darwin

### LINKED ISSUE #1361: Add more checks to MacOS install script
Implement the following checks: https://github.com/SumoLogic/sumologic-documentation/pull/3346

swiatekm:
Check if the launchd task is running after installation, same as we do for systemd. Add an error message to both of these cases.

I think this also resolves #1361 - @astencel-sumo @sumo-drosiek can you have a look?

### DIFF
--- a/pkg/scripts_test/install_darwin_test.go
+++ b/pkg/scripts_test/install_darwin_test.go
@@ -81,6 +81,7 @@ func TestInstallScriptDarwin(t *testing.T) {
 				checkLaunchdConfigCreated,
 				checkHomeDirectoryCreated,
 			},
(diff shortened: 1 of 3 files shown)


swiatekm-sumo:
> LGTM, but tests are failing

Unfortunately, it's not that easy to check if Otelcol is failing on MacOS, as the state can briefly be `running` even if it exits with an error, and launchctl doesn't have any simple way of checking for this. I'm going to try looking at the exit code instead.

sumo-drosiek (approved):

Issue #1325, opened 2023-11-08
Add linux/arm64 FIPS-capable build


c-kruse:
Currently we only build and publish FIPS capable binaries for linux/amd64. The same cryptographic module used by our amd64 builds has been validated for aarch64 as well and is included in the go standard library. See our [FIPS Documentation](https://github.com/SumoLogic/sumologic-otel-collector/blob/main/docs/fips.md) as it applies to obtaining and running the collector in FIPS-approved mode.

PR #1381, opened 2023-12-07
Dev build/add linux arm64 fips static

portertech:
Original pull-request: https://github.com/SumoLogic/sumologic-otel-collector/pull/1321

Closes: https://github.com/SumoLogic/sumologic-otel-collector/issues/1325

Background:
Our linux_amd64 FIPS binary has a required CGO dependency (boringcrypto) and is currently built as a dynamically linked executable. By contrast, our non-fips linux builds are pure go and are statically linked executables. Because of this our build infrastructure is limited, largely focused on producing a single portable static binary per target. In order to work around the system dependency introduced by dynamically linking c libraries, we have resorted to building our FIPS binary on an amazonlinux:2 container, as it has the lowest glibc version for any distribution we know of that we have agreed is necessary to support.

This change:

- Adds a linux_arm64 FIPS build (cross compiled from the amd64 github hosted runners)
- Changes our FIPS build to produce a static executable - linked with the musl libc instead of glibc (which cannot be reliably statically linked) in order to remove the system dependency and increase portability.

Future notes:
Once a native aarch64 build environment becomes available, it would be preferable to switch to using the native toolchain from alpine linux rather than building the toolchain ourselves. Let's link an issue to get that cleaned up.


portertech:
We need to validate fips operation on Amazon Linux.

sumo-drosiek (approved):
Looks fine for me

Issue #1387, opened 2023-12-13
requirefips build tag enabled in non-FIPS Windows build of v0.90.1-sumo-0


eslavich:
When I install the collector from the following URL:

```
https://github.com/SumoLogic/sumologic-otel-collector/releases/download/v0.90.1-sumo-0/otelcol-sumo_0.90.1.0_en-US.x64.msi
```

the collector service fails to start with this output:

```
panic: cngcrypto: not in FIPS mode

goroutine 1 [running]:
crypto/internal/backend.init.0()
        crypto/internal/backend/cng_windows.go:34 +0x85
```

when I try again with GOFIPS=0, the output changes to this:

```
panic: the 'requirefips' build tag is enabled, but it conflicts with the detected env variable GOFIPS=0 which would disable FIPS mode

goroutine 1 [running]:
crypto/internal/backend.envGoFIPS()
        crypto/internal/backend/common.go:37 +0xf4
crypto/internal/backend.init.0()
        crypto/internal/backend/cng_windows.go:28 +0x17
```

does that mean the non-FIPS build was inadvertently compiled with requirefips enabled?  Or am I doing something wrong?

sumo-drosiek:
> I'm not fluent in Makefile, but this line appears to be setting requirefips for all Windows builds:
> 
> https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.90.1-sumo-0/otelcolbuilder/Makefile#L58

It shouldn't as this is more readable version:

```
ifeq ($(FIPS_SUFFIX),"")
  GOEXPERIMENT=""
  LDFLAGS="-s -w"
  GOFIPS=1
# We want to use boringcrypto for fips setup
else
  ifeq ($(shell go env GOOS),windows)
    GOEXPERIMENT=systemcrypto
    GOTAGS := $(GOTAGS) requirefips
    LDFLAGS="-w"
  else
    GOEXPERIMENT=boringcrypto
    LDFLAGS="-w"
  endif
endif
```

> does that mean the non-FIPS build was inadvertently compiled with requirefips enabled? Or am I doing something wrong?

I figured out that we always set `GOFIPS=1`, which may be the reason.

As quick solution I would recommend to run with `GOFIPS=0`
Long-term solution: https://github.com/SumoLogic/sumologic-otel-collector/pull/1388

sumo-drosiek:
I tested both binaries (fips disabled on windows)

```
[...]

PR #1388, opened 2023-12-14
fix: remove GOFIPS from build as this is runtime check

sumo-drosiek:


ccressent:
Do you think that will address #1387?
I'm curious because it seems to me like the issue is also about the `requirefips` build flag?

swiatekm-sumo:
Does GOFIPS even do anything at build time?

ccressent:
I don't believe so, I think this is a runtime configuration only for the Microsoft version of the Go toolchain: https://github.com/microsoft/go/blob/microsoft/main/eng/doc/fips/README.md#usage-runtime

ccressent:
I think this PR can be abandoned?
It was opened while @sumo-drosiek was looking at #1387, but this has been handled by other means.

swiatekm-sumo:
@sumo-drosiek can you just delete GOFIPS from the Makefile, if it doesn't do anything?

otelcolbuilder/Makefile
```diff
@@ -50,9 +50,9 @@ endif
 ifeq ($(FIPS_SUFFIX),"")
 GOEXPERIMENT=""
 LDFLAGS="-s -w"
-GOFIPS=1
 # We want to use boringcrypto for fips setup
 else
+GOFIPS=1
```
sumo-drosiek, line 55:
```suggestion
```

otelcolbuilder/Makefile
```diff
@@ -53,9 +53,7 @@ endif
 # We don't want to strip symbols for FIPS builds, as they're useful for verifying the build
 ifeq ($(FIPS_SUFFIX),"")
 GOEXPERIMENT=""
-LDFLAGS+= -s -w
-GOFIPS=1
-
+LDFLAGS="-s -w"
```
sumo-drosiek, line 56:
```suggestion
LDFLAGS+= -s -w
```

swiatekm-sumo (approved):

PR #1399, opened 2023-12-21
Use sumologic extension base URL to determine the appropriate OpAMP endpoint

portertech:
The sumologic extension may update/alter the API base URL used, following the collector registration process. This base URL redirection is tied to the installation token which the OpAMP extension isn't aware of. Update the OpAMP extension to leverage the sumologic extension's base URL to determine the appropriate OpAMP endpoint (i.e. region).

### DIFF
--- a/pkg/extension/opampextension/go.mod
+++ b/pkg/extension/opampextension/go.mod
@@ -3,7 +3,7 @@ module github.com/open-telemetry/opentelemetry-collector-contrib/extension/opamp
 go 1.20
 
 require (
-	github.com/SumoLogic/sumologic-otel-collector/pkg/extension/sumologicextension v0.91.0-sumo-0-rc.0
+	github.com/SumoLogic/sumologic-otel-collector/pkg/extension/sumologicextension v0.91.0-sumo-0
 	github.com/google/uuid v1.3.1
 	github.com/knadh/koanf/parsers/yaml v0.1.0
 	github.com/knadh/koanf/providers/rawbytes v0.1.0
--- a/pkg/extension/opampextension/go.mod
+++ b/pkg/extension/opampextension/go.mod
@@ -3,7 +3,7 @@ module github.com/open-telemetry/opentelemetry-collector-contrib/extension/opamp
 go 1.20
 
 require (
-	github.com/SumoLogic/sumologic-otel-collector/pkg/extension/sumologicextension v0.77.0-sumo-0
+	github.com/SumoLogic/sumologic-otel-collector/pkg/extension/sumologicextension v0.91.0-sumo-0-rc.0
 	github.com/google/uuid v1.3.1
 	github.com/knadh/koanf/parsers/yaml v0.1.0
 	github.com/knadh/koanf/providers/rawbytes v0.1.0
--- a/pkg/extension/opampextension/opamp_agent.go
+++ b/pkg/extension/opampextension/opamp_agent.go
@@ -240,8 +240,6 @@ func (o *opampAgent) setEndpoint() {
 	e = strings.Replace(e, "open-events", "opamp-events", 1)
 	e = strings.Replace(e, "open-collectors", "opamp-collectors", 1)
 	o.endpoint = strings.TrimRight(e, "/") + "/v1/opamp"
-
-	return
 }
 
 func newOpampAgent(cfg *Config, logger *zap.Logger, build component.BuildInfo, res pcommon.Resource) (*opampAgent, error) {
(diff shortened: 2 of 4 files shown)


echlebek:
I feel this approach is a bit hacky and could fail if the URL structure of the base URL changes. However, given the transitional nature of the landscape here, perhaps it's fine. What do reviewers think?

sumo-drosiek:
This is very hacky and I wouldn't say reliable as long-term solution. Can we ensure that this will be done properly in the future? The address should be returned by the API backend

pkg/extension/opampextension/opamp_agent_test.go
```diff
@@ -213,3 +214,53 @@ func TestReload(t *testing.T) {
...
+func TestHackSetEndpoint(t *testing.T) {
+	tests := []struct {
+		name         string
+		url          string
+		wantEndpoint string
+	}{
+		{
+			name:         "empty url defaults to config endpoint",
+			wantEndpoint: "https://example.com",
+		},
```
echlebek, line 231: I am not sure that this test case indicates valid behaviour. It's a reproduction of the business logic that was encoded in the draft PR, and may need to be changed.

echlebek:
@swiatekm-sumo if you could take a look, I would appreciate it, thank you! :pray:

pkg/extension/opampextension/opamp_agent_test.go
```diff
@@ -213,3 +214,53 @@ func TestReload(t *testing.T) {
...
+		{
+			name:         "url variant b",
+			url:          "https://sumo-open-collectors.example.com",
+			wantEndpoint: "wss://sumo-opamp-collectors.example.com/v1/opamp",
+		},
+		{
+			name:         "url variant c",
+			url:          "https://example.com",
+			wantEndpoint: "wss://example.com/v1/opamp",
+		},
```
swiatekm-sumo, line 246: Can we add at least one test case for actual API urls this will convert?

pkg/extension/opampextension/opamp_agent.go
```diff
@@ -223,6 +237,33 @@ func (o *opampAgent) watchCredentials(ctx context.Context, callback func(ctx con
...
+	}
+
+	u.Scheme = "wss"
+	u.Path = path.Join(u.Path, "/v1/opamp")
+
+	// These replacements are specific to Sumo Logic's current domain naming,
+	// and are made provisionally for the OTRM beta. In the future, the backend
+	// will inform the agent of the correct OpAMP URL to use.
+	u.Host = strings.Replace(u.Host, "open-events", "opamp-events", 1)
+	u.Host = strings.Replace(u.Host, "open-collectors", "opamp-collectors", 1)
```
swiatekm-sumo, line 260: I thought we'd agreed to parse the Sumo deployment name out of the domain and explicitly construct a new one. Any reason to do a string replace here?

pkg/extension/opampextension/opamp_agent_test.go
echlebek: Sure, not a problem. :+1:

pkg/extension/opampextension/opamp_agent.go
echlebek: Maybe it's my naivete regarding Sumo deployments, but it seems even more error prone to me. How do I know which deployments correspond to open-events, and which correspond to open-collectors (and correspondingly, opamp-events and opamp-collectors). This knowledge is not encoded anywhere in the codebase, unless I'm mistaken, and I don't know how it would be determined either.

swiatekm-sumo: I was hoping to do `https://long-open-events.sumologic.net` -> deployment = `long` -> construct `wss://{{ deployment }}-opamp-events.sumologic.net/v1/opamp`. But looking at it now, these domains are actually different in production environments, so this might be even more complicated.

I'm fine with it if we add test cases for all the possible forms of this url. For example, on prod it's just `open-collectors.sumologic.com`.

pkg/extension/opampextension/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/open-telemetry/opentelemetry-collector-contrib/extension/opamp
 go 1.20
 
 require (
-	github.com/SumoLogic/sumologic-otel-collector/pkg/extension/sumologicextension v0.77.0-sumo-0
+	github.com/SumoLogic/sumologic-otel-collector/pkg/extension/sumologicextension v0.91.0-sumo-0-rc.0
```
astencel-sumo, line 6:
```suggestion
	github.com/SumoLogic/sumologic-otel-collector/pkg/extension/sumologicextension v0.91.0-sumo-0
```

astencel-sumo (approved):

PR #1401, opened 2023-12-22
fix(k8stagger)!: change default pod id attribute name to k8s.pod.uid

### LINKED ISSUE #1379: [k8sprocessor] including `k8s.pod.uid` in `extract.metadata` creates `k8s.pod.id` field, but it should create `k8s.pod.uid`
Version: `v0.89.0-sumo-2`

https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.89.0-sumo-2/pkg/processor/k8sprocessor/kube/kube.go#L43

aboguszewski-sumo:
Fixed #1379 
The previous name was k8s.pod.id.

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -1,7 +1,10 @@
 # Upgrading
 
 - [Upgrading](#upgrading)
-  - [Upgrading to v0.90.1-sumo-0](#upgrading-to-v0910-sumo-0)
+  - [Upgrading to v0.91.0-sumo-0](#upgrading-to-v0910-sumo-0)
+    - [Sumo Logic Schema processor replaced with Sumo Logic processor](#sumo-logic-schema-processor-replaced-with-sumo-logic-processor)
+    - [`k8s_tagger` processor: default name of podID attribute has changed](#k8s_tagger-processor-default-name-of-podid-attribute-has-changed)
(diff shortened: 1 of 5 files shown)


sumo-drosiek (commented):
We do need a migration for customers who would like to keep old behavior

aboguszewski-sumo:
> We do need a migration for customers who would like to keep old behavior
A migration guide or a migration script/config option?

sumo-drosiek:
> > We do need a migration for customers who would like to keep old behavior
> 
> A migration guide or a migration script/config option?

Migration guide should be enough. Not sure if we need a migration process (deprecation and so on) for this particular change. cc: @astencel-sumo

astencel-sumo:
I think an upgrade doc in the Upgrading.md file should be fine, explaining what's changed and how to revert to the previous behavior if the user needs that.

docs/upgrading.md
```diff
@@ -1,7 +1,8 @@
 # Upgrading
 
 - [Upgrading](#upgrading)
-  - [Upgrading to v0.90.1-sumo-0](#upgrading-to-v0910-sumo-0)
+  - [Upgrading to v0.91.0-sumo-0](#upgrading-to-v0910-sumo-0)
```
sumo-drosiek, line 4: missing sub point

sumo-drosiek (approved):

PR #1416, opened 2024-01-10
Use musl toolchain to build static fips release binary

amdprophet:
Updates the release CI job to use the musl toolchain to build FIPS binaries. This should fix an issue where users with older GLIBC versions could not run the FIPS distribution of otelcol-sumo.

A test run can be found on this fork: https://github.com/sensu/sumologic-otel-collector/actions/runs/7468333702/job/20323564335

Running ldd on the resulting binary produces the following output:

```
[root@rhel-7-9 ~]# ldd otelcol-sumo-0.90.1-sumo-10000-fips-linux_amd64 
        not a dynamic executable
```

### DIFF
--- a/.github/workflows/release_builds.yml
+++ b/.github/workflows/release_builds.yml
@@ -229,7 +229,9 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        arch_os: [ 'linux_amd64' ]
+        arch_os:
+          - 'linux_amd64'
+          - 'linux_arm64'
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


sumo-drosiek (approved):

swiatekm-sumo:
I think we should add a changelog entry for this. It is actually a new feature, that the FIPS binary can now be used irrespective of the host system's libc.

.github/workflows/release_builds.yml
```diff
@@ -279,7 +292,7 @@ jobs:
       - name: Store binary as action artifact
         uses: actions/upload-artifact@v4
         with:
-          name: ${{matrix.arch_os}}_fips
+          name: "${{matrix.arch_os}}-fips"
```
astencel-sumo, line 295: Not sure why this is changed, but I think if you change this then you also need to change the artifact name in `Build container` job's `Download binary action artifact from build phase (FIPS)` step below.

https://github.com/SumoLogic/sumologic-otel-collector/pull/1416/files#diff-e048a1bb70d286fe06c2fdaa5a9ad1890a305dcea7eb3894e82599cd4c7e0101R352

PR #1419, opened 2024-01-10
Fix non-FIPS release MSI files

ccressent:
While packaging Windows MSI files, the release workflow was mistakenly using the FIPS binary in all cases. This lead to issues for people using the non-FIPS MSI file or the PowerShell installation script, e.g. #1387.

This change makes the workflow grab the right executable in the right case (FIPS vs non-FIPS) and should fix the issue for releases going forward.


swiatekm-sumo (approved):
Imo we should add a changelog entry for this, since it's a user-facing bug.

Issue #1414, opened 2024-01-09
k8s_tagger: Service metadata is not removed from cache after Pods are deleted


swiatekm-sumo:
We use Endpoints resources to track Service to Pod relationships and add service metadata to resource attributes. This metadata is only removed from cache when the respective Endpoints resource is deleted. However, this resource has a lifetime equal to the lifetime of the Service resource itself, and removing Pods from the Service is accomplished by updating the Endpoints resource. We don't delete any data from our cache on Endpoints updates, and so the cache entries for deleted Pods stay in the cache potentially indefinitely.

As a result, we have a slow memory leak that is only really apparent in large clusters over long time periods.

PR #1422, opened 2024-01-11
feat!(k8sprocessor): use Endpointslices for Service metadata

swiatekm-sumo:
Use the EndpointSlices resources instead of the Endpoints resource for tracking Pod to Service relationships. This is a newer API which is more efficient - EndpointSlices are smaller, so we don't need to reprocess all of a Service's endpoints whenever one of them changes, which is currently the case.

I've also added a transform function that drops any data we don't use 

This improves memory utilization significantly for large clusters. Aside from unit tests I've also run our Helm Chart integration tests on this change, as well as putting it in our staging environment for a day.

This is a breaking change because it requires new RBAC permissions for k8sprocessor to add Servide metadata.

This is part of the fix for #1414, for now I've added a failing test that is currently skipped.

### DIFF
--- a/pkg/processor/k8sprocessor/kube/owner.go
+++ b/pkg/processor/k8sprocessor/kube/owner.go
@@ -21,6 +21,7 @@ import (
 
 	"go.uber.org/zap"
 	api_v1 "k8s.io/api/core/v1"
+	discovery_v1 "k8s.io/api/discovery/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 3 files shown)


sumo-drosiek (approved):

kkujawa-sumo (approved):

pkg/processor/k8sprocessor/kube/owner.go
```diff
@@ -338,7 +363,13 @@ func (op *OwnerCache) addOwnerInformer(
 		}),
 	})
 	if err != nil {
-		op.logger.Error("error adding event handler to namespace informer", zap.Error(err))
+		op.logger.Error("error adding event handler to owner informer", zap.Error(err), zap.String("kind", kind))
+	}
+
+	if transformFunc != nil {
+		if err = informer.SetTransform(transformFunc); err != nil {
```
rnishtala-sumo, line 370: Is the transform function going to remove stale cache entries?

swiatekm-sumo: No, this doesn't fix the memory leak at all. I added a test for it, but it's skipped for now.

PR #1425, opened 2024-01-11
fix(k8sprocessor): Pod Service cache invalidation

### LINKED ISSUE #1414: k8s_tagger: Service metadata is not removed from cache after Pods are deleted

swiatekm-sumo:
Fixes #1414 by separately handling update events for EndpointSlices. The assumption behind the fix is that every Pod exists exactly once in one of the EndpointSlices associated with a Service, and we can never get the updates in the wrong order if it's re-added to the Service after having been removed from it.

A more robust solution, where we periodically recompute the whole Pod to Service map based on EndpointSlice data, is possible, but more expensive and more complex.

I've also replaced some homegrown slice manipulation code with the `slices` module, that's part of the standard library in Go 1.21.


pkg/processor/k8sprocessor/kube/owner_test.go
```diff
@@ -579,8 +578,20 @@ func Test_OwnerProvider_GetServices(t *testing.T) {
...
+		}, 5*time.Second, 10*time.Millisecond)
 
-			return len(services) == 1
+		// update back to the original value
+		_, err = c.DiscoveryV1().EndpointSlices(namespace).
+			Update(context.Background(), endpointSlice2, metav1.UpdateOptions{})
+		require.NoError(t, err)
+		assert.Eventually(t, func() bool {
+			services := op.GetServices(pod.Name)
+			if len(services) != 2 {
```
rnishtala-sumo, line 590: Why are we checking is the length is not 2 here?

swiatekm-sumo: It should be 2, we continue waiting while it's not 2. And it should be 2 because we updated back to the original state, where the EndpointSlice contains the Pod.

pkg/processor/k8sprocessor/kube/owner.go
```diff
@@ -505,12 +504,43 @@ func (op *OwnerCache) genericEndpointSliceOp(obj interface{}, serviceFunc func(p
...
+		}
+	}
+
+	// for each Pod name which was in the old slice, but not in the new slice, schedule a delete
+	for _, endpoint := range oldEndpointSlice.Endpoints {
+		if endpoint.TargetRef != nil && endpoint.TargetRef.Kind == "Pod" {
+			podName := endpoint.TargetRef.Name
+			if slices.Index(newPodNames, podName) == -1 {
+				// not a deferred delete, as this is a dynamic property which can change often
+				op.deleteServiceFromPod(podName, serviceName)
```
rnishtala-sumo, line 542: Is a call to `deleteServiceFromPod` going to ensure that our cache doesn't have any stale entries? So my understanding here is that whenever any EndpointSlice for a service is updated this method gets called with the params as mentioned here - https://pkg.go.dev/k8s.io/client-go@v0.29.0/tools/cache#ResourceEventHandlerFuncs.OnUpdate

swiatekm-sumo: Yes, that's the idea. An update to an EndpointSlice can mean that some Pods were removed from the Service, and this is what we do here.

rnishtala-sumo (approved):

PR #1426, opened 2024-01-15
feat(k8sprocessor): add new owner metrics

swiatekm-sumo:
Add new and update existing metrics for the k8sprocessor:

* the `other` metrics counting events for non-Pod resources now have a `kind` dimension indicating the resource
* add a `owner_table_size` metric
* add a `service_table_size` for the size of the Pod to Services map

### DIFF
--- a/.changelog/1426.added.txt
+++ b/.changelog/1426.added.txt
@@ -1,0 +1,1 @@
+feat(k8sprocessor): add new owner metrics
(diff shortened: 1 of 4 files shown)


aboguszewski-sumo (approved):
We should have these metrics documented somewhere - but that's probably a problem that affects other components too

swiatekm-sumo:
> We should have these metrics documented somewhere - but that's probably a problem that affects other components too

I can add document all the metrics in a separate PR, but I don't think any components do this right now. It'd be nice to have a standard format, maybe `metadata.yaml`?

PR #1432, opened 2024-01-22
Reuse compression configuration from the confighttp helper

rnishtala-sumo:
- Deprecate compress_encoding and document using compression instead
- Remove compress_encoding and all of our own compression code for HTTP
- Reuse compression configuration from the confighttp helper


pkg/exporter/sumologicexporter/sender_test.go
```diff
@@ -1032,82 +1026,6 @@ func TestInvalidPipeline(t *testing.T) {
 	assert.EqualError(t, err, `unknown pipeline type: invalidPipeline`)
 }
 
-func TestSendCompressGzip(t *testing.T) {
```
rnishtala-sumo, line 1035: The sender module no longer has any compression logic, instead the compression encoding is now set in the `HttpClientSettings`

.changelog/1432.changed.txt
```diff
@@ -0,0 +1 @@
+Remove compress_encoding and all of our own compression code in favor of using the confighttp hhelper
```
sumo-drosiek, line 1:
```suggestion
feat: remove compress_encoding and all of our own compression code in favor of using the confighttp helper
```

pkg/exporter/sumologicexporter/README.md
```diff
@@ -17,7 +17,10 @@ exporters:
     # if sumologicextension is not being used, the endpoint is required
     endpoint: <HTTP_Source_URL>
     # Compression encoding format, empty string means no compression, default = gzip
+    # NOTICE: compress_encoding is deprecated and will be removed in an upcoming release
     compress_encoding: {gzip, deflate, ""}
+    # Compression encoding format, empty string means no compression, default = gzip
+    compression: {gzip, zstd, deflate, ""}
```
sumo-drosiek, line 23: Do we need to deprecate it? Does it make sense to reuse it. It would just add zstd to `compress_encoding`

sumo-drosiek: Ok, I see it is easier that way, and actually we already support `compress` from confighttp helper

pkg/exporter/sumologicexporter/exporter_test.go
```diff
@@ -120,10 +120,9 @@ func prepareExporterTest(t *testing.T, cfg *Config, cb []func(w http.ResponseWri
 
 func TestInitExporter(t *testing.T) {
 	_, err := initExporter(&Config{
-		LogFormat:        "json",
-		MetricFormat:     "otlp",
-		CompressEncoding: "gzip",
-		TraceFormat:      "otlp",
+		LogFormat:    "json",
+		MetricFormat: "otlp",
+		TraceFormat:  "otlp",
```
sumo-drosiek, line 125: Shouldn't we use `Compress`?

sumo-drosiek (commented):
I cannot find place where old setting is respected. Also what is behavior if both settings are set? I believe `compress_encoding` should be preferred and we should add some log to inform about it

pkg/exporter/sumologicexporter/config.go
```diff
@@ -106,6 +109,16 @@ func CreateDefaultHTTPClientSettings() confighttp.HTTPClientSettings {
 
 func (cfg *Config) Validate() error {
 
+	switch cfg.HTTPClientSettings.Compression {
```
sumo-drosiek, line 112: It should use `CompressEncoding` value, so we respect actual customer configuration

rnishtala-sumo: This is the new validation for the `Compress` field I'll keep the validation for the deprecated field `CompressEncoding` as well.

.changelog/1432.changed.txt
rnishtala-sumo: feat: Deprecate compress_encoding and remove all of our own compression code in favor of using the confighttp helper

pkg/exporter/sumologicexporter/exporter_test.go
rnishtala-sumo: Yes I added this back.

pkg/exporter/sumologicexporter/config.go
```diff
@@ -106,6 +109,25 @@ func CreateDefaultHTTPClientSettings() confighttp.HTTPClientSettings {
...
+
+	default:
+		return fmt.Errorf("invalid compression encoding type: %v", cfg.HTTPClientSettings.Compression)
+	}
+
+	switch cfg.HTTPClientSettings.Compression {
+	case GZIPCompression:
+	case NoCompression:
+	case DeflateCompression:
+	case ZSTDCompression:
```
swiatekm-sumo, line 125: Since we know Zstd is currently bugged, we should prevent users from setting it until the upstream issue is resolved.

pkg/exporter/sumologicexporter/config.go
```diff
@@ -198,11 +200,13 @@ const (
...
-	GZIPCompression CompressEncodingType = "gzip"
+	GZIPCompression configcompression.CompressionType = "gzip"
+	// ZSTDCompression represents compress_encoding: zstd
+	ZSTDCompression configcompression.CompressionType = "zstd"
 	// DeflateCompression represents compress_encoding: deflate
-	DeflateCompression CompressEncodingType = "deflate"
+	DeflateCompression configcompression.CompressionType = "deflate"
 	// NoCompression represents disabled compression
-	NoCompression CompressEncodingType = ""
+	NoCompression configcompression.CompressionType = ""
```
swiatekm-sumo, line 209: Is there any reason to define these constants vs just using the ones from the core package?

rnishtala-sumo: ah yes, I'll just use the upstream ones.

pkg/exporter/sumologicexporter/config.go
rnishtala-sumo: A thought on this, could we allow the user to configure zstd with the new `Compress` config option and link to the known issue with memory usage? `CompressEncoding` wouldn't support this option

rnishtala-sumo:
@sumo-drosiek if `CompressEncoding` is set we just set `Compression` in the HTTPClientSettings to that value, that way we still support all the original compression types
https://github.com/SumoLogic/sumologic-otel-collector/blob/zstd-support/pkg/exporter/sumologicexporter/exporter.go#L346

pkg/exporter/sumologicexporter/config.go
swiatekm-sumo: We can just return an error. Why would that be a problem?

rnishtala-sumo: The fix for the upstream issue may take a while, and I was wondering if we could include zstd support in the next release and mark it experimental, with a link to the bug. That way we may get some customer feedback as well while I work on the upstream issue. `gzip` will remain as the default and if a customer runs into memory issues with `zstd` they can always go back to the default (gzip).

swiatekm-sumo: I guess if it's just for testing, we can include it. But we should print a warning if the user picks it.

pkg/exporter/sumologicexporter/exporter.go
```diff
@@ -386,6 +343,10 @@ func (se *sumologicexporter) configure(ctx context.Context) error {
 		foundSumoExt bool
 	)
 
+	if se.config.CompressEncoding != NoCompression {
```
swiatekm-sumo, line 346: The default for compress_encoding is `gzip`, so this will happen if the user doesn't set anything, which I don't think is what we want? We should check for the default here.

pkg/exporter/sumologicexporter/README.md
```diff
@@ -17,7 +17,10 @@ exporters:
     # if sumologicextension is not being used, the endpoint is required
     endpoint: <HTTP_Source_URL>
     # Compression encoding format, empty string means no compression, default = gzip
+    # NOTICE: compress_encoding is deprecated and will be removed in an upcoming release
```
astencel-sumo, line 20: Deprecation notice should mention what to use instead.

.changelog/1432.changed.txt
```diff
@@ -0,0 +1 @@
+feat: Deprecate compress_encoding and remove all of our own compression code in favor of using the confighttp helper
```
astencel-sumo, line 1:
```suggestion
feat(sumologicexporter): Deprecate compress_encoding and remove all of our own compression code in favor of using the confighttp helper
```

pkg/receiver/jobreceiver/README.md
```diff
@@ -12,20 +12,20 @@ of downloading runtime assets necessary to run a particular monitoring job.
 
 ## Configuration
 
-| Configuration | Default | Description
-| ------------ | ------------ | ------------
-| exec | required | A `exec` configuration block. See details [below](#execution-configuration)
-| schedule | required | A `schedule` configuration block. See details [below](#schedule-configuration)
-| output | | An `output` configuration block. See details [below](#output-configuration)
+| Configuration | Default | Description|
```
astencel-sumo, line 15: I don't think this file should be modified as part of this PR?

pkg/exporter/sumologicexporter/exporter.go
rnishtala-sumo: actually `compress_encoding` isn't set by default anymore. Instead, `Compression` is now set by default to be `gzip` here - https://github.com/SumoLogic/sumologic-otel-collector/blob/zstd-support/pkg/exporter/sumologicexporter/config.go#L103

pkg/receiver/jobreceiver/README.md
rnishtala-sumo: It was throwing lint errors in the build, so I just added a `|` at the end nothing else.

pkg/exporter/sumologicexporter/config.go
rnishtala-sumo: ok I'll into this. At present if `CompressEncoding` is set it takes precedence over `Compression` if both are set.

pkg/exporter/sumologicexporter/exporter.go
rnishtala-sumo: so, the default is now driven by `Compression` and if `CompressEncoding` is set it takes precedence over `Compression`.

rnishtala-sumo: @swiatekm-sumo does this make sense?, I think the current condition is valid, because `CompressEncoding` isn't set by default anymore.

pkg/exporter/sumologicexporter/config.go
rnishtala-sumo: I've added a condition for this - https://github.com/SumoLogic/sumologic-otel-collector/blob/zstd-support/pkg/exporter/sumologicexporter/config.go#L131C84-L131C107

pkg/exporter/sumologicexporter/exporter.go
swiatekm-sumo: Ok, makes sense.

swiatekm-sumo (approved):

PR #1434, opened 2024-02-05
Add staticcheck github action

echlebek:
The golangci-lint action takes around 10 minutes to complete. Testing to see if the staticcheck action can do it in less time, as many of the golangci-lint linters do redundant work.


echlebek:
The default action can't be used as the project doesn't expose a single root-level module, I'll continue to investigate to see if a workaround exists.

PR #1440, opened 2024-02-08
feat: build UBI-based image

aboguszewski-sumo:
This is more like a PoC than ready PR:
- only linux/amd64 was added
- dev builds are only to confirm that it builds successfully: IMO we should only have this enabled for releases
- `LATEST_TAG_FIPS_SUFFIX` was used as a temporal solution: imo we should rename it just to `LATEST_TAG_SUFFIX` and use it as usual


swiatekm-sumo:
> dev builds are only to confirm that it builds successfully: IMO we should only have this enabled for releases

I disagree, anything we build for releases should also be built for dev. Otherwise we're never going to catch breakage.

The approach in this PR looks reasonable to me in general.

sumo-drosiek:
>     * `LATEST_TAG_FIPS_SUFFIX` was used as a temporal solution: imo we should rename it just to `LATEST_TAG_SUFFIX` and use it as usual

Can we rename it in this PR, or another one and merge it before this PR?

ci/build-push-multiplatform.sh
```diff
@@ -82,7 +82,7 @@ function build_push() {
     esac
 
     local TAG
-    readonly TAG="${REPO_URL}:${BUILD_TAG}-${BUILD_ARCH}"
+    readonly TAG="${REPO_URL}:${BUILD_TAG}${LATEST_TAG_FIPS_SUFFIX}-${BUILD_ARCH}"
```
sumo-drosiek, line 85: Does it affect current builds (fips especially)?

astencel-sumo:
> > dev builds are only to confirm that it builds successfully: IMO we should only have this enabled for releases
> 
> I disagree, anything we build for releases should also be built for dev. Otherwise we're never going to catch breakage.

I agree we should build, but we probably don't need to push the non-release UBI images to an image repository, right?

Dockerfile_ubi
```diff
@@ -0,0 +1,41 @@
...
+RUN microdnf update && microdnf install systemd
+
+ENV TAG $BUILD_TAG
+ARG USER_UID=10001
+USER ${USER_UID}
+ENV HOME /etc/otel/
+
+COPY otelcol-sumo /
+COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY --from=directories --chown=${USER_UID}:${USER_UID} /etc/otel/ /etc/otel/
```
astencel-sumo, line 38: Why is this directory copied from Alpine and not just created here?

aboguszewski-sumo: see https://github.com/SumoLogic/sumologic-otel-collector/pull/1440#discussion_r1484357187

ci/build-push-multiplatform.sh
sumo-drosiek: We can, but what was fips tag before change wand what is it now?

swiatekm-sumo:
> > > dev builds are only to confirm that it builds successfully: IMO we should only have this enabled for releases
> > 
> > 
> > I disagree, anything we build for releases should also be built for dev. Otherwise we're never going to catch breakage.
> 
> I agree we should build, but we probably don't need to push the non-release UBI images to an image repository, right?

We currently do it for all builds on `main`, which I think is valid, as it makes it much easier to run external tests on dev versions.

ci/build-push-multiplatform.sh
aboguszewski-sumo: Okay, it seems that it does not affect it, but I bugged it and this is why there are failures. I will take care of that.

Dockerfile_ubi
```diff
@@ -0,0 +1,33 @@
...
+## Build RedHat compliant image
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.9
+
+ENV SUMMARY="UBI based sumologic-otel-collector" \
+    DESCRIPTION="Sumo Logic Distribution for OpenTelemetry Collector is a Sumo Logic-supported distribution of the OpenTelemetry Collector. It is a single agent to send logs, metrics and traces to Sumo Logic."
+
+LABEL name="sumologic-otel-collector" \
+      vendor="Sumo Logic" \
+      version="$BUILD_TAG" \
+      release="4" \
```
swiatekm-sumo, line 14: What is this label for?

aboguszewski-sumo: I wish I knew! This was already in the original Dockerfile for ubi, so I left it: https://github.com/SumoLogic/sumologic-openshift-images/blob/main/sumologic-otel-collector/0.57.2-sumo-1/Dockerfile

swiatekm-sumo: @kkujawa-sumo do you know?

swiatekm-sumo: Looking at this: https://developers.redhat.com/articles/2021/11/11/best-practices-building-images-pass-red-hat-container-certification#best_practice__6__embed_identifying_information_inside_your_image, this should probably just be `0`.

We should probably also apply security updates as per section 5 of that guide.

aboguszewski-sumo: Doesn't `microdnf update` already apply these updates? The command in the section 5 does not work with `microdnf` (some unsupported flags).

aboguszewski-sumo: Also, what should be equal to `0`? The `release` label?

swiatekm-sumo (approved):
LGTM, although we should add UBI FIPS images in a subsequent PR.

PR #1442, opened 2024-02-08
Add a test for zlib compressed data, as in our documentation

rnishtala-sumo:
Add a test for zlib compressed data, as in our documentation
https://help.sumologic.com/docs/send-data/hosted-collectors/http-source/logs-metrics/#compressed-data

### DIFF
--- a/pkg/exporter/sumologicexporter/sender_test.go
+++ b/pkg/exporter/sumologicexporter/sender_test.go
@@ -18,6 +18,7 @@ import (
 	"bufio"
 	"bytes"
 	"compress/gzip"
+	"compress/zlib"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


pkg/exporter/sumologicexporter/sender_test.go
```diff
@@ -1132,6 +1145,27 @@ func TestSendCompressZstd(t *testing.T) {
...
+	test := prepareSenderTest(t, configcompression.Deflate, []func(res http.ResponseWriter, req *http.Request){
+		func(res http.ResponseWriter, req *http.Request) {
+			res.WriteHeader(200)
+			if _, err := res.Write([]byte("")); err != nil {
+				res.WriteHeader(http.StatusInternalServerError)
+				assert.FailNow(t, "err: %v", err)
+				return
+			}
+			body := decodeZlib(t, req.Body)
+			assert.Equal(t, "deflate", req.Header.Get("Content-Encoding"))
```
sumo-drosiek, line 1158: why is it `deflate`?

swiatekm-sumo: That's the actual encoding the zlib format uses here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding.

swiatekm-sumo (approved):

PR #1443, opened 2024-02-09
fix: support --remote-config startup option with Windows services

fguimond:
With the introduction of remote management the application command line parameters were modified to include the `--remote-config` startup parameter. However this was unintentionally done only for interactive mode. Since the Windows service initialization code uses a different path the new parameter does not work correctly and the application exits.

The Windows service launch code path from the upstream repository uses the [flags](https://github.com/open-telemetry/opentelemetry-collector/blob/de6287de839a7d2c0273dd1df3a064a08f733904/otelcol/flags.go#L32) function defined in the upstream code, which doesn't support the `--remote-config` command line parameter. For interactive mode that [function was redefined](https://github.com/SumoLogic/sumologic-otel-collector/blob/c9a38b805ab7339fe02e8a03223421994dc572f3/otelcolbuilder/cmd/flags.go#L55) in our implementation of OtelCol to include the `--remote-config` command line parameter. 

This PRs copies the upstream Windows service launch routine and modifies it slightly to use our version of the `args` function instead of the upstream version, making it possible to launch the OtelCol as a Windows server.

Testing has been performed by installing a MSI that includes this modified code. The Windows service now launches correctly as illustrated by the Sumo Logic console and Windows Service console.

image
image

### DIFF
--- a/otelcolbuilder/cmd/collector_windows.go
+++ b/otelcolbuilder/cmd/collector_windows.go
@@ -1,0 +1,247 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build windows
+
+// This is a copy of https://github.com/open-telemetry/opentelemetry-collector/blob/20710ff2aeaed3fe1c3e46423291522674db44d6/otelcol/collector_windows.go.
+// This is required to maintain our command line flags when running the collector as a Windows service, until either
+// the necessary names become public, or there's a better way of customizing config providers.
+
+package main // import "go.opentelemetry.io/collector/otelcol"
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+	"go.opentelemetry.io/collector/confmap"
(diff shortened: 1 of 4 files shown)


otelcolbuilder/cmd/collector_windows.go
```diff
@@ -0,0 +1,247 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build windows
+
+// This is a copy of https://github.com/open-telemetry/opentelemetry-collector/blob/20710ff2aeaed3fe1c3e46423291522674db44d6/otelcol/collector_windows.go.
+// This is required to maintain our command line flags when running the collector as a Windows service, until either
+// the necessary names become public, or there's a better way of customizing config providers.
```
rnishtala-sumo, line 8: it looks like `NewSvcHandler` is public already, which seems to be the only thing we use in the patch, do we need other names/functions to be public? If so, maybe we could identify those and create an upstream issue for it.

rnishtala-sumo (approved):

PR #1448, opened 2024-02-15
chore(k8sprocessor): Improve logging of missing data events

swiatekm-sumo:
Log warnings when we're missing data we should have. This usually means the cache is out of sync. Right now we don't have any observability for these kinds of events.

### DIFF
--- a/pkg/processor/k8sprocessor/kube/client.go
+++ b/pkg/processor/k8sprocessor/kube/client.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/internal/k8sconfig"
+
 	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/k8sprocessor/observability"
 )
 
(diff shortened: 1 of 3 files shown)


sumo-drosiek (approved):

pkg/processor/k8sprocessor/kube/client.go
```diff
@@ -321,6 +322,9 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
 	}
 
 	if c.Rules.NodeName {
+		if len(pod.Spec.NodeName) == 0 {
+			c.logger.Warn("missing Node name for Pod, cache may be out of sync", zap.String("pod", pod.Name))
```
astencel-sumo, line 326: We need to watch if these logs are not too verbose.
Metrics for these would be a good solution too.

astencel-sumo (approved):

pkg/processor/k8sprocessor/kube/client.go
swiatekm-sumo: I thought about metrics, but realistically this is a fault that won't happen very often. For this to occur, a Pod needs to get scheduled, start, emit logs, and the logs get collected and pass through the processor, and in the meantime the processor can't get any updates from the API Server. There's already similar warnings in client-go itself, they're just not very specific.

PR #1458, opened 2024-02-19
fix(scripts/install.sh): improve install failure detection

swiatekm-sumo:
I've had some reports of the Mac install succeeding despite otelcol itself failing. I've tweaked the failure detection to be more precise and increased the timeout to 15s. This appears to fix the problem in all cases I'm aware of, but we should keep an eye on it - maybe even abandon this detection entirely or do a different kind of check?

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1947,9 +1947,20 @@ if [[ "${OS_TYPE}" == "darwin" ]]; then
     launchctl unload "${LAUNCHD_CONFIG}"
     launchctl load -w "${LAUNCHD_CONFIG}"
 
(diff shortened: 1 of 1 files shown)


sumo-drosiek (approved):
In case it won't help, we can try to use metrics endpoint or another more reliable heath checking way

swiatekm-sumo:
> In case it won't help, we can try to use metrics endpoint or another more reliable heath checking way

Maybe that'd be a better solution, detect if otelcol has started and then set a timeout for when the health check should pass? Then we can have false positives the other way though.

PR #1462, opened 2024-02-20
fic(docs): fix upgrading.md

sumo-drosiek:


astencel-sumo (approved):

swiatekm-sumo (changes requested):
Let's leave this as-is and retract 0.94.0-sumo-1 instead.

PR #1469, opened 2024-02-26
feat(OTRM): Add validation for otel configuration in the opamp agent

rnishtala-sumo:
Add validation for otel configuration in the opamp agent.  These changes cover syntax/semantic issues with the configuration. Reverting configuration after a runtime issue is not part of this change.


pkg/extension/opampextension/components.go
```diff
@@ -0,0 +1,65 @@
...
+	var errs error
+
+	extensions, err := extension.MakeFactoryMap(
+		ballastextension.NewFactory(),
+		sumologicextension.NewFactory(),
+	)
+	errs = multierr.Append(errs, err)
+
+	receivers, err := receiver.MakeFactoryMap(
+		filelogreceiver.NewFactory(),
```
rnishtala-sumo, line 38: This is not an exhaustive list of receivers, exporters, processors and extensions yet.

pkg/receiver/jobreceiver/output/config.go
```diff
@@ -77,7 +77,7 @@ func (c *Config) Unmarshal(component *confmap.Conf) error {
 	}
 
 	c.Builder = factory.CreateDefaultConfig()
-	if err := subComponent.Unmarshal(c.Builder, confmap.WithErrorUnused()); err != nil {
```
rnishtala-sumo, line 80: confmap version was bumped up.

pkg/exporter/sumologicexporter/config.go
```diff
@@ -83,18 +83,18 @@ func CreateDefaultHTTPClientSettings() confighttp.HTTPClientSettings {
 func (cfg *Config) Validate() error {
 
 	switch cfg.CompressEncoding {
-	case configcompression.Gzip:
-	case configcompression.Deflate:
+	case configcompression.TypeGzip:
```
rnishtala-sumo, line 86: Changes that account for upstream deprecations

pkg/extension/opampextension/opamp_agent.go
```diff
@@ -146,7 +147,7 @@ func (o *opampAgent) startClient(ctx context.Context) error {
 		InstanceUid:    o.instanceId.String(),
 		Callbacks: types.CallbacksStruct{
 			OnConnectFunc: func() {
-				o.logger.Info("Connected to the OpAMP server")
+				o.logger.Info("==================Connected to the OpAMP server==================")
```
echlebek, line 150: IMO having extra punctuation here will just add noise to peoples' logs.

pkg/extension/opampextension/opamp_agent.go
```diff
@@ -424,6 +425,18 @@ func (o *opampAgent) saveEffectiveConfig(dir string) error {
...
+		o.logger.Debug("Loading Component Factories...")
+		factories, err := Components()
+		if err != nil {
+			return fmt.Errorf("cannot get the list of factories: %v", err)
+		}
+		o.logger.Info("Loading Configuration to Validate...")
+		_, errValidate := otelcoltest.LoadConfigAndValidate(p, factories)
+		if errValidate != nil {
+			o.logger.Error("Validation Failed... %v", zap.Error(errValidate))
+			return fmt.Errorf("cannot validate config named %v", errValidate)
```
echlebek, line 437: Is this error formatting correct? Seems like it should be something like `Errorf("cannot validate config named %q: %s", name, err)`

pkg/extension/opampextension/opamp_agent.go
```diff
@@ -158,7 +159,7 @@ func (o *opampAgent) startClient(ctx context.Context) error {
 				o.remoteConfigStatus = status
 			},
 			GetEffectiveConfigFunc: func(ctx context.Context) (*protobufs.EffectiveConfig, error) {
-				o.logger.Info("OpAMP server requested the effective configuration")
+				o.logger.Info("==================OpAMP server requested the effective configuration==================")
```
echlebek, line 162: Ditto

pkg/extension/opampextension/opamp_agent.go
rnishtala-sumo: This is a test which shows the example formatting
https://github.com/SumoLogic/sumologic-otel-collector/pull/1469/files#diff-8c8da29ce13818d83c803112ef661f02f923afe91f5c30f120447ae2204d896fR240

pkg/exporter/sumologicexporter/config.go
swiatekm-sumo: Can you submit these in a separate PR? Would make this one easier to review.

pkg/extension/opampextension/components.go
swiatekm-sumo: Is the plan to make this extension depend on every component it validates configuration for? That sounds like a massive headache.

rnishtala-sumo: I didn't find an easier way to initialize `factories` which is a pre-requisite to validate config. This is used as a reference for all the components defined in a configuration file. The intent here is to work with the backend team to ensure that all the components defined in source templates are initialized here, so we explicitly run validations on them.

It is definitely more maintenance overhead, but one could potentially reduce it by commonly used receivers/processors  in this PR. I don't think new components will be added to our backend that often except for maybe new `receivers` and `processors`. Other components like `extensions` and `exporters` shouldn't change much.

rnishtala-sumo: ConfigProvider needs these component factories to be passed in as a parameter

https://pkg.go.dev/go.opentelemetry.io/collector/otelcol#ConfigProvider
which then returns a `Config` type that one can run `Validate` on 

This is the same thing that the existing command line validate option does
https://github.com/open-telemetry/opentelemetry-collector/blob/main/otelcol/collector.go#L223

swiatekm-sumo: I was thinking of the alternative of simply running `validate` in a subprocess directly. Then we don't have to explicitly depend on all the components we *could* have in a remote configuration. Correct me if I'm wrong, but I believe a user can supply a custom remote config as well, and we won't be able to validate that if we only include a subset of components here.

rnishtala-sumo: yes we would need to account for all the components used in Source templates. `validate` does the same thing under the hood where it calls `factories, err := collector.set.Factories()` where `collector` is an instance of the otel collector.

One advantage with defining our own components is that we'll be able to validate any custom components that haven't been upstreamed yet. So anytime a new component needs to be validated we wouldn't need to create a new upstream issue for it.

swiatekm-sumo: Correct me if I'm wrong, but `validate` in a custom collector build like ours uses all the components included in the collector, right? I was thinking that we could call `./otelcol-sumo -c <config.yaml> validate` in a subprocess. We already have the collector binary with all the necessary components. The only problem would be parsing the output. But in that case we can avoid having to depend on the whole universe in this extension.

rnishtala-sumo: ah yes that makes sense, it does seem a little counterintuitive to build the otelcol binary within the opamp extension and run it as a subprocess though. Are there other places that use this approach.

swiatekm-sumo: I don't want to build the otelcol binary, we already have it built, that's how our extension is running in the first place. I don't think anyone else does this, but I'd like to check if it's viable before committing to the dependency route.

rnishtala-sumo: it looks like one downside of this approach is that it's hard to write tests for it. I may need to build an executable in the tests.

rnishtala-sumo: Well, this didn't seem to work, here are some logs from my attempt
```
2024-02-28T15:07:06.330-0500	info	opampextension@v0.0.0-00010101000000-000000000000/opamp_agent.go:465	==============File found	{"kind": "extension", "name": "opamp", "file": "/etc/otelcol-sumo/opamp.d/00000000000047E4.yaml"}
2024-02-28T15:07:06.842-0500	info	opampextension@v0.0.0-00010101000000-000000000000/opamp_agent.go:465	==============File found	{"kind": "extension", "name": "opamp", "file": "/usr/local/bin/otelcol-sumo"}

Failed to apply OpAMP agent remote config	{"kind": "extension", "name": "opamp", "error": "cannot save the OpAMP effective config to /etc/otelcol-sumo/opamp.d: fork/exec /usr/local/bin/otelcol-sumo validate --config /etc/otelcol-sumo/opamp.d/00000000000047E4.yaml: no such file or directory"}
```
Eventhough the binary and the config exist `exec.command` errored out with `no such file or directory`. My understanding here is that the subprocess has the same permissions as the parent from which it was forked. Looks like running `validate` requires elevated permissions.

Below is the code snippet that I used for this
```
		execPath, err := os.Executable()
		if err != nil {
			return err
		}
		command := execPath + " validate --config " + p
                 o.waitForFile(p, 2*time.Second)
		o.waitForFile(execPath, 2*time.Second)
		out, err := exec.Command(command).Output()
		if err != nil {
			o.logger.Error("Config Validation Failed...", zap.String("config", p))
			return err
		}
		o.logger.Info(string(out))

```

rnishtala-sumo: Indeed, I need to run it as sudo
```
sudo /usr/local/bin/otelcol-sumo validate --config /etc/otelcol-sumo/opamp.d/00000000000047E4.yaml
Error: exporters::sumologic: 'max_elapsed' time must be non-negative
2024/02/28 15:18:02 collector server run finished with error: exporters::sumologic: 'max_elapsed' time must be non-negative
```

rnishtala-sumo: Well, this didn't seem to work, here are some logs from my attempt
```
2024-02-28T15:07:06.330-0500	info	opampextension@v0.0.0-00010101000000-000000000000/opamp_agent.go:465	==============File found	{"kind": "extension", "name": "opamp", "file": "/etc/otelcol-sumo/opamp.d/00000000000047E4.yaml"}
2024-02-28T15:07:06.842-0500	info	opampextension@v0.0.0-00010101000000-000000000000/opamp_agent.go:465	==============File found	{"kind": "extension", "name": "opamp", "file": "/usr/local/bin/otelcol-sumo"}

Failed to apply OpAMP agent remote config	{"kind": "extension", "name": "opamp", "error": "cannot save the OpAMP effective config to /etc/otelcol-sumo/opamp.d: fork/exec /usr/local/bin/otelcol-sumo validate --config /etc/otelcol-sumo/opamp.d/00000000000047E4.yaml: no such file or directory"}
```
Eventhough the binary and the config exist `exec.command` errored out with `no such file or directory`. My understanding here is that the subprocess has the same permissions as the parent from which it was forked. Looks like running `validate` requires elevated permissions.

Below is the code snippet that I used for this
```
		execPath, err := os.Executable()
		if err != nil {
			return err
		}
		command := execPath + " validate --config " + p
                 o.waitForFile(p, 2*time.Second)
		o.waitForFile(execPath, 2*time.Second)
		out, err := exec.Command(command).Output()
		if err != nil {
			o.logger.Error("Config Validation Failed...", zap.String("config", p))
			return err
		}
		o.logger.Info(string(out))

```

rnishtala-sumo: @swiatekm-sumo please let me know your thoughts on the above, I'm not sure if there's a point in elevating the permissions to validate the configuration.
```
-r-xr-xr-x  1 root  wheel  253120592 Feb 20 07:35 otelcol-sumo
```

swiatekm-sumo: Ok, that's a good reason not to do it. Let's proceed with your current idea. We'll need to be careful with what we include in these component lists though.

pkg/extension/opampextension/opamp_agent.go
rnishtala-sumo: @echlebek does the above example make sense or would you like to see a change in the format?

echlebek: I see, that makes sense

echlebek (approved):

PR #1475, opened 2024-03-05
test(install-script): refactor tests in prep for Windows

swiatekm-sumo:
Move around the test code to make it easier to share with Windows. I'm planning to add tests for the Windows install script as a follow-up.


rnishtala-sumo (approved):
makes sense, it looks like the intent is to have these tests run only on a linux box/container

sumo-drosiek (approved):

PR #1480, opened 2024-03-08
test(scripts/install.ps1): check file ownership and permissions

swiatekm-sumo:
Check file ownership and permissions in Windows install script tests.

### DIFF
--- a/pkg/scripts_test/consts_windows.go
+++ b/pkg/scripts_test/consts_windows.go
@@ -3,8 +3,8 @@
 package sumologic_scripts_tests
 
 const (
-	systemGroup string = "otelcol-sumo"
-	systemUser  string = "otelcol-sumo"
+	// See: https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-identifiers
+	localSystemSID string = "S-1-5-18"
 
 	packageName string = "OpenTelemetry Collector"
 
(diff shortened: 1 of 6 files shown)


pkg/scripts_test/consts_windows.go
```diff
@@ -23,10 +24,10 @@ const (
...
-	configPathFilePermissions       uint32 = 0440
-	confDPathFilePermissions        uint32 = 0644
-	etcPathPermissions              uint32 = 0551
-	opampDPermissions               uint32 = 0750
+	commonConfigPathFilePermissions uint32 = 0777
+	configPathDirPermissions        uint32 = 0777
+	configPathFilePermissions       uint32 = 0666
+	confDPathFilePermissions        uint32 = 0777
+	etcPathPermissions              uint32 = 0777
+	opampDPermissions               uint32 = 0777
```
echlebek, line 32: While I'm sure there is a good reason for it, this appears to be quite liberal permissions, maybe you could add a comment explaining why?

echlebek (commented):
Looks good to me, but I find the 777 and 666 permissions a bit suspicious and could use an explainer.

pkg/scripts_test/consts_windows.go
swiatekm-sumo: Will do. To be honest, I should really use Windows ACLs here and make more detailed assertions. Using os.Stat like we do for Windows or Unix, we can only really get 0444 for readonly files and 0666 for writable ones (+0111 for directories). So currently all this checks is whether the files are writable by their owner (which is the LocalSystem user who owns files for shared Programs.)

astencel-sumo (approved):

pkg/scripts_test/consts_windows.go
swiatekm-sumo: I added a comment explaining these. Looks like Windows ACLs aren't trivial to check, so I'm going to merge this PR, and add those in a followup.

PR #1481, opened 2024-03-11
Upgrade opamp-go to tip

echlebek:
```
    Upgrade opamp-go to tip
    
    Upgrade to an untagged opamp-go in order to take advantage of the recent
    support for HTTP redirection that has been added.
    
    The project has broken several things since 0.10.x, and this breakage
    has been reflected as minor changes to the opamp extension as well.
    
    opamp-go should be upgraded to a tagged version on a subsequent release.
```

### DIFF
--- a/.changelog/1481.added.txt
+++ b/.changelog/1481.added.txt
@@ -1,0 +1,1 @@
+opampextension: add support for redirection on websocket connect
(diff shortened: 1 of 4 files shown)


swiatekm-sumo (approved):
Can you also create an issue to track switching back to stable once these changes are released?

PR #1508, opened 2024-03-18
ci: trigger packaging for release builds

swiatekm-sumo:
Automatically trigger the packaging workflow for release builds.

### DIFF
--- a/.github/workflows/release_builds.yml
+++ b/.github/workflows/release_builds.yml
@@ -632,3 +632,10 @@ jobs:
           artifacts: "artifacts/*/*"
           artifactErrorsFailBuild: true
           replacesArtifacts: true
+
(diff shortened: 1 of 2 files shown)


aboguszewski-sumo:
Do you have any test workflow that passed? Can you link it here?

swiatekm-sumo:
> Do you have any test workflow that passed? Can you link it here? Or is it something we've been using for a while?

We've been triggering it automatically for dev builds for a while now, and doing it manually for release builds. I haven't tested it (it's painful to do on a fork), but it should be very safe. Worst case, it fails during the next release and we do it manually just like we've done till now, so the risk is small.

aboguszewski-sumo (approved):

PR #1509, opened 2024-03-18
feat(opamp): Support adding and validating more receivers and extensions in the source templates

rnishtala-sumo:
Adds validation for the windows event log receiver and other extensions. This PR ensures that the following components are recognized and validated by the opamp agent.

- windows event log receiver
- opamp extension
- sumologic extension
- healthcheck extension
- pprof extension
- filestorage extension
- Ngnix 
- Redis 
- Kafka 
- Docker 
- RabbitMQ 
- Windows Host metrics


echlebek (approved):
Looks good to me!

PR #1512, opened 2024-03-20
fix(windows): Windows service not starting up

fguimond:
Upstream code was modified, which broke the Windows service startup logic. Brought the latest code from upstream to override the Windows service arguments.


otelcolbuilder/cmd/collector_windows.go
```diff
@@ -220,20 +224,25 @@ func withWindowsCore(elog *eventlog.Log) func(zapcore.Core) zapcore.Core {
...
-		var err error
-		set.ConfigProvider, err = otelcol.NewConfigProvider(newDefaultConfigProviderSettings(configFlags))
-		if err != nil {
-			return nil, err
+		if len(configFlags) > 0 {
+			resolverSet.URIs = configFlags
+		}
+		if len(resolverSet.URIs) == 0 {
+			return errors.New("at least one config flag must be provided")
+		}
```
sumo-drosiek, line 237: I would change the order of validation

```suggestion
		if len(configFlags) == 0 {
			return errors.New("at least one config flag must be provided")
		}
		if len(configFlags) > 0 {
			resolverSet.URIs = configFlags
		}
```

fguimond: @sumo-drosiek - Agreed, however this code was copied directly from [upstream](https://github.com/open-telemetry/opentelemetry-collector/blob/4627d3c18ea8cd82061a70f8db30ef98426e77db/otelcol/command.go#L44) so I left it alone.

sumo-drosiek (approved):

rnishtala-sumo (approved):

PR #1516, opened 2024-03-25
feat(sumologicextension): deprecate in favor of OpenTelemetry Contrib extension

sumo-drosiek:
Deprecate Sumo Logic Extension in favor of upstream ones

### DIFF
--- a/examples/sumologic-windows.yaml
+++ b/examples/sumologic-windows.yaml
@@ -3,7 +3,7 @@
 extensions:
   ## Configuration for Sumo Logic Extension
   ## Manages registration, heartbeats and authentication to Sumo Logic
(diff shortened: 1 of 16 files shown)


swiatekm-sumo:
Should we also change the extension in sumo exporter's go.mod?

sumo-drosiek:
Converting to draft as waiting for new OpenTelemetry Collector Release

rnishtala-sumo:
@sumo-drosiek Are there any breaking changes with this? Maybe we could capture differences (if any) between the upstream extension and the local one, just to be sure they haven't diverged feature-wise.

sumo-drosiek:
@rnishtala-sumo There shouldn't be any. I don't see any traffic in extension from 3 month. It should be basically the same code

PR #1519, opened 2024-03-27
Remove URL-rewriting hack

echlebek:
This commit removes a hack that rewrites the collector URL for use with OpAmp. The collector now supports redirection, and we expect the Sumo Logic backend to correctly redirect the client.


echlebek:
Notes for reviewers:

* tested successfully against us2data long, with remote management (I don't know if redirection occurred or not)
* users should not notice any difference, I think it's probably better to omit a changelog entry

swiatekm-sumo (approved):
The change itself looks good to me, but we should check against a few more environments than just us2. We should also get approval from the relevant Opamp backend SMEs before merging this imo.

skumar-sumo (changes requested):
please move this changes behind feature flag rather removing the old code. 
As the otrm for existing customers will break till we test this feature end-to-end on production.

echlebek:
@skumar-sumo I don't see the logic of having a feature flag here, what's the suggested default? Can't we support both the old and new strategy on the backend?

ggautam-sumo (changes requested):
This change will break OT remote management. As of now, 
1. Backend changes are not tested. Its pending on Agent build that supports http/wss redirection.
2. After testing with Agent, BE solution will be rollout on test environment and then to small prods.
3. We would also need QE sign off on the redirection solution to rollout everywhere. 
4. We don't have a rollback solution discussed or in place. (Hope we don't need it)

After, 1, 2 and 3 is done. We can safely remove Agent side hack.

echlebek:
The agent side hack has been shown to be buggy and does not work correctly in all circumstances. Per recent discussion, it should be removed.

rnishtala-sumo (approved):

rnishtala-sumo:
@echlebek should we just use this PR to set the new opamp endpoint? I think the install script has moved to the [packaging repository](https://github.com/SumoLogic/sumologic-otel-collector-packaging/tree/main/install-script).

echlebek:
QE has found an issue with this PR, will update.

PR #1522, opened 2024-04-02
chore: fix some shellcheck problems

swiatekm-sumo:
* Update shellcheck to the latest 0.10.0
* Run shellcheck on the install script. This was not the case until now, probably by mistake.
* Delete shellcheck.sh and instead explicitly list the files we're checking in the Makefile target.
* Fix some warning and info messages in our scripts.

### DIFF
--- a/.github/workflows/pull_requests.yml
+++ b/.github/workflows/pull_requests.yml
@@ -104,7 +104,7 @@ jobs:
   shellcheck:
     runs-on: ubuntu-20.04
     env:
-      SHELLCHECK_VERSION: 0.8.0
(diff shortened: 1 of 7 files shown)


aboguszewski-sumo (approved):
A nit, because I'm not sure I understand: do these problems break anything or is it just improving life quality?
Nevertheless, LGTM

swiatekm-sumo:
> A nit, because I'm not sure I understand: do these problems break anything or is it just improving life quality? Nevertheless, LGTM

They're either:
* Removing dead code
* Fixing actual minor bugs
* Making the code more robust, for example by preventing it from swallowing errors

PR #1525, opened 2024-04-05
fix(scripts/install.sh): don't sort releases alphabetically

swiatekm:
This fails whenever we add a digit to a version. For example, right now, we get the 0.96.0-938 MacOS package instead of v0.96.0-1057.

I've added tests for this behaviour for both install scripts.

### DIFF
--- a/pkg/scripts_test/consts_common.go
+++ b/pkg/scripts_test/consts_common.go
@@ -18,50 +18,45 @@ var (
 	latestAppVersion string
 )
 
-func getAppReleaseVersions() ([]string, error) {
+func getLatestAppReleaseVersion() (string, error) {
 	githubApiBaseUrl, err := url.Parse(GithubApiBaseUrl)
 	if err != nil {
-		return nil, err
+		return "", err
 	}
-	githubApiReleasesUrl := githubApiBaseUrl.JoinPath(
+	githubApiLatestReleaseUrl := githubApiBaseUrl.JoinPath(
 		"repos",
 		GithubOrg,
 		GithubAppRepository,
-		"releases")
-	response, err := http.Get(githubApiReleasesUrl.String()) //nolint:noctx
+		"releases",
+		"latest")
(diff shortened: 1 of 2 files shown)


sumo-drosiek:
What if last listed version from github is bugfix or version which won't be the real latest, eg. packaging fix for previous otel version?

Can we use [the following](https://man7.org/linux/man-pages/man1/sort.1.html)?

>        -V, --version-sort
>              natural sort of (version) numbers within text

swiatekm-sumo:
> What if last listed version from github is bugfix or version which won't be the real latest, eg. packaging fix for previous otel version?
> 
> Can we use [the following](https://man7.org/linux/man-pages/man1/sort.1.html)?
> 
> > ```
> >    -V, --version-sort
> >          natural sort of (version) numbers within text
> > ```

Done. 

But do we really even need to do this manually? Why not just fetch the latest release via the API explicitly? It looks like the only reason we need to know about other versions is that we want to print breaking changes between the currently installed version and the one we're installing. So it looks like we can only fetch the latest version,

sumo-drosiek:
> But do we really even need to do this manually? Why not just fetch the latest release via the API explicitly?

I think we can, but Github doesn't seem to respect `latest` label. Also now, it seems that the releases are sorted alphabetically 😕 
image

Conclusion is that for now it should work, unless github changes something

swiatekm-sumo:
> > But do we really even need to do this manually? Why not just fetch the latest release via the API explicitly?
> 
> I think we can, but Github doesn't seem to respect `latest` label. Also now, it seems that the releases are sorted alphabetically 😕 image
> 
> Conclusion is that for now it should work, unless github changes something

Github sorts releases by publishing date if you ask for all of them, but if you ask for the latest one specifically (`/releases/latest`), you'll get the one marked as `latest`. Which is what we want.

sumo-drosiek:
> Github sorts releases by publishing date if you ask for all of them

Not really sure about it anymore
image

> Which is what we want. But I'd rather make that change in a new PR.

I'm fine with that

PR #1531, opened 2024-04-09
docs(vagrant): document Windows dev environment on Apple M1

swiatekm-sumo:
For now, just point to UTM. We can provide better instructions once we figure this out.


echlebek (approved):
Seems reasonable :+1:

PR #1536, opened 2024-04-10
chore: remove install scripts

swiatekm-sumo:
The scripts have been moved to the packaging repository. We still attach them to releases, but we use the version stored in that repository. There is no reason to keep them here anymore.


echlebek (approved):

rnishtala-sumo:
@swiatekm-sumo just want to confirm that this [action](https://github.com/SumoLogic/sumologic-otel-collector/blob/main/.github/workflows/release_builds.yml#L530) will still upload the install script because it's now available in `SumoLogic/sumologic-otel-collector-packaging` and that our releases will still include this as an asset? I think the collector installation workflow in the UI depends on it.

swiatekm-sumo:
> @swiatekm-sumo just want to confirm that this [action](https://github.com/SumoLogic/sumologic-otel-collector/blob/main/.github/workflows/release_builds.yml#L530) will still upload the install script artifact because it's now available in `SumoLogic/sumologic-otel-collector-packaging` and that our collector releases will still include this as an asset? I think the collector installation workflow in the UI depends on it.

Yes, it will. You can see it happen for dev builds, here for example: https://github.com/SumoLogic/sumologic-otel-collector/actions/runs/8643263953.

rnishtala-sumo (approved):

PR #1547, opened 2024-04-16
feat(otrm): Support validating filter processors in the source templates

rnishtala-sumo:
Support validating filter and transform processors in the source templates


pkg/extension/opampextension/testdata/opamp.d/opamp-filter-processor.yaml
```diff
@@ -0,0 +1,103 @@
...
+          - windowseventlog/Microsoft-Windows-PowerShell/Operational/09c98ea4-6686-4ab9-b3cd-f04070906441
+          - windowseventlog/Amazon EC2Launch/09c98ea4-6686-4ab9-b3cd-f04070906441
+        processors:
+          - memory_limiter
+          - resourcedetection/system
+          - batch
+          - resource/09c98ea4-6686-4ab9-b3cd-f04070906441
+          - filter/logs/09c98ea4-6686-4ab9-b3cd-f04070906441
+        exporters:
+          - sumologic
```
sumo-drosiek, line 103: Do we need all of this configuration just to test filterprocessor?

rnishtala-sumo: this is an example source template from the backend team, that I thought would be useful to run a test against.

sumo-drosiek: I feel like it is more than needed. Personally I would like to limit test scope to what is necessary, eventually we can add another test which will be more complex, but now most of the configuration seems to be unrelated to the name of the test/file

sumo-drosiek: If the test fail, it won't be obvious why

rnishtala-sumo: ok I'll trim some portions of the config.

rnishtala-sumo: how does this look - https://github.com/SumoLogic/sumologic-otel-collector/pull/1547/commits/7fcad3e5406cebd3fb5e8bf231ab66953b3f8851

it has 1 receiver, 2 filter processors, 1 transform processor, 1 exporter.

pkg/extension/opampextension/testdata/opamp.d/opamp-filter-processor.yaml
```diff
@@ -0,0 +1,45 @@
+  receivers:
+    hostmetrics/09c98ea4-6686-4ab9-b3cd-f04070906441:
+      collection_interval: 5m
+      scrapers:
+        network:
+  processors:
+    batch:
+      send_batch_size: 1_024
+      timeout: 1s
```
sumo-drosiek, line 9: 🙈

pkg/extension/opampextension/testdata/opamp.d/opamp-filter-processor.yaml
sumo-drosiek: Yup, thats better, LGTM :) (I marked batch processor, you can clean it up if unnecessary)

sumo-drosiek (approved):

PR #1551, opened 2024-04-19
chore: add non-k8s for tracing examples

mat-rumian:
PR adds example configuration for tracing.


swiatekm-sumo (approved):
I wonder if we should move all this to the docs site.

PR #1555, opened 2024-04-22
ci: sign MacOS dev builds

swiatekm-sumo:


swiatekm-sumo:
> LGTM but strange github actions arreor about buildx and arm 😕

Restart fixed it :shrug:

PR #1562, opened 2024-04-24
ci: set the right app version for release builds

swiatekm:
#1559 fixed replacing tags in the builder configuration, but the release workflow also sets the tags incorrectly. This PR adds a job which determines the application version, and then uses that value in all subsequent jobs.

### DIFF
--- a/.github/workflows/release_builds.yml
+++ b/.github/workflows/release_builds.yml
@@ -21,26 +21,46 @@ env:
   GO_VERSION: "1.21.4"
 
 jobs:
+  get-version:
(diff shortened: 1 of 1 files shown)


sumo-drosiek (approved):

swiatekm-sumo:
Release workflow with this change in my fork: https://github.com/swiatekm-sumo/sumologic-otel-collector/actions/runs/8820929475. And the resulting release: https://github.com/swiatekm-sumo/sumologic-otel-collector/releases/tag/v0.98.0-sumo-1-rc.8.

PR #1564, opened 2024-04-25
ci: set version tags for dev build

swiatekm-sumo:
This change moves dev builds closer to how release builds are. It consistently sets the same version for the dev build anywhere applicable:

* The application version is now reported as such: 
   ```
    buildinfo:
    command: otelcol-sumo
    description: *** distribution
    version: v0.98.0-sumo-0-6-g3bb36
    ```
* Internally, our component versions are now set to this value:
   ```
  2024-04-25T11:24:46.172+0200    info    sumologicexporter@v0.98.0-sumo-0-6-g3bb36/exporter.go:88        Sumo Logic Exporter configured  {"kind": "exporter", "data_type": "logs", "name": "sumologic", "log_format": "otlp", "metric_format": "otlp", "trace_format": "otlp"}
  ```

* Dev container images now use this version as the tag, instead of the commit hash:
  ```
  public.ecr.aws/sumologic/sumologic-otel-collector-dev:0.98.0-sumo-0-6-g3bb36
  ```

The added parameter to our reusable build workflow will also let us use it for release builds.

### DIFF
--- a/.github/workflows/dev_builds.yml
+++ b/.github/workflows/dev_builds.yml
@@ -15,6 +15,23 @@ env:
 
 jobs:
 
+  get-version:
+    name: Get application version for this revision
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get-version.outputs.version }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Get version
+        id: get-version
+        run: |
+          TAG=$(git describe --tags --abbrev=5 --match "v[0-9]*")
+          echo "version=${TAG/v}" > $GITHUB_OUTPUT
+      - name: Print version
+        run: echo ::notice title=Version::${{ steps.get-version.outputs.version }}
+
   # Add lint to dev builds as that's the only way for cache to be shared across branches.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -210,10 +210,10 @@ delete-remote-tag:
 .PHONY: prepare-tag
 prepare-tag: install-gsed
 	@[ "${TAG}" ] || ( echo ">> env var TAG is not set"; exit 1 )
-	$(SED) -i 's#\(gomod: github.com/SumoLogic/sumologic-otel-collector/.*\) v0.0.0-00010101000000-000000000000#\1 ${TAG}#g' \
+	$(SED) -i'' -e 's#\(gomod: github.com/SumoLogic/sumologic-otel-collector/.*\) v0.0.0-00010101000000-000000000000#\1 ${TAG}#g' \
```
sumo-drosiek, line 213: Why do we need it if we use `$(SED)`?

swiatekm-sumo: Good question! Looking at https://github.com/SumoLogic/sumologic-otel-collector/actions/runs/8829744265/job/24241433602, apparently the target depends on `go` being installed. Which is reasonable, but not obvious. I'll modify the workflow instead of making this change.

sumo-drosiek (approved):

mat-rumian (approved):

aboguszewski-sumo (approved):

PR #1576, opened 2024-05-06
fix(ci): set build version in Windows ProductVersion to ot_channel_version

kkujawa-sumo:
related CI failure: https://github.com/SumoLogic/sumologic-otel-collector/actions/runs/8970924338/job/24636297516
doc: https://learn.microsoft.com/en-us/windows/win32/msi/productversion

so the output for will be following:
```
% export VERSION_TAG=v3.99.4-sumo-5-rc.6
% ./get_version.sh productversion                                                               
3.99.4.5
% export VERSION_TAG=v3.99.4-sumo-5     
% ./get_version.sh productversion  
3.99.4.5
```
so as it was before - no difference in Windows ProductVersion for release candidate and regular release


kkujawa-sumo:
The issue is resolved in different way, please see: https://github.com/SumoLogic/sumologic-otel-collector/pull/1577

Issue #1571, opened 2024-05-02
Puppet module is broken post change of install scripts URL


maulik-shah-tala:
Post changes done in https://github.com/SumoLogic/sumologic-otel-collector/commit/dd29f3aec4ccfca716f195844ebdf1acb8a057da - the install scripts are now available at https://raw.githubusercontent.com/SumoLogic/sumologic-otel-collector-packaging/main/install-script/install.sh instead of https://raw.githubusercontent.com/SumoLogic/sumologic-otel-collector/main/scripts/install.sh

This breaks puppet as the init.pp from the Puppet tarball at https://github.com/SumoLogic/sumologic-otel-collector/releases/latest/download/puppet-module.tar.gz still uses the old install script URL and it is not a variable that can be overriden in puppet

Can someone please update the puppet tarball to use the right install URL?

cc @swiatekm-sumo

PR #1580, opened 2024-05-07
Revert "chore: remove install scripts"

swiatekm-sumo:
Reverts SumoLogic/sumologic-otel-collector#1553. As per https://github.com/SumoLogic/sumologic-otel-collector/issues/1571, we missed urls in the puppet/chef/ansible modules. I'll fix these in a follow-up. We'll need to keep the scripts around in `main` for a bit longer though, it looks like.

### DIFF
--- a/.changelog/1553.changed.txt
+++ b/.changelog/1553.changed.txt
@@ -1,5 +1,0 @@
-chore: remove install scripts
-
-The install scripts were moved to the [packaging repository](https://github.com/SumoLogic/sumologic-otel-collector-packaging).
-They will continue to be shipped as artifacts for collector releases for the foreseeable future, but it is strongly
-suggested to switch any urls to the new location.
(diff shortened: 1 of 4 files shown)


aboguszewski-sumo (approved):

rnishtala-sumo:
is there a reason why we shouldn't just change these urls in chef/puppet instead of reverting this?

- https://github.com/SumoLogic/sumologic-otel-collector/blob/main/examples/puppet/modules/install_otel_collector/manifests/init.pp#L30

- https://github.com/SumoLogic/sumologic-otel-collector/blob/main/examples/chef/sumologic-otel-collector/resources/default.rb#L24

swiatekm-sumo:
> is there a reason why we shouldn't just change these urls in chef/puppet instead of reverting this?
> 
>     * https://github.com/SumoLogic/sumologic-otel-collector/blob/main/examples/puppet/modules/install_otel_collector/manifests/init.pp#L30
> 
>     * https://github.com/SumoLogic/sumologic-otel-collector/blob/main/examples/chef/sumologic-otel-collector/resources/default.rb#L24

That won't fix anything until a new release, forcing all existing users to upgrade. We owe our users a few releases worth of transition time.

PR #1583, opened 2024-05-08
doc: Update readme with the support matrix for OT collectors

rnishtala-sumo:
Update readme with the support matrix for OT collectors

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,25 +61,24 @@ The following supported/unsupported versions cover the OpenTelemetry Collector r
 
 ### Supported versions
 
-Below is a table with documentation for every supported minor release. EOL for the latest release will be three months after the next minor release.
+Below is a table with documentation for every supported Opentelemetry minor release. EOL for the latest release will be three months after the next minor release. The Otel minor version is linked to the latest minor/patch version of the Sumo Logic distribution of the collector.
+Only critical bug fixes and CVEs will be backported to supported versions.
 
 | version                                                                                                 | planned end of life date |
-| ------------------------------------------------------------------------------------------------------- | ------------------------ |
-| [v0.98.0-sumo-0](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.98.0-sumo-0/README.md)   | 2024-08-09               |
-| [v0.97.0-sumo-1](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.97.0-sumo-1/README.md)   | 2024-07-23               |
-| [v0.97.0-sumo-0](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.97.0-sumo-0/README.md)   | 2024-07-23               |
-| [v0.96.0-sumo-1](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.96.0-sumo-1/README.md)   | 2024-07-11               |
-| [v0.96.0-sumo-0](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.96.0-sumo-0/README.md)   | 2024-07-11               |
-| [v0.94.0-sumo-2](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.94.0-sumo-2/README.md)   | 2024-06-12               |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -55,6 +55,32 @@ In order to learn more, pleasee see [purpose of Sumo Logic Distribution for Open
...
+| [v0.97.0-sumo-1](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.97.0-sumo-1/README.md)   | 2024-07-23               |
+| [v0.97.0-sumo-0](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.97.0-sumo-0/README.md)   | 2024-07-23               |
+| [v0.96.0-sumo-1](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.96.0-sumo-1/README.md)   | 2024-07-11               |
+| [v0.96.0-sumo-0](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.96.0-sumo-0/README.md)   | 2024-07-11               |
+| [v0.94.0-sumo-2](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.94.0-sumo-2/README.md)   | 2024-06-12               |
+| [v0.94.0-sumo-1](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.94.0-sumo-1/README.md)   | 2024-06-12               |
+| [v0.94.0-sumo-0](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.94.0-sumo-0/README.md)   | 2024-06-12               |
+| [v0.93.0-sumo-0](https://github.com/SumoLogic/sumologic-otel-collector/blob/v0.93.0-sumo-0/README.md)   | 2024-05-14               |
```
sumo-drosiek, line 76: three months is a lot in terms of OpenTelemetry. Do we really need to support all of these versions where most likely update to the newest version will resolve most of the issues?

Also we shouldn't support `v0.94.0-sumo-1` as `v0.94.0-sumo-2` is the same version of the base collector.

I'm not sure if we put it in documentation but primarily we declared to support only latest version

swiatekm-sumo: In terms of process, it'd be much easier to just say we support the last 6 minor releases. That works out to around 3 months, and we won't need to constantly update this table.

swiatekm-sumo: But we should also clarify what "support" means in this case. I'm not willing to promise we'll backport bug fixes to all of these version, except in very extreme circumstances.

rnishtala-sumo: I defined it this way, because this seems to be recorded in the doc from our last discussion

>We should support upto 3 months older versions. ~6 versions.

I wanted to call out the versions explicitly to reduce ambiguity. If we say the last 6 versions, its possible to have multiple interpretations, for example otel versions vs sumo distribution versions (including patches).

>Also we shouldn't support v0.94.0-sumo-1 as v0.94.0-sumo-2 is the same version of the base collector

makes sense, we haven't talked about how to differentiate a patch from the base version yet (in terms of support). Maybe we don't list the patch version as a separate item and instead provide link the latest patch version in the same row as  the base version. Calling it the "Latest patch version".

Lastly, I'm open to opinions about what "Support" should mean. I think technically it means that we continue releasing patches for that minor version, based on customer issues. 

Patches will especially be useful in the realm of OTRM where the customer wants a new component/configuration supported/validated without risking breaking changes in a later minor/major version.

rnishtala-sumo: This change captures the fact that our releases track upstream versions of the collector.

https://github.com/SumoLogic/sumologic-otel-collector/pull/1583/commits/dbb1daf581081cadd40d64184399a069b7c8f01d

sumo-drosiek (approved):

kkujawa-sumo (approved):

swiatekm-sumo (approved):
Please update the table before merging. We should also add updating it to release instructions.

sumo-drosiek:
According to support poloicy we should backport security fix to all supported versions

echlebek:
@rnishtala-sumo what's the status of this PR? Seems to have fallen by the wayside.

rnishtala-sumo:
@echlebek I believe the final decision was that we're not taking this approach to support multiple prior versions. We're going to stick to the current, which is only support the latest version.

PR #1587, opened 2024-05-20
feat(build): use the builder to set config providers

### LINKED ISSUE #1565: Do we need to have collector_windows.go in this repo?
Do we need to have [collector_windows.go](https://github.com/SumoLogic/sumologic-otel-collector/blob/main/otelcolbuilder/cmd/collector_windows.go) in this repo?
please see this comment: https://github.com/SumoLogic/sumologic-otel-collector/pull/1560#discussion_r1579175092

swiatekm-sumo:
The builder now supports setting config providers and converters via its configuration. Use that instead of our own patches. This change gets rid of a bunch of custom build code. We only keep the flags necessary to check if `--remote-config` and `--config` haven't been set at the same time.

I've also removed our custom Windows service code. The Windows service initialization in core now behaves very similarly to starting the collector normally, so doing the same thing gives the right result. One modification I had to make was deleting the opamp flag from `os.Args`, as the service handler calls `flags.Parse`, which errors on unknown flags, and there isn't a way to add our custom flag there.

Removing the custom windows code makes our custom `--remote-config` flag not appear in the help text when running as a Windows service, but I think that's worth it in return for being able to use the upstream service handler.

Resolves #1565


sumo-drosiek (approved):

otelcolbuilder/cmd/collector_windows.go
```diff
@@ -226,41 +195,3 @@ func withWindowsCore(elog *eventlog.Log) func(zapcore.Core) zapcore.Core {
...
-		// TODO: Remove this after CollectorSettings.ConfigProvider is removed and instead
-		// do it in the builder.
-		if len(resolverSet.ProviderFactories) == 0 && len(resolverSet.ConverterFactories) == 0 {
-			set.ConfigProviderSettings = newDefaultConfigProviderSettings(resolverSet.URIs)
-		}
-	}
-	return nil
-}
-
-func newDefaultConfigProviderSettings(uris []string) otelcol.ConfigProviderSettings {
```
rnishtala-sumo, line 251: this can be removed, since the defaults are now in the builder `otelcol-builder.yaml`?

otelcolbuilder/cmd/configprovider.go
```diff
@@ -45,9 +39,8 @@ func UseCustomConfigProvider(params *otelcol.CollectorSettings) error {
 	if err != nil {
 		return fmt.Errorf("setting feature gate flags failed: %s", err)
 	}
-	// to create the provider, we need config locations passed in via the command line
-	// to get these, we take the command the service uses to start, parse the flags, and read the values
-	flagset := flags(featuregate.GlobalRegistry())
+	// we need to check if any config locations have been set alongside the opamp config
+	flagset := flags()
```
rnishtala-sumo, line 43: is there a reason that we don't need to pass in the featuregate anymore?

otelcolbuilder/cmd/configprovider.go
```diff
@@ -63,53 +56,23 @@ func UseCustomConfigProvider(params *otelcol.CollectorSettings) error {
...
-		// if no locations, use defaults
-		// either this is a command, or the default provider will throw an error
+	if opAmpPath == "" {
+		// let the default config provider handle things, we don't need to do anything
 		return nil
 	}
 
-	// create the config provider using the locations
-	if len(locations) > 0 {
-		params.ConfigProviderSettings = NewConfigProviderSettings(locations)
```
rnishtala-sumo, line 80: don't we need this to support other configproviders if the opamp path isn't set? or is the builder ensuring this is done?

swiatekm-sumo: This was actually already doable before the builder change, we just didn't take advantage of it. The only reason this code was here was to add config locations to the settings, and that's not necessary anymore since https://github.com/open-telemetry/opentelemetry-collector/pull/9228.

swiatekm-sumo: But yes, now we *also* don't need to specify the providers here, they're specified in the `main.go` generated by the builder.

otelcolbuilder/cmd/collector_windows.go
swiatekm-sumo: Hm, now that you mention it, this shouldn't work because it doesn't go through command parsing... But it did  seem to work for me in manual tests. I'll re-check.

otelcolbuilder/cmd/configprovider.go
swiatekm-sumo: We never did on Linux and Mac, because they're parsed again later anyway. But it looks like we still may need to on Windows.

otelcolbuilder/Makefile
```diff
@@ -157,9 +157,6 @@ build-fips:
 generate-sources:
 	@$(MAKE) _builder SKIP_COMPILATION=true
 	# the changes below allow us to use a custom config provider
-	# this requires some additions to the main.go file generated by the builder
-	cd cmd && go get github.com/SumoLogic/sumologic-otel-collector/pkg/configprovider/globprovider@v0.0.0-00010101000000-000000000000
-	cd cmd && go get github.com/SumoLogic/sumologic-otel-collector/pkg/configprovider/opampprovider@v0.0.0-00010101000000-000000000000
```
echlebek, line 162: Very happy to see these lines removed :tada:

swiatekm-sumo:
Converting to draft until I figure out how the service registration should work. I *think* it should just work with the upstream service handler, but that's not actually the case.

otelcolbuilder/cmd/configprovider.go
swiatekm-sumo: Don't need to do it on Windows either. We only need to include flags here whose input we need, which means just the config ones.

fguimond (approved):
Nice cleanup. Also validated that it works as a Windows service.

rnishtala-sumo (approved):

Issue #1573, opened 2024-05-03
cascadingfilterprocessor should support filtering on trace status


aclowes:
The `min_number_of_errors` filter is not useful for accepting traces only when the root span has an error. It would be useful to add a property to the trace like `statuscode` where we can filter to `ERROR`, like in the UI. Thanks.

chan-tim-sumo:
Hi Alec,

Just to make sure, you're talking about filtering the traces by status_code for the root span only right?

aclowes:
> Just to make sure, you're talking about filtering the traces by status_code for the root span only right?

Correct! Thank you.

chan-tim-sumo:
Hey Alec, you can [filter traces by status code for the root span now](https://github.com/SumoLogic/sumologic-otel-collector/tree/main/pkg/processor/cascadingfilterprocessor#just-filtering-out-status-code). Let me know if you have any questions!

aclowes:
Thank you!

PR #1600, opened 2024-06-07
feat(cascadingprocessor): Added status code filtering to cascading filter processor

chan-tim-sumo:
Links:
1. github issue: https://github.com/SumoLogic/sumologic-otel-collector/issues/1573
2. jira: https://sumologic.atlassian.net/browse/OSC-726
3. [cascading filter processor](https://github.com/SumoLogic/sumologic-otel-collector/tree/main/pkg/processor/cascadingfilterprocessor) 

From the github issue, they wanted to be able to filter traces by status code for ROOT span only. Currently the only way to filter is to use [minimum number of errors ](https://github.com/SumoLogic/sumologic-otel-collector/blob/main/pkg/processor/cascadingfilterprocessor/config/config.go#L48) which basically checks the minimum number of errors within a trace, which isn't as great.

For example, if minimum number of errors is set to 1 (this means 1 or more errors): 
Trace 1:
- span1 (root) = OK
- span2 = ERROR
- span3 = OK

Trace 2:
- span1 (root) = ERROR
- span2 = ERROR
- span3 = OK

This would return trace 1 and 2. But what they would want is to accept trace 2 only, with status code filtering, using the same example from above, will return trace 2 only.


pkg/processor/cascadingfilterprocessor/cascade_test.go
```diff
@@ -33,6 +33,8 @@ import (
 var testValue = 10 * time.Millisecond
 var probabilisticFilteringRate = int32(10)
 var healthCheckPattern = "health"
+var statusCode = "ERROR"
```
sumo-drosiek, line 36: I would change it to `ptrace.StatusCodeError.String()` and removed `mapStatusCode`

pkg/processor/cascadingfilterprocessor/sampling/drop_trace_factory.go
```diff
@@ -55,11 +57,16 @@ func NewDropTraceEvaluator(logger *zap.Logger, cfg config.TraceRejectCfg) (DropT
...
+	if cfg.StatusCode != nil {
+		statusCode = cfg.StatusCode
+	}
+
 	return &dropTraceEvaluator{
 		stringAttr:  stringAttrFilter,
 		numericAttr: numericAttrFilter,
 		attrs:       attrsFilter,
 		operationRe: operationRe,
+		statusCode:  statusCode,
```
sumo-drosiek, line 69: I would change it to the following as it sets `nil` if `cfg.StatusCode` is `nil` otherwise the value

```suggestion
		statusCode:  cfg.StatusCode,
```

No need for:

```
var statusCode *string // it sets statusCode to nil

// It sets statusCode to the value if cfg.StatusCode is not nil
if cfg.StatusCode != nil {
  statusCode = cfg.StatusCode
}
```

pkg/processor/cascadingfilterprocessor/sampling/drop_trace_factory.go
```diff
@@ -104,6 +112,15 @@ func (dte *dropTraceEvaluator) ShouldDrop(_ pcommon.TraceID, trace *TraceData) b
 						matchingNumericAttrFound = checkIfNumericAttrFound(span.Attributes(), dte.numericAttr)
 					}
 
+					isRootSpan := span.ParentSpanID().IsEmpty()
+
+					if dte.statusCode != nil && !matchingStatusCodeFound && isRootSpan {
```
sumo-drosiek, line 117:
```suggestion
					if !matchingStatusCodeFound && dte.statusCode != nil && span.ParentSpanID().IsEmpty() {
```

There is no need to run `span.ParentSpanID().IsEmpty()` for every span, so I would move it after checking for status code

sumo-drosiek: Also I moved `!matchingStatusCodeFound` to the beginning as this is more logical order

pkg/processor/cascadingfilterprocessor/cascade_test.go
chan-tim-sumo: done!

pkg/processor/cascadingfilterprocessor/sampling/drop_trace_factory.go
chan-tim-sumo: done!

pkg/processor/cascadingfilterprocessor/sampling/drop_trace_factory.go
chan-tim-sumo: yea that makes sense

sumo-drosiek (approved):
Just one more question. Is `status_code` case sensitive? Is it ok to type `error`, `Error`, `eRRoR`? If not, can we fix the user's input or raise an error?

pkg/processor/cascadingfilterprocessor/sampling/drop_trace_factory.go
```diff
@@ -28,10 +29,27 @@ type dropTraceEvaluator struct {
...
 
 	logger *zap.Logger
 }
 
+func validateStatusCode(statusCode *string) error {
+	if statusCode == nil {
+		return nil
+	}
+
+	validStatusCodes := []string{"Error", "Ok", "Unset"}
```
sumo-drosiek, line 42: We should use constants

chan-tim-sumo: done!

pkg/processor/cascadingfilterprocessor/config_test.go
```diff
@@ -45,6 +45,7 @@ func TestLoadConfig(t *testing.T) {
 	probFilteringRate := int32(100)
 	namePatternValue := "foo.*"
 	healthCheckNamePatternValue := "health.*"
+	statusCode := "Error"
```
sumo-drosiek, line 48:
```suggestion
	statusCode := ptrace.StatusCodeError.String()
```

pkg/processor/cascadingfilterprocessor/sampling/drop_trace_factory.go
```diff
@@ -23,15 +24,38 @@ import (
...
 
 	logger *zap.Logger
 }
 
+func validateStatusCode(statusCode *string) error {
+	if statusCode == nil {
+		return nil
+	}
+
+	validStatusCodes := []string{StatusCodeError, StatusCodeOk, StatusCodeUnset}
```
sumo-drosiek, line 48:
```suggestion
	validStatusCodes := []string{
	  ptrace.StatusCodeError.String(),
	  ptrace.StatusCodeOk.String(),
	  ptrace.StatusCodeUnset.String(),
	}
```

I would also generate error out of this array

chan-tim-sumo: done!

sumo-drosiek: Can we use it in error string as well?

chan-tim-sumo: yep fixed!

pkg/processor/cascadingfilterprocessor/sampling/drop_trace_factory.go
```diff
@@ -28,10 +30,31 @@ type dropTraceEvaluator struct {
...
+		ptrace.StatusCodeUnset.String(),
+	}
+
+	for _, valid := range validStatusCodes {
+		if *statusCode == valid {
+			return nil
+		}
+	}
+
+	return errors.New("invalid status code: must be one of 'Error', 'Ok', or 'Unset' ")
```
sumo-drosiek, line 55:
```suggestion
	return errors.New("invalid status code: must be one of 'Error', 'Ok', or 'Unset'")
```

I was talking about above string 😅
It's up to you. It can go that way 🤗

PR #1604, opened 2024-06-11
feat(ConfigProvider): Modify config merge to do replace instead of update for Specific tags for Remotely managed collectors

jagan2221:
Modifying config merge functionality to do replace for collector_fields tag. Merge behaviour for other fields remains the same existing flow.

Why? 
[Collector tag edit behaviour change requirement](https://docs.google.com/document/d/1U9yb2SAW3Q-c3zWimMwoFRAP0jsdGioijlbcUCHZ4wY/edit#heading=h.eazeox5ci0m
)

Public doc which describes this merge behaviour. [doc link](https://help.sumologic.com/docs/send-data/opentelemetry-collector/data-source-configurations/overview/#configuration-structure)

Note: We are overriding this merge behaviour only for collector_fields tag of Remotely Managed collectors. Config merge flow of Locally managed collectors remain the same existing flow.


pkg/configprovider/opampprovider/provider.go
```diff
@@ -100,6 +100,9 @@ func (p *Provider) Retrieve(ctx context.Context, configPath string, fn confmap.W
 	if err != nil {
 		return nil, err
 	}
+	//Order of conf parameter is imp, see method comments
```
sumo-drosiek, line 103:
```suggestion
	// Order of conf parameters is important, see method comments
```

pkg/configprovider/providerutil/utils.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"go.opentelemetry.io/collector/confmap"
+	"strings"
+)
+
+func removeKeyFromSrcMap(srcMap map[string]interface{}, mergeMap map[string]interface{}, keys []string, index int) map[string]interface{} {
+
+	if index == len(keys) { //out of index
+		return srcMap
+	}
```
sumo-drosiek, line 12: Another approach is to pass `keys[1:]` in the next iteration and if `len(keys) == 1` then it means this is the leaf key. I think it will be maybe less performant

pkg/configprovider/providerutil/utils.go
```diff
@@ -0,0 +1,45 @@
...
+
+	currentKey := keys[index]
+	leafIndex := len(keys) - 1
+	if index == leafIndex { // Got Leaf key, if leaf key exists in both maps, remove it from source map
+		_, existInSrc := srcMap[currentKey]
+		_, existInMerge := mergeMap[currentKey]
+		// If leaf key exists in both maps, remove from source map
+		if existInSrc && existInMerge {
+			delete(srcMap, currentKey)
+		}
```
sumo-drosiek, line 22: We can `return` here and then we can remove `else` statement (unnecessary nesting lowers readability)

pkg/configprovider/providerutil/utils.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// Checks and prepares scrConf for replace behavior instead of map merge
+// Hack for confmap.Conf.Merge method to replace specific fields instead of merging them
+// Merge method merges field values from source and mergeConf, so by removing existing values from
+// source map, we can achieve replace behavior
+func PrepareForReplaceBehavior(srcConf *confmap.Conf, mergeConf *confmap.Conf) {
+	keyPathsWithReplaceBehavior := []string{
+		"extensions#sumologic#collector_fields",
+	}
```
sumo-drosiek, line 40: We can use map instead to avoidf any separator. As far as it's not configurable, we shouldn't complicate it, and if we will expose it to the configueration we can simply convert it:

```suggestion
	keyPathsWithReplaceBehavior := [][]string{
		{"extensions", "sumologic", "collector_fields"},
	}
```

pkg/configprovider/providerutil/utils.go
jagan2221: It seems only references are passed around when using keys[1:], so no actual performance impact, using this method as we don't have to pass around indexes

pkg/configprovider/providerutil/utils_test.go
```diff
@@ -0,0 +1,95 @@
...
+					"sumologic": map[string]interface{}{
+						"collector_fields1": map[string]interface{}{
+							"cluster": "cluster-1",
+						},
+					},
+				},
+				"anotherKey": "anotherValue",
+			},
+		},
+	}
```
sumo-drosiek, line 83: Why `collector_fields1` is not in the expected map?

sumo-drosiek: Ok, it's only about the first map 😅

pkg/configprovider/providerutil/utils_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			srcConf := confmap.NewFromStringMap(tt.srcMap)
+			mergeConf := confmap.NewFromStringMap(tt.mergeMap)
+			PrepareForReplaceBehavior(srcConf, mergeConf)
+			if !reflect.DeepEqual(srcConf.ToStringMap(), tt.expectedMap) {
+				t.Errorf("PrepareForReplaceBehavior() = %v, want %v", srcConf.ToStringMap(), tt.expectedMap)
+			}
```
sumo-drosiek, line 92: Can we use `assert.Equal` here?

pkg/configprovider/providerutil/utils_test.go
```diff
@@ -0,0 +1,95 @@
...
+				},
+				"anotherKey": "anotherValue",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			srcConf := confmap.NewFromStringMap(tt.srcMap)
+			mergeConf := confmap.NewFromStringMap(tt.mergeMap)
```
sumo-drosiek, line 88: Are those conversions necessary?

jagan2221: We can avoid conversions by storing test data as ConfMap , will do that

sumo-drosiek:
LGTM, but we should mark this as breaking change and explain in https://github.com/SumoLogic/sumologic-otel-collector/blob/main/docs/upgrading.md how to keep old behavior (by moving all `collector_fields` to one (the latest in order) config file)

.changelog/1604.changed.txt
```diff
@@ -0,0 +1 @@
+feat(ConfigProvider): Modify config merge to do replace instead of update for Specific tags
```
sumo-drosiek, line 1:
```suggestion
feat(ConfigProvider)!: Modify config merge to do replace instead of update for Specific tags
```

and also please rename file to `1604.breaking.txt`

jagan2221:
https://github.com/SumoLogic/sumologic-otel-collector/blob/j_v103_0_upgrading_md/docs/upgrading.md. - Added notes for breaking changes in this separate branch, we can use it while preparing next release

sumo-drosiek:
> https://github.com/SumoLogic/sumologic-otel-collector/blob/j_v103_0_upgrading_md/docs/upgrading.md. - Added notes for breaking changes in this separate branch, we can use it while preparing next release

I believe it should be done in this PR

docs/upgrading.md
```diff
@@ -53,6 +55,79 @@
...
+    collector_description: "My OpenTelemetry Collector"
+    collector_fields:
+      zone: "eu"
+    some_list:
+      - element 3
+      - element 4
+```
+
+If you have multiple config files with `collector_fields` key specified, only the value from the last file(**alphabetically** sorted order) will be present in effective configuration.
+So, for locally managed collectors, Please don't maintain `collector_fields` key in multiple configuration files and move them to a single file.
```
sumo-drosiek, line 125:
```suggestion
Due to above, avoid maintaining `collector_fields` in multiple configuration files and move them to a single file.
```

sumo-drosiek (commented):
LGTM, but please get another approval from @SumoLogic/open-source-collection-team

pkg/configprovider/providerutil/utils.go
```diff
@@ -0,0 +1,43 @@
...
+		if existInSrc && existInMerge {
+			delete(srcMap, currentKey)
+		}
+		return srcMap
+	}
+	// More levels to go, descend into child if current key present in both maps
+	srcNestedMap, isCurrKeyInSrcMap := srcMap[currentKey].(map[string]interface{})
+	mergeNestedMap, isCurrKeyInMergeMap := mergeMap[currentKey].(map[string]interface{})
+	if isCurrKeyInSrcMap && isCurrKeyInMergeMap {
+		removeKeyFromSrcMap(srcNestedMap, mergeNestedMap, keys[1:])
```
chan-tim-sumo, line 27: is recursion safe to do here?

docs/upgrading.md
```diff
@@ -53,6 +55,79 @@
...
+    collector_description: "My OpenTelemetry Collector"
+    collector_fields:
+      cluster: "cluster-1"
+      zone: "eu"
+    some_list:
+      - element 3
+      - element 4
+```
+
+effective configuration (`new behaviour`)
```
rnishtala-sumo, line 111: is there an issue describing the need for the new behavior? If so please link the issue to the PR. If its a Jira issue then let's maybe describe the requirement in the PR.

pkg/configprovider/providerutil/utils.go
jagan2221: Yes @chan-tim-sumo , we have base condition to end the recursive call, Also only when both maps have the field path, the call will proceed.

docs/upgrading.md
jagan2221: https://docs.google.com/document/d/1U9yb2SAW3Q-c3zWimMwoFRAP0jsdGioijlbcUCHZ4wY

This change was done as part of this feature.

pkg/configprovider/providerutil/utils.go
rnishtala-sumo: Why do we need to implement recursion here? isn't `collector_fields` just a list of tags?

pkg/configprovider/providerutil/utils.go
```diff
@@ -0,0 +1,43 @@
+package providerutil
+
+import (
+	"go.opentelemetry.io/collector/confmap"
+)
+
+func removeKeyFromSrcMap(srcMap map[string]interface{}, mergeMap map[string]interface{}, keys []string) map[string]interface{} {
```
rnishtala-sumo, line 7: We should add a comment describing what this method does.

pkg/configprovider/providerutil/utils.go
jagan2221: We need to traverse down the map(tree structure) according to the tags. Either we need to use loop/recursion @rnishtala-sumo

chan-tim-sumo: on line 23, your comment says:
```
// More levels to go, descend into child if current key present in both maps
```

at the bottom of this doc: https://help.sumologic.com/docs/send-data/opentelemetry-collector/data-source-configurations/overview/#configuration-structure

it seems like the map is one level deep, there aren't nested values:

```
collector_fields:
  cluster: "cluster-1"
  zone: "eu"
```

If tags can be more than one level deep, then I don't think recursion should be used, doing this iteratively would be better imo.

I might be missing something here, but can tags be more than one level deep? can you provide an example?

jagan2221: @chan-tim-sumo 
https://github.com/SumoLogic/sumologic-otel-collector/blob/main/examples/sumologic.yaml#L55

There is no limitation on the number of levels, it can go deeper according to the use case. Above config goes processors#resourcedetection/system#system#hostname_sources - 4 levels

The method(removeKeyFromSrcMap) is now generic to handle any key at any level deep.

echlebek: @chan-tim-sumo The Go data type for `collector_fields` is `map[string]any`, so the values can be of any data type. The document is somewhat trusted, as it's owned by root, so I don't think stack overflow is a huge concern here.

echlebek:
Please make sure you format with gofmt.

docs/upgrading.md
```diff
@@ -81,6 +83,79 @@ to this:
       compression: ""
   ```
 
+## Upgrading to v0.103.0-sumo-0
```
sumo-drosiek, line 86: should be in 0.104.0 section

rnishtala-sumo:
@jagan2221 Couple of questions - 
- which option from the design doc is this PR implementing?
- My understanding is that this config merge change is only limited to collector fields and shouldn't affect the other fields?

pkg/configprovider/providerutil/utils_test.go
```diff
@@ -0,0 +1,91 @@
...
+						},
+						"childKey2": "value2",
+					},
+				},
+				"anotherKey": "anotherValue",
+			}),
+			expectedConf: confmap.NewFromStringMap(map[string]interface{}{
+				"extensions": map[string]interface{}{
+					"sumologic": map[string]interface{}{
+						"childKey2": "value2",
```
rnishtala-sumo, line 44: why isn't "zone": "eu" the  expected field here?

pkg/configprovider/providerutil/utils_test.go
```diff
@@ -0,0 +1,91 @@
...
+							"cluster": "cluster-1",
+						},
+					},
+				},
+				"anotherKey": "anotherValue",
+			}),
+		},
+	}
+
+	for _, tt := range tests {
```
rnishtala-sumo, line 85: Could we add a test for non-collector-tag fields to demonstrate that other config fields aren't impacted by this change? i.e no regression.

pkg/configprovider/providerutil/utils.go
```diff
@@ -0,0 +1,44 @@
...
+		if existInSrc && existInMerge {
+			delete(srcMap, currentKey)
+		}
+		return srcMap
+	}
+	// More levels to go, descend into child if current key present in both maps
+	srcNestedMap, isCurrKeyInSrcMap := srcMap[currentKey].(map[string]interface{})
+	mergeNestedMap, isCurrKeyInMergeMap := mergeMap[currentKey].(map[string]interface{})
+	if isCurrKeyInSrcMap && isCurrKeyInMergeMap {
+		removeMatchingKeyFromSrcMap(srcNestedMap, mergeNestedMap, keys[1:])
```
rnishtala-sumo, line 28: is this effectively removing the first entry from a map? How are we ensuring that the keys are in the right order, i.e sorted by their file name?

jagan2221:
> @jagan2221 Couple of questions -
> 
> * which option from the design doc is this PR implementing?
        The part where opamp agent merges incoming config messages/stored config files to form effective configuration
> * My understanding is that this config merge change is only limited to collector fields and shouldn't affect the other fields?
       Yes, that's correct. We have modifed the merge behaviour of collectror_fields tag only. Others remain unaffected.

jagan2221:
> > @jagan2221 Couple of questions -
> > 
> > * which option from the design doc is this PR implementing?
> >   The part where opamp agent merges incoming config messages/stored config files to form effective configuration
> > * My understanding is that this config merge change is only limited to collector fields and shouldn't affect the other fields?
> >   Yes, that's correct. We have modifed the merge behaviour of collectror_fields tag only. Others remain unaffected.

@rnishtala-sumo

pkg/configprovider/providerutil/utils_test.go
jagan2221: Have added just some sample test values. Do you think we should have specific values here?

rnishtala-sumo:
>The part where opamp agent merges incoming config messages/stored config files to form effective configuration

@jagan2221 I mean which option from the design doc?

pkg/configprovider/providerutil/utils_test.go
jagan2221: https://github.com/SumoLogic/sumologic-otel-collector/blob/3e13dd3d4dd3a3b88ebd22b6c8988452123203fb/pkg/configprovider/providerutil/utils_test.go#L51

It is covered here. @rnishtala-sumo

pkg/configprovider/providerutil/utils_test.go
rnishtala-sumo: I mean, I don't understand the assertion here. My questions are
- why doesn't the expected config have any collector_fields.
- why are we removing `"cluster": "cluster-1"` from the src/original conf? and why isn't `collector_fields` from `mergeConf` in the expected result?

jagan2221:
> Collector tag edit behaviour change requirement

https://docs.google.com/document/d/1U9yb2SAW3Q-c3zWimMwoFRAP0jsdGioijlbcUCHZ4wY/edit#heading=h.eazeox5ci0m
@rnishtala-sumo Exactly this requirement section from design doc

pkg/configprovider/providerutil/utils_test.go
rnishtala-sumo: @jagan2221 would appreciate clarification here, maybe I'm missing something?

jagan2221:
> @jagan2221 is this PR going to change the default behavior? If so, could we add this behind a feature flag for the next release? We could keep the default behavior the same.

It only affected "collector_fields" key specifically, i think there won't be real use case where currently customer's use this key in multiple config files for locally managed collectors. Usually for locally managed collectors , customer's maintain this in sumologic.yaml right, so i feel there won't be a case for having this in another config file. 
And we have explained this as a breaking change for corner case in release notes too. Your thoughts on this ?@rnishtala-sumo

rnishtala-sumo:
@jagan2221 I think since this affects config merge behavior, its a big enough change to be behind a feature flag for the next release, where the customer needs to set the flag to use this feature. Then eventually we could let this be the default.

jagan2221:
@rnishtala-sumo This is a feature which needs to be enabled by default for collector tag edit to work. If we make customers to enable this, then we need to rely on this flag status for tag edit feature to work/shown in UI. 
Like we are enabling this feature in UI and rely on another flag in agent to enable it.

pkg/configprovider/providerutil/utils_test.go
chan-tim-sumo: agreed with Raj, the expected config should have something in ```collector_fields```. i remember there was a sumo documentation showing the merge functionality, think it was this: 

https://help.sumologic.com/docs/send-data/opentelemetry-collector/data-source-configurations/overview/#configuration-structure:~:text=Note%20that%20the%20list%20has%20been%20overridden%20and%20maps%20have%20been%20merged.

rnishtala-sumo: @jagan2221 shouldn't the `collector_fields` in the expected config have `"zone": "eu"` since we now replace the existing tag instead of appending to it? Also, am I right in understanding that `mergeConf` is simply a second configuration file?

I have listed the expected config below

Original behavior

Expected config
```
collector_fields:
  "cluster": "cluster-1"
  "zone": "eu"
```

New (Replace) behavior

Expected config
```
collector_fields:
  "zone": "eu"
```

what am I missing here?

chan-tim-sumo: oh after staring at this for a bit, think I know why the expected config is this:
```
expectedConf: confmap.NewFromStringMap(map[string]interface{}{
    "extensions": map[string]interface{}{
	    "sumologic": map[string]interface{}{
		    "childKey2": "value2",
	    },
    },
    "anotherKey": "anotherValue",
}),
```

since Jagan mentioned:

> We are not testing the merging part, The func PrepareForReplaceBehaviour prepares sourceConf for replace behaviour, ie) If sourceConf has collector_fields key and merge conf also has the same , then collector_fields key will be removed from sourceConf.

> Actual merge flow remains the same, we are just overriding sourceConf(by removing collector_fields key based on condition that mergeConf also has it), so that existing merge will add collector_fields from mergeConf into expectedConf

this is testing ONLY the replacing functionality and not the merging. if so, then this test makes sense. what made it confusing for me was because I'm assuming this would also test for the merging functionality. 

the merging logic happens in [globprovider](https://github.com/SumoLogic/sumologic-otel-collector/blob/3e13dd3d4dd3a3b88ebd22b6c8988452123203fb/pkg/configprovider/globprovider/provider.go#L71-L72)

now i'm wondering should a new test be added in [globprovider test](https://github.com/SumoLogic/sumologic-otel-collector/blob/3e13dd3d4dd3a3b88ebd22b6c8988452123203fb/pkg/configprovider/globprovider/provider_test.go) that will specifically test the whole flow of replacing to merging flow.

rnishtala-sumo: I see, I wonder why none of the existing tests failed, with this change in merge behavior. But I tend to agree, I think we need a test for the whole flow including the merge. I'd rather be too careful for a feature that potentially impacts all of our configuration than otherwise. Thanks for being patient.

jagan2221: Added test case for whole merge flow with enable/disable new merge behaviour in globprovider test
@rnishtala-sumo @chan-tim-sumo

pkg/configprovider/globprovider/provider.go
```diff
@@ -22,27 +22,33 @@ import (
...
+	"gopkg.in/yaml.v3"
 )
 
 const (
 	schemeName   = "glob"
 	schemePrefix = schemeName + ":"
 )
 
-type provider struct{}
+type Provider struct{
```
rnishtala-sumo, line 35: Why do we need this to be public?

pkg/configprovider/opampprovider/provider.go
```diff
@@ -84,6 +85,10 @@ func (p *Provider) Retrieve(ctx context.Context, configPath string, fn confmap.W
 	}
 	conf := confmap.New()
 	glob := p.GlobProvider
+	if globProvider, ok := glob.(*globprovider.Provider); ok {
```
rnishtala-sumo, line 88: why do we need this condition? wouldn't it suffice to call `SetRemotelyManagedMergeFlow`?

jagan2221: https://github.com/SumoLogic/sumologic-otel-collector/blob/j_config_merge_override/pkg/configprovider/opampprovider/provider.go#L60

We have referred GlobProvider in opamp provider as confmap.Provider interface, so this interface don't directly know about SetRemotelyManagedMergeFlow method. So we are re-referencing it to globProvider.Provider in this part.
Alternately we can make GlobProvider of type globprovider.Provider instead of confmap.provider but it will make opamp provider tightly coupled with a concrete type, so had this type check for additional methods.

pkg/configprovider/globprovider/provider.go
jagan2221: https://github.com/SumoLogic/sumologic-otel-collector/blob/j_config_merge_override/pkg/configprovider/opampprovider/provider.go#L88

To do the type casting here.

pkg/configprovider/globprovider/provider.go
```diff
@@ -68,18 +74,21 @@ func (fmp *provider) Retrieve(ctx context.Context, uri string, _ confmap.Watcher
 			return &confmap.Retrieved{}, err
 		}
 		pathConf := confmap.NewFromStringMap(rawConf)
+		if fmp.remotelyManagedMergeFlow {
```
rnishtala-sumo, line 77: Would it simpler to pass this in as an argument to `Retrieve`? Not sure it makes sense to add the RemotelyManaged boolean to `Provider`. Also, would it benefit us to make this boolean configurable by the user?

pkg/configprovider/opampprovider/provider.go
rnishtala-sumo: Please see this comment - https://github.com/SumoLogic/sumologic-otel-collector/pull/1604/files#r1868090338

pkg/configprovider/globprovider/provider.go
jagan2221: Retrieve is an interface function provided by confmap and i think that overriding it would introduce changes in more places. Also since there is no default parameters in go, we are introducing changes and forcing to pass this flag in all function calls to Retrieve method.

rnishtala-sumo: ok makes sense, how about making boolean configurable in the collector (opamp) config? Do we need a new boolean or could we use an existing opamp config param to set trigger the remoteMergeFlow?

jagan2221: There is no use case to make this configurable externally. We are just controlling how GlobProvider behaves for Locally managed and remotely managed collectors.

rnishtala-sumo (changes requested):
@jagan2221 PTAL this comment - https://github.com/SumoLogic/sumologic-otel-collector/pull/1604#discussion_r1868211712

pkg/configprovider/globprovider/provider.go
jagan2221: Added the flag in opamp config(default enabled, flag can be used to disable the new merge flow).
Added UT's in opampprovider as well for testing the flag toggled.

pkg/configprovider/opampprovider/testdata/config_new_merge_disabled.yaml
```diff
@@ -0,0 +1,5 @@
+extensions:
+  opamp:
+    endpoint: "wss://example.com/v1/opamp"
+    remote_configuration_directory: ../globprovider/testdata/mergefunc
+    new_configmergeflow_disabled: true
```
rnishtala-sumo, line 5: I think this field could be named - `disable_tag_config_merge` or something similar to better represent it. Also need a documentation update for it along with the other params.

pkg/extension/opampextension/README.md
```diff
@@ -31,6 +31,9 @@ It manages:
   configuration received from the OpAMP server. This directory must coincide
   with a configuration provider (e.g. glob) for the configuration to be loaded
   by the collector.
+- `disable_tag_new_config_merge`: (optional) Boolean flag to disable new config merge
+  flow introduced for tag edit feature for remotely managed collectors. Default
+  value is disabled.
```
rnishtala-sumo, line 36: Nit: default value - false.

Also, how does this name sound - `disable_tag_replacement`. Since we're replacing tags (under collector_fields) correct?

pkg/configprovider/opampprovider/provider.go
```diff
@@ -38,6 +39,7 @@ type ConfigFragment struct {
 	Extensions struct {
 		OpAmp struct {
 			RemoteConfigurationDirectory string `yaml:"remote_configuration_directory"`
+			NewConfigMergeFlowDisabled bool `yaml:"disable_tag_new_config_merge"`
```
chan-tim-sumo, line 42: In provider.go, since the name got changed to ```disable_tag_new_config_merge```, shouldn't the variable also be changed to DisableTagNewConfigMerge (next to bool)? Just for consistency 😆 

```
	Extensions struct {
		OpAmp struct {
			RemoteConfigurationDirectory string `yaml:"remote_configuration_directory"`
			NewConfigMergeFlowDisabled bool `yaml:"disable_tag_new_config_merge"`
		} `yaml:"opamp"`
	} `yaml:"extensions"`
```

pkg/extension/opampextension/README.md
jagan2221: In general, we are introducing a new config merge flow and tag replacement is a more concrete term. In future, we may have this different merge flow for other parameters as well if required. so i think the general term would be the new config merge flow.

pkg/configprovider/opampprovider/provider.go
jagan2221: Done @chan-tim-sumo

pkg/extension/opampextension/README.md
rnishtala-sumo: "New config merge" doesn't mean much to the customer and I don't think we'll be doing this for other collector fields, since that would most likely break collector configuration. From the customer's point of view it helps to be specific about how the field changes their config. If it replaces their tags, that's what the config field could say and if disables tag replacement the field could be called the same.

jagan2221: Changed the param name @rnishtala-sumo

pkg/extension/opampextension/README.md
```diff
@@ -31,6 +31,9 @@ It manages:
   configuration received from the OpAMP server. This directory must coincide
   with a configuration provider (e.g. glob) for the configuration to be loaded
   by the collector.
+- `disable_tag_replacement`: (optional) Boolean flag to disable new config merge
+  flow which replaces tags instead of updating them introduced for tag edit
```
rnishtala-sumo, line 35: to disable new config merge flow which replaces tags instead of updating them introduced for tag edit feature for remotely managed collectors
```suggestion
to disable tag replacement which replaces tags instead of appending to them for remotely managed collectors
```

pkg/extension/opampextension/README.md
rnishtala-sumo: One more comment, thanks for your patience.

pkg/configprovider/globprovider/testdata/mergefunc/a_srcconf.yaml
```diff
@@ -0,0 +1,6 @@
+extensions:
+  sumologic:
+    childKey: "value"
+    collector_fields:
```
rnishtala-sumo, line 4: I wanted to clarify one last thing - if `collector_fields` was renamed to `collector_fields_1`,  my understanding is that the tags would be appended and not replaced. Could we have a test demonstrating this as well?

jagan2221: Done

pkg/configprovider/providerutil/utils.go
```diff
@@ -0,0 +1,44 @@
...
+	return srcMap
+}
+
+// Checks and prepares scrConf for replace behavior instead of map merge
+// Hack for confmap.Conf.Merge method to replace specific fields instead of merging them
+// Merge method merges field values from source and mergeConf, so by removing existing values from
+// source map, we can achieve replace behavior
+func PrepareForReplaceBehavior(srcConf *confmap.Conf, mergeConf *confmap.Conf) {
+	keyPathsWithReplaceBehavior := [][]string{
+		{"extensions", "sumologic", "collector_fields"},
```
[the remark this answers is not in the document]
jagan2221: Done

pkg/configprovider/globprovider/testdata/mergefunc/a_srcconf.yaml
jagan2221: Added collector_fields1 along with collector_fields and test will check collector_fields1 will have append behaviour 
 always.

PR #1607, opened 2024-06-20
Add otelcol-config

echlebek:
This commit adds otelcol-config, a tool designed to be used for configuration otelcol-sumo installations.

The tool will mainly write values to conf.d/00-otelcol-config-settings.yaml, conf.d/99-otelcol-config-overrides.yaml, and link and unlink specific configuration files in conf.d from conf.d/available.


sumo-drosiek:
Awesome! Do we consider to move other functionality from the install script as well? (checking otel version, verifying what is the latest one and so on?)

swiatekm-sumo:
Shouldn't this live in the packaging repo?

amdprophet:
> Awesome! Do we consider to move other functionality from the install script as well? (checking otel version, verifying what is the latest one and so on?)

I think this falls outside of the existing scope of `otelcol-config`. We're moving away from using curl in the shell script all together and instead will be using packaging repositories where the latest version can be determined from package managers (e.g. yum/dnf, apt, etc.)



> Shouldn't this live in the packaging repo?

The `otelcol-config` binary will be packaged with the `otelcol-sumo` binary. Its features are tightly coupled with the features of `otelcol-sumo`. I believe it's best to keep it in the collector repository as a result and use the same versioning as the `otelcol-sumo` binary.

chan-tim-sumo (approved):

PR #1619, opened 2024-07-10
chore: upgrade otel core to 0.104.0

amdprophet:
Upgrades core & contrib to v0.104.0, updates the Makefile to distinguish between stabilized and unstabilized OTC Go modules.


Makefile
```diff
@@ -16,6 +16,47 @@ else
 SED ?= sed
 endif
 
+# A list of "unstable" OT Collector Go modules in which the update-ot task will
```
sumo-drosiek, line 19: Wouldn't be a more logical to have list of "stable" OT Collector Go modules? If we miss something on the list it will be considered unstable.

Also do we need this hack at all? As I see we provide old and new version to the script, so if we pass `0.103.0` and `0.104.0`, we won't replace `0.10.0` to `0.104.0`

amdprophet: I've removed this code for now as it's not critical to finish for this release. We can look into a longer term solution separately.

sumo-drosiek (approved):
LGTM

PR #1621, opened 2024-07-12
Add --set-installation-token to otelcol-config

echlebek:
This commit adds the set-installation-token action.


chan-tim-sumo (approved):
:shipit:

PR #1625, opened 2024-07-16
Add --set-opamp-endpoint to otelcol-config

echlebek:
This commit adds the set-opamp-endpoint action to otelcol-config. The set-opamp-endpoint action sets the OpAmp endpoint in sumologic-remote.yaml. If sumologic-remote.yaml does not exist, then an error is printed. The action also validates that the URL being set has a valid websocket scheme.

### DIFF
--- a/pkg/tools/otelcol-config/opamp.go
+++ b/pkg/tools/otelcol-config/opamp.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"net/url"
+
+	"github.com/mikefarah/yq/v4/pkg/yqlib"
+)
+
+var errRemoteControlNotEnabled = errors.New("remote control not enabled")
+
+func SetOpAmpEndpointAction(ctx *actionContext) error {
+	endpoint := ctx.Flags.SetOpAmpEndpoint
+	if err := validateEndpoint(endpoint); err != nil {
+		return err
+	}
+	conf, err := ReadConfigDir(ctx.ConfigDir)
+	if err != nil {
+		return err
+	}
+
+	writer := ctx.WriteSumologicRemote
+	doc := conf.SumologicRemote
+
+	if len(doc) == 0 {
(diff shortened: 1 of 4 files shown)


pkg/tools/otelcol-config/opamp.go
```diff
@@ -0,0 +1,58 @@
...
+	if err := validateEndpoint(endpoint); err != nil {
+		return err
+	}
+	conf, err := ReadConfigDir(ctx.ConfigDir)
+	if err != nil {
+		return err
+	}
+
+	writer := ctx.WriteSumologicRemote
+	doc := conf.SumologicRemote
```
ccressent, line 24: Nitpick: not 100% sure about the name of `doc`

pkg/tools/otelcol-config/opamp.go
```diff
@@ -0,0 +1,58 @@
...
+
+	encoder := yqlib.YamlFormat.EncoderFactory()
+	decoder := yqlib.YamlFormat.DecoderFactory()
+	eval := yqlib.NewStringEvaluator()
+
+	expression := fmt.Sprintf(".extensions.opamp.endpoint = %q", endpoint)
+	result, err := eval.EvaluateAll(expression, string(doc), encoder, decoder)
+	if err != nil {
+		return fmt.Errorf("couldn't write installation token: %s", err)
+	}
```
ccressent, line 38: Are you sure this is the correct error message?

ccressent (commented):
Looking good; just see my comment about what I think is the wrong error message.

pkg/tools/otelcol-config/opamp.go
echlebek: It's a yaml doc

pkg/tools/otelcol-config/opamp.go
echlebek: Nope, this is incorrect, thanks!

ccressent (approved):

PR #1628, opened 2024-07-17
Add --add-tag to otelcol-config

echlebek:
This commit adds the add-tag action to otelcol-config. --add-tag adds a tag to either 00-otelcol-config-settings.yaml or
99-otelcol-config-overrides.yaml, depending on whether or not the --override flag is set.

Tag values can be any valid yaml message.


pkg/tools/otelcol-config/add_tag.go
```diff
@@ -0,0 +1,77 @@
...
+			continue
+		}
+		result, err := eval.EvaluateAll(expression, string(doc), encoder, decoder)
+		if err != nil {
+			// perhaps the value needs to be quoted, try again with the value quoted.
+			expression = fmt.Sprintf(quoteKeyFmt, tagName, tag)
+			var qerr error
+			result, qerr = eval.EvaluateAll(expression, string(doc), encoder, decoder)
+			if qerr != nil {
+				return fmt.Errorf("can't add tag %s: error evaluating yq expression: %s", tagName, err)
```
ccressent, line 65: I think that's very nice for user friendliness!

Don't hesitate to abstract that into a separate function (e.g. `QuoteAndEvaluateAll()` or something) if you use it again elsewhere as I suspect you might. Ditto for `EvaluateNew()` above.

ccressent (approved):

pkg/tools/otelcol-config/add_tag.go
echlebek: I suspect this is the only place it will be used, as for write-kv we expect a fully formed yq expression. But I will definitely factor it out if it comes up elsewhere.

PR #1629, opened 2024-07-17
Add --delete-tag to otelcol-config

echlebek:
This commit adds the delete-tag action to otelcol-config.

The delete-tag action deletes a key out of the collector_fields object in conf.d settings. If the --override flag is provided, then the action deletes the tag out of conf.d overrides.

If a user has manually added a tag to conf.d, and tries to delete it with this tool in override mode, then an error will be returned, and otelcol-config will refuse to proceed.

### DIFF
--- a/pkg/tools/otelcol-config/delete_tag.go
+++ b/pkg/tools/otelcol-config/delete_tag.go
@@ -1,0 +1,97 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"sort"
+
+	"github.com/mikefarah/yq/v4/pkg/yqlib"
+)
+
+// DeleteTagAction deletes a collector tag from conf.d.
+// If the --override flag is present, it will delete the tag from both the
+// overrides file and the settings file. If the --override flag is present and
+// the tag exists in a user-controlled file, an error will be returned.
+func DeleteTagAction(ctx *actionContext) error {
+	conf, err := ReadConfigDir(ctx.ConfigDir)
+	if err != nil {
+		return err
+	}
+
+	if conf.SumologicRemote != nil {
+		return errors.New("delete-tag not supported for remote-controlled collectors")
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/tools/otelcol-config/delete_tag.go
```diff
@@ -0,0 +1,97 @@
...
+		return deleteTag(ctx, conf, ConfDSettings, ctx.WriteConfD)
+	}
+}
+
+func deleteTagOverride(ctx *actionContext, conf ConfDir) error {
+	readOrder := sort.StringSlice{}
+	for key := range conf.ConfD {
+		readOrder = append(readOrder, key)
+	}
+	sort.Sort(sort.Reverse(readOrder))
```
ccressent, line 37: I'm not 100% sure why we need to sort here?

ccressent (approved):

pkg/tools/otelcol-config/delete_tag.go
chan-tim-sumo: https://help.sumologic.com/docs/send-data/opentelemetry-collector/data-source-configurations/overview/#configuration-structure 

```
If there are any conflicts between the files, the last loaded configuration file will take precedence.
```

i think its so it gets the last loaded config? 🤔

echlebek: It's actually not strictly required to visit these files in order in this specific case, this is copypasta from read-kv :laughing:

echlebek: But yes, there is an order of precedence that needs to be considered here sometimes.

echlebek: I think I'm going to leave it sorted, so that errors are stable on subsequent runs in the case of there being more than one error in the corpus.

PR #1631, opened 2024-07-22
[OSC-829] Refactoring unit test

harshshahsumo:
Refactoring unit test. Link to the ticket https://sumologic.atlassian.net/browse/OSC-829


pkg/extension/opampextension/opamp_agent_test.go
```diff
@@ -31,12 +31,36 @@ import (
 	semconv "go.opentelemetry.io/collector/semconv/v1.18.0"
 )
 
-func TestNewOpampAgent(t *testing.T) {
-	cfg := createDefaultConfig()
+func setupOpampAgent(t *testing.T) (*opampAgent, *Config) {
```
rnishtala-sumo, line 34: I believe the goal here is to convert this to be parameterized tests in go - https://stackoverflow.com/questions/72856565/go-golang-test-parameterized

sumo-drosiek: Yes, we should follow the manner: https://github.com/SumoLogic/sumologic-otel-collector/blob/c7417c6cbeadbe7e79b3c6c2b84038c1ee30b8af/pkg/exporter/sumologicexporter/config_test.go#L12-L111

so, we have the `testcases` struct with parameters and we iterate over it, so instead of having 5 different `Test` functions, we have just 5-element list

pkg/extension/opampextension/opamp_agent_test.go
```diff
@@ -159,435 +49,482 @@ func TestApplyRemoteConfig(t *testing.T) {
...
+		validate func(*opampAgent, *testing.T)
+	}{
+		{
+			name: "NewOpampAgent",
+			setup: func() (*Config, extension.Settings) {
+				cfg := createDefaultConfig()
+				set := extensiontest.NewNopSettings()
+				set.BuildInfo = component.BuildInfo{Version: "test version", Command: "otelcoltest"}
+				return cfg.(*Config), set
+			},
```
sumo-drosiek, line 68: I see the setup is the same or almost the same across many tests? Can we unify it, so we put it into test body and not as test argument?

If there are more groups (2) we can split them to two tests

sumo-drosiek:
I am have an impression that you are trying to put all tests into one function, but not sure if thats possible. The aim of the issue was to group the tests such way, we avoid copying code. In current shape, we still copy code, just we do it in the struct instead of having separate functions

pkg/extension/opampextension/opamp_agent_test.go
```diff
@@ -159,435 +49,482 @@ func TestApplyRemoteConfig(t *testing.T) {
...
+				changed, err := o.applyRemoteConfig(rc)
+				assert.NoError(t, err)
+				assert.True(t, changed)
+				assert.NotEqual(t, len(o.effectiveConfig), 0)
+
+				o.cfg.AcceptsRemoteConfiguration = false
+				changed, err = o.applyRemoteConfig(rc)
+				assert.False(t, changed)
+				assert.Error(t, err)
+				assert.Equal(t, "OpAMP agent does not accept remote configuration", err.Error())
```
chan-tim-sumo, line 203: seems like there're more than 2 duplicates, should this be some type of function? 🤔

chan-tim-sumo:
should this be broken down into multiple (2) PRs? kinda hard to see whats being changed/refactored here 🤔

pkg/extension/opampextension/opamp_agent_test.go
```diff
@@ -159,435 +49,423 @@ func TestApplyRemoteConfig(t *testing.T) {
...
+				path := filepath.Join("testdata", "opamp.d", "opamp-host-config.yaml")
+				rc := remoteConfig(t, path)
+
+				changed, err := o.applyRemoteConfig(rc)
+				assert.NoError(t, err)
+				assert.True(t, changed)
+				assert.NotEqual(t, len(o.effectiveConfig), 0)
+
+				o.cfg.AcceptsRemoteConfiguration = false
+				changed, err = o.applyRemoteConfig(rc)
```
chan-tim-sumo, line 126: looks like alot of the ApplyRemote tests are repeated. i feel like the test should look something like this (didn't include all the different config yaml files and a little more adjustments) 

```
func TestApplyRemoteConfig(t *testing.T) {
    tests := []struct {
        name           string
        file           string
        acceptsRemote  bool
        expectChanged  bool
        expectError    bool
        errorMessage   string
    }{
        {"Valid Config", "testdata/opamp.d/opamp-remote-config.yaml", true, true, false, ""},
        {"Does Not Accept Remote", "testdata/opamp.d/opamp-apache-config.yaml", false, false, true, "OpAMP agent does not accept remote configuration"},
        {"Invalid Config", "testdata/opamp.d/opamp-invalid-remote-config.yaml", true, false, true, "'max_elapsed_time' must be non-negative"},
    }

    for _, tt := range tests {
        t.Run(tt.name, func(t *testing.T) {
            d, err := os.MkdirTemp("", "opamp.d")
            assert.NoError(t, err)
            defer os.RemoveAll(d)

            cfg := createDefaultConfig().(*Config)
            cfg.RemoteConfigurationDirectory = d
            set := extensiontest.NewNopSettings()
            o, err := newOpampAgent(cfg, set.Logger, set.BuildInfo, set.Resource)
            assert.NoError(t, err)

            cfg.AcceptsRemoteConfiguration = tt.acceptsRemote

            path := filepath.Join(tt.file)
            rb, err := os.ReadFile(path)
            assert.NoError(t, err)

            rc := &protobufs.AgentRemoteConfig{
                Config: &protobufs.AgentConfigMap{
                    ConfigMap: map[string]*protobufs.AgentConfigFile{
                        "default": {
                            Body: rb,
                        },
                    },
                },
                ConfigHash: []byte("b2b1e3e7f45d564db1c0b621bbf67008"),
            }

            changed, err := o.applyRemoteConfig(rc)
            if tt.expectError {
                assert.Error(t, err)
                assert.ErrorContains(t, err, tt.errorMessage)
            } else {
                assert.NoError(t, err)
            }
            assert.Equal(t, tt.expectChanged, changed)
        })
    }
}
```
so now the code isn't as repeated :D

pkg/extension/opampextension/opamp_agent_test.go
```diff
@@ -27,567 +27,313 @@ import (
...
+	}
 }
+func TestDefaultOpampAgent(t *testing.T) {
+	testCases := []struct {
+		name     string
+		validate func(*opampAgent, *testing.T)
+	}{
+		{
+			name: "GetAgentCapabilities",
+			validate: func(o *opampAgent, t *testing.T) {
```
chan-tim-sumo, line 155: should do something similar here also, we shouldn't have validate as a input (I see the rest of the changes have this).

harshshahsumo: as discussed offline, we reduced 15-20 lines by using `defaultSetup` for all the testcases in `TestDefaultOpampAgent` hence not touching it

chan-tim-sumo: I mentioned the ```defaultSetup``` was good for refactoring. I feel these tests doesn't need to be inside a table test since most of them (if any) in here doesn't have repeated code. In my opinion, validate shouldn't be an input, but in this specific test case, it isn't WRONG to leave it like this. 

What does the team think about this?

rnishtala-sumo: I tend to agree, validations/assertions are typically not seen as a test input.

harshshahsumo: addressed

pkg/extension/opampextension/opamp_agent_test.go
```diff
@@ -27,40 +27,98 @@ import (
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			d, err := os.MkdirTemp("", "opamp.d")
+			assert.NoError(t, err)
+			defer os.RemoveAll(d)
+
+			cfg := createDefaultConfig().(*Config)
+			cfg.RemoteConfigurationDirectory = d
+			set := extensiontest.NewNopSettings()
+			o, err := newOpampAgent(cfg, set.Logger, set.BuildInfo, set.Resource)
```
chan-tim-sumo, line 88: might be super nit, but 2 small things:

1. this line seems to be repeated in a bunch of tests, should this be refactored as well?
2. i see that this test uses ```createDefaultConfig()```, a bunch of other tests also used that before but is now replaced with ```defaultSetup()```, should this also use that?

harshshahsumo: done!

pkg/extension/opampextension/opamp_agent_test.go
harshshahsumo: addressed

pkg/extension/opampextension/opamp_agent_test.go
harshshahsumo: addressed

pkg/extension/opampextension/opamp_agent_test.go
chan-tim-sumo: ```newOpampAgent``` is repeated in about 5+ tests

pkg/extension/opampextension/opamp_agent_test.go
```diff
@@ -27,40 +27,95 @@ import (
...
 	"go.opentelemetry.io/collector/config/confighttp"
+	"go.opentelemetry.io/collector/extension"
 	"go.opentelemetry.io/collector/extension/extensiontest"
+
 	semconv "go.opentelemetry.io/collector/semconv/v1.18.0"
 )
 
-func TestNewOpampAgent(t *testing.T) {
-	cfg := createDefaultConfig()
+const errMsgRemoteConfigNotAccepted = "OpAMP agent does not accept remote configuration"
```
chan-tim-sumo, line 36: more than 1 const should in a block.

pkg/extension/opampextension/opamp_agent_test.go
```diff
@@ -565,14 +231,13 @@ func TestHackSetEndpoint(t *testing.T) {
 			name:         "prod sumologic url",
 			url:          "https://open-collectors.sumologic.com/api/v1",
 			wantEndpoint: "wss://opamp-collectors.sumologic.com/v1/opamp",
-		},
-		{
+		}, {
```
chan-tim-sumo, line 234: typo?

pkg/extension/opampextension/opamp_agent_test.go
```diff
@@ -27,40 +27,95 @@ import (
...
-	assert.Equal(t, "otelcol-sumo", o.agentType)
-	assert.Equal(t, "sumo.0", o.agentVersion)
-	assert.Equal(t, "7RK6DW2K4V8RCSQBKZ02EJ84FC", o.instanceId.String())
+func setupWithRemoteConfig(t *testing.T, d string) (*Config, extension.Settings) {
+	cfg, set := defaultSetup()
+	cfg.RemoteConfigurationDirectory = d
+	return cfg, set
+}
+
+func setupWithBuildInfo(version, command string) (*Config, extension.Settings) {
```
chan-tim-sumo, line 53: what is the point of this function? can't this line be in ```defaultSetup```?

chan-tim-sumo: this was being used in the test ```setupWithBuildInfo("test version", "otelcoltest")```, shouldn't this be in the setup function?

harshshahsumo: some functions use `	set.BuildInfo = component.BuildInfo{Version: version, Command: command}`
while some do not, so in such places we use either defaultSetup or setupWithBuilderInfo depending upon the testcase

chan-tim-sumo: i agree with that, then you can use a blank identifier to make it cleaner. so that way instead of writing smaller functions, it can all be in one function instead.

pkg/extension/opampextension/opamp_agent_test.go
harshshahsumo: addressed

pkg/extension/opampextension/opamp_agent_test.go
harshshahsumo: addressed

pkg/extension/opampextension/opamp_agent_test.go
harshshahsumo: addressed

chan-tim-sumo (approved):

PR #1632, opened 2024-07-22
fix(opamp): Add the Default provider to the provider list in ConfigProviderSettings

rnishtala-sumo:
Add the Default provider to the provider list in ConfigProviderSettings. When the Scheme for a config URI isn't specified, the DefaultScheme is set to the env, therefore this provider now needs to be included in the provider list.

Related to this upstream breaking change - https://github.com/open-telemetry/opentelemetry-collector/pull/10435

### DIFF
--- a/.changelog/1632.fixed.txt
+++ b/.changelog/1632.fixed.txt
@@ -1,0 +1,1 @@
+fix(opamp): Add the Default(env) provider to the provider list in ConfigProviderSettings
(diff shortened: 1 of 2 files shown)


echlebek (approved):
Looks good. Can you add an explainer in the commit message?

.changelog/1632.fixed.txt
```diff
@@ -0,0 +1 @@
+fix(opamp): Add the Default provider to the provider list in ConfigProviderSettings
```
sumo-drosiek, line 1: It's envprovider

rnishtala-sumo: according to the otel spec the env provider is the default

sumo-drosiek: Hmm, isn't fileprovider default as well?

PR #1639, opened 2024-07-25
Don't use race detector for jobreceiver tests

echlebek:
This commit overrides the race detection in the jobreceiver tests. While this prevents us from catching data races in the jobreceiver in CI, it also keeps CI from failing sporadically on Darwin.

There's something amiss in the jobreceiver tests and it causes CI to fail quite regularly, almost always on Darwin. I suspect there is some concurrency bug in the test harness that is more likely to deadlock when the race detector is enabled. (The race detector slows everything down, and that can make deadlocks more likely to happen by exposing timing bugs).


sumo-drosiek (commented):
Could you add an issue, so we do not loose track on it and eventually investigate in the future?

echlebek:
Good idea. Filed https://github.com/SumoLogic/sumologic-otel-collector/issues/1641

chan-tim-sumo (approved):

PR #1642, opened 2024-07-26
Add smoke tests for otelcol-config

echlebek:
Adds smoke tests demonstrating that main does not crash or exit when compatible combinations of flags are used.

### DIFF
--- a/pkg/tools/otelcol-config/smoke_test.go
+++ b/pkg/tools/otelcol-config/smoke_test.go
@@ -1,0 +1,70 @@
+//go:build !windows
+
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func createSkeleton(dir string) error {
+	if err := os.Mkdir(filepath.Join(dir, ConfDotD), 0700); err != nil {
+		return err
+	}
+	if err := os.Mkdir(filepath.Join(dir, ConfDotDAvailable), 0700); err != nil {
+		return err
+	}
+	if err := os.WriteFile(filepath.Join(dir, ConfDotDAvailable, getHostMetricsFilename()), []byte("hostmetrics"), 0600); err != nil {
+		return err
+	}
+	if err := os.WriteFile(filepath.Join(dir, ConfDotDAvailable, ephemeralYAML), []byte("ephemeral"), 0600); err != nil {
+		return err
+	}
+	return nil
+}
+
+func TestLocallyManagedSmoke(t *testing.T) {
+	// run main against a tempdir with all locally managed flags
+	configDir := t.TempDir()
+	if err := createSkeleton(configDir); err != nil {
+		t.Fatal(err)
+	}
+	flags := []string{
+		"otelcol-config",
+		"--config", configDir,
+		"--add-tag", "foo=bar",
+		"--add-tag", "bar=baz",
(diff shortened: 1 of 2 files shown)


pkg/tools/otelcol-config/smoke_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+func TestRemotelyManagedSmoke(t *testing.T) {
+	// run main against a tempdir with all remotely managed flags
+	configDir := t.TempDir()
+	if err := createSkeleton(configDir); err != nil {
+		t.Fatal(err)
+	}
+	flags := []string{
+		"otelcol-config",
+		"--config", configDir,
```
rnishtala-sumo, line 61: do we need to test any of the tag related flags in a remotely managed collector?

pkg/tools/otelcol-config/smoke_test.go
```diff
@@ -0,0 +1,70 @@
...
+		"--add-tag", "foo=bar",
+		"--add-tag", "bar=baz",
+		"--delete-tag", "bar",
+		"--set-installation-token", "abcdef",
+		"--enable-hostmetrics",
+		"--disable-hostmetrics",
+		"--enable-ephemeral",
+		"--disable-ephemeral",
+		"--set-api-url", "https://example.com",
+		"--write-kv", ".hello.world = \"yes\"",
```
rnishtala-sumo, line 45: what is the `write-kv` flag for?

pkg/tools/otelcol-config/smoke_test.go
```diff
@@ -0,0 +1,70 @@
...
+	if err := os.WriteFile(filepath.Join(dir, ConfDotDAvailable, getHostMetricsFilename()), []byte("hostmetrics"), 0600); err != nil {
+		return err
+	}
+	if err := os.WriteFile(filepath.Join(dir, ConfDotDAvailable, ephemeralYAML), []byte("ephemeral"), 0600); err != nil {
+		return err
+	}
+	return nil
+}
+
+func TestLocallyManagedSmoke(t *testing.T) {
```
rnishtala-sumo, line 27: do we need to add an assertion here on what the expected config file should look like after running all the otel-config actions?

pkg/tools/otelcol-config/smoke_test.go
echlebek: Tags are not locally writable for remotely managed collectors.

pkg/tools/otelcol-config/smoke_test.go
echlebek: write-kv is a developer-oriented flag that modifies `conf.d/00-otel-config-settings.yaml`. It's a generic feature so that we can add support for new configuration in e2e tests without requiring changes to otelcol-config.

pkg/tools/otelcol-config/smoke_test.go
echlebek: No. Data correctness is tested for by the unit tests, and should not be part of the smoke tests, which are only to ensure that the tool runs without crashing. Since correct changes could change the output of the tool, it would create churn to add validation here, and I don't think it's worthwhile.

rnishtala-sumo (approved):

PR #1645, opened 2024-07-30
fix(install): Fix script to always install the latest stable version

rnishtala-sumo:
There's an issue with the script when we hit the github rate limit. When this happens we download an RC if its the most recent release. One could reproduce this by running the old curl command that this PR replaces.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -447,7 +447,9 @@ function get_latest_version() {
 
     # get latest version directly from website if there is no versions from api
     if [[ -z "${versions}" ]]; then
-        curl --retry 5 --connect-timeout 5 --max-time 30 --retry-delay 0 --retry-max-time 150 -s https://github.com/SumoLogic/sumologic-otel-collector/releases | grep -oE '/SumoLogic/sumologic-otel-collector/releases/tag/(.*)"' | head -n 1 | sed 's%/SumoLogic/sumologic-otel-collector/releases/tag/v\([^"]*\)".*%\1%g'
+        curl --retry 5 --connect-timeout 5 --max-time 30 --retry-delay 5 --retry-max-time 150 -s https://github.com/SumoLogic/sumologic-otel-collector/releases \
+        | grep -Eo '/SumoLogic/sumologic-otel-collector/releases/tag/v[0-9]+\.[0-9]+\.[0-9]+-sumo-[0-9]+[^-]' \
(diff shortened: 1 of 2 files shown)


scripts/install.sh
```diff
@@ -447,7 +447,20 @@ function get_latest_version() {
 
     # get latest version directly from website if there is no versions from api
     if [[ -z "${versions}" ]]; then
-        curl --retry 5 --connect-timeout 5 --max-time 30 --retry-delay 0 --retry-max-time 150 -s https://github.com/SumoLogic/sumologic-otel-collector/releases | grep -oE '/SumoLogic/sumologic-otel-collector/releases/tag/(.*)"' | head -n 1 | sed 's%/SumoLogic/sumologic-otel-collector/releases/tag/v\([^"]*\)".*%\1%g'
+        curl \
+        --connect-timeout 5 \
+        --max-time 30 \
+        --retry 5 \
+        --retry-delay 0 \
```
echlebek, line 454: Seems like we should either not specify retry-delay (in which case it uses a default) or it should be set to greater than zero?

sumo-drosiek (changes requested):
Please merge https://github.com/SumoLogic/sumologic-otel-collector-packaging/pull/102/files first

scripts/install.sh
rnishtala-sumo: yeah, a 5 second delay sounds reasonable. The default is exponential backoff we probably don't need that.

scripts/install.sh
```diff
@@ -447,7 +447,9 @@ function get_latest_version() {
 
     # get latest version directly from website if there is no versions from api
     if [[ -z "${versions}" ]]; then
-        curl --retry 5 --connect-timeout 5 --max-time 30 --retry-delay 0 --retry-max-time 150 -s https://github.com/SumoLogic/sumologic-otel-collector/releases | grep -oE '/SumoLogic/sumologic-otel-collector/releases/tag/(.*)"' | head -n 1 | sed 's%/SumoLogic/sumologic-otel-collector/releases/tag/v\([^"]*\)".*%\1%g'
+        curl --retry 5 --connect-timeout 5 --max-time 30 --retry-delay 5 --retry-max-time 150 -s https://github.com/SumoLogic/sumologic-otel-collector/releases \
```
rnishtala-sumo, line 450: made changes to the curl after review in the other PR - https://github.com/SumoLogic/sumologic-otel-collector-packaging/pull/102/files#r1698052229

sumo-drosiek (approved):
Great 🚀

PR #1656, opened 2024-08-15
ci: build & test otelcol-config

amdprophet:
Adds build & test CI jobs for the new otelcol-config command. Refactors the toolchain code used for building FIPS images to allow the Docker container to build other binaries.


.github/workflows/workflow-build-otelcol-config.yml
```diff
@@ -0,0 +1,166 @@
...
+        run: |
+          go tool nm ${{ steps.set-binary-name.outputs.binary_name }} | \
+          grep "_Cfunc__goboringcrypto_"
+
+      # TODO: find a way to test if the binary runs
+      # - name: Test binary
+      #   if: inputs.arch_os == env.ARCH_OS
+      #   working-directory: ./pkg/tools/otelcol-config
+      #   run: |
+      #     ./${{ steps.set-binary-name.outputs.binary_name }} -t foo
```
echlebek, line 144: There's already a smoke test in otelcol-config itself, so this is probably not needed

.github/workflows/workflow-test-otelcol-config.yml
```diff
@@ -0,0 +1,77 @@
...
+        description: Save the module and build caches.
+        default: false
+        type: boolean
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  GO_VERSION: "1.21.4"
```
echlebek, line 31: This should be upgraded to at least 1.21.11

.github/workflows/workflow-build-otelcol-config.yml
amdprophet: This was copied from the otelcol-sumo build job. It's meant to ensure that the binary runs after running make. We ran into an issue where the binaries were actually text files in the past.

It may be better to check the file type for every platform instead. We'd need to check for the correct file type for the platform though (e.g. `Mach-O 64-bit executable arm64` versus `ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, Go BuildID=vataBbMZrfQgeqsJPb3y/CT0CNbBkuU-ZpeiBCFvZ/rX1IRAQtPBnQNJWqqDt9/SENzp0KECm910ny3Zm29, stripped`)

.github/workflows/workflow-test-otelcol-config.yml
amdprophet: Done!

amdprophet:
@echlebek ready for review!

.github/workflows/workflow-build-otelcol-config.yml
echlebek: That seems kind of nitty to me. How did it regress to a text file?

echlebek (approved):
Looking forward to packaging, inshallah :pray:

.github/workflows/workflow-build-otelcol-config.yml
amdprophet: I'm not positive but I believe it was a problem with binaries moving between jobs through the CI cache system which isn't happening the same way here. We can leave this out.

PR #1659, opened 2024-08-19
Fix a developer error in the jobreceiver tests

echlebek:
Calling t.FailNow or functions that call t.FailNow indirectly from goroutines outside the function under test is not allowed. In this instance, it seems it was causing the test to deadlock.

### DIFF
--- a/pkg/receiver/jobreceiver/command/command_test.go
+++ b/pkg/receiver/jobreceiver/command/command_test.go
@@ -22,19 +22,14 @@ func TestExecute(t *testing.T) {
 	t.Log("Running basic test")
 	// Basic command execution
 	t.Run("basic", func(t *testing.T) {
-		t.Log("debug 1")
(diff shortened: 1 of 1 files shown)


pkg/receiver/jobreceiver/command/command_test.go
```diff
@@ -110,21 +107,32 @@ func withTestHelper(t *testing.T, r ExecutionRequest) ExecutionRequest {
...
+	out := make(chan executionOutput, 1)
 	stdout, err := i.Stdout()
 	require.NoError(t, err)
 	stderr, err := i.Stderr()
 	require.NoError(t, err)
 	go func() {
+		defer close(out)
 		var buf bytes.Buffer
 		_, err := io.Copy(&buf, stdout)
-		require.NoError(t, err)
```
rnishtala-sumo, line 123: looks like calling `NoError` in the go routine was the issue...

rnishtala-sumo (approved):

PR #1661, opened 2024-08-22
bugfix(otelcol-config): quote tag names

echlebek:
In `otelcol-config`, `--add-tag` is creating hierarchical tag structures when a `.` exists in the tag name. However, the backend only accepts string values. This change makes sure that tags have a flat structure.


echlebek:
Changelog has been skipped as otelcol-config has not yet been shipped for the first time.

ccressent (approved):

PR #1687, opened 2024-10-16
Add find-build-num tool

echlebek:
The find_build_num tool is useful for finding which package build corresponds to a particular sumologic-otel-collector SHA or tag.

This code is not part of the Sumo OTC distribution - it is glue code for relating jobs between the sumologic-otel-collector repository and the sumologic-otel-collector-packaging repository.


echlebek:
Looks like golangci-lint needs to be updated

PR #1688, opened 2024-10-18
Use staticcheck instead of golangci-lint

echlebek:
This commit changes the definition of the "golint" make target. It now uses staticcheck instead of golangci-lint.

Most of the value in golangci-lint is derived from staticcheck, and staticcheck on its own runs in about 25% of the time as golangci-lint. So, there is little reason to use golangci-lint.

Additionally, golangci-lint configures staticcheck in a non-standard way. As such, some lint errors were masked. They have been fixed in this commit as well.


echlebek:
Reviewers can see the lint errors for themselves by running this on the main branch:
```
find . -type f -name 'go.mod' -exec dirname {} \; | xargs -L 1 bash -c 'cd "${0}" && staticcheck && cd'
```

pkg/receiver/rawk8seventsreceiver/receiver.go
```diff
@@ -103,22 +103,27 @@ func newRawK8sEventsReceiver(
...
 	for _, namespace := range namespaces {
 		namespaceListWatch := listerWatcherFactory(restClient, "events", namespace, fields.Everything())
-		_, namespaceController = cache.NewInformer(namespaceListWatch, &corev1.Event{}, 0, cache.ResourceEventHandlerFuncs{ // nolint:staticcheck
-			AddFunc: func(obj interface{}) {
-				event := obj.(*corev1.Event)
-				eventCh <- &eventChange{
-					changeType: eventChangeTypeAdded,
-					event:      event,
-				}
+		opts := cache.InformerOptions{
```
rnishtala-sumo, line 106: Should we raise a different PR for this change? I think we'll need a changelog entry for this.

echlebek: In my opinion this is a simpler change than it may seem outwardly. If you have a look at the definition of cache.NewInformer, it just calls cache.NewInformerWithOptions in the same way it's called here. There would be no change to communicate in the changelog.

echlebek: (cache.NewInformer is deprecated, and a notice has been added to switch to cache.NewInformerWithOptions)

rnishtala-sumo (approved):

PR #1689, opened 2024-10-21
feat(k8s_tagger): Added pagination when fetching pods from the k8s List API

rnishtala-sumo:
Added pagination when fetching pods from the k8s List API.

Unset the ResourceVersion to get the MostRecent change for an object
https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions

### DIFF
--- a/pkg/processor/k8sprocessor/kube/informer.go
+++ b/pkg/processor/k8sprocessor/kube/informer.go
@@ -16,6 +16,7 @@ package kube
 
 import (
 	"context"
+	"time"
 
 	"go.uber.org/zap"
 	api_v1 "k8s.io/api/core/v1"
@@ -36,6 +37,7 @@ type InformerProvider func(
 	namespace string,
 	labelSelector labels.Selector,
 	fieldSelector fields.Selector,
+	limit int,
 ) cache.SharedInformer
 
 func newSharedInformer(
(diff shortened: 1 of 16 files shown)


pkg/processor/k8sprocessor/kube/informer.go
```diff
@@ -53,11 +56,27 @@ func newSharedInformer(
...
 }
 
-func informerListFuncWithSelectors(client kubernetes.Interface, namespace string, ls labels.Selector, fs fields.Selector) cache.ListFunc {
+func informerListFuncWithSelectors(logger *zap.Logger, client kubernetes.Interface, namespace string, ls labels.Selector, fs fields.Selector) cache.ListFunc {
 	return func(opts metav1.ListOptions) (runtime.Object, error) {
-		opts.LabelSelector = ls.String()
-		opts.FieldSelector = fs.String()
-		return client.CoreV1().Pods(namespace).List(context.Background(), opts)
+		podList := &api_v1.PodList{}
+		for {
```
amdprophet, line 62: Should this be bounded to a timeout to prevent this from potentially running forever? I think you can create a new context, call [WithDeadline](https://pkg.go.dev/context#WithDeadline), and pass the context to the call to `List()` on line 68. This raises other questions though:
* How long should the timeout be?
* Should the timeout be configurable?

rnishtala-sumo: yes, made the change to use `WithDeadline` thanks! I let the timeout be 60 seconds at the moment. I made the page size (limit) in the API call configurable.

amdprophet (approved):
LGTM 👍

chan-tim-sumo (approved):

pkg/processor/k8sprocessor/kube/informer.go
echlebek: @rnishtala-sumo FYI you can use `context.WithTimeout(duration)` instead of `context.WithDeadline` which saves you computing the `time.Time`

rnishtala-sumo: thanks!, made the change

PR #1691, opened 2024-10-22
otelcol-config: add launchd support

amdprophet:
Adds support for launchd on macOS as it is used to manage the installation token.


pkg/tools/otelcol-config/installation_token_test.go
```diff
@@ -72,6 +74,44 @@ func TestSetInstallationTokenAction(t *testing.T) {
...
+		{
+			Name:  "launchd",
+			Flags: []string{"--set-installation-token", "abcdef"},
+			Launchd: fstest.MapFS{
+				launchdConfigPlist: &fstest.MapFile{
+					Data: []byte(defaultLaunchdConfigXML),
+				},
+			},
+			ExpWrite: []byte(xmlPreamble +
+				`<plist version="1.0">` +
```
chan-tim-sumo, line 86: just curious, how was this created? 🤔

echlebek: FYI, when using backticks, there is no need to concatenate multiple lines. Newlines are included in backtick strings.

amdprophet: I was trying to avoid the line wrapping that happens when using backticks with multiple lines. Backticks also prevent having to escape double-quotes.

amdprophet: > just curious, was this auto generated? 🤔
> 
> ```
> <plist version="1.0">` +
> `<dict>` +
> `<key>EnvironmentVariables</key>` +
> `<dict>` +
> `<key>SUMOLOGIC_INSTALLATION_TOKEN</key>` +
> ```

Yes, this comes from the packaging repo and is installed by the packages.

chan-tim-sumo (approved):

PR #1693, opened 2024-10-31
chore(ci): pkg trigger and release workflow updates

amdprophet:
Packaging is now only triggered from Dev & PR builds. The only input sent when triggering packaging is now the Workflow Run ID of the workflow run that packaging should pull its artifacts from.

The version of otelcol-sumo & otelcol-config is now determined using a combination of the otelcol_version in
`otelcolbuilder/.otelcol-builder.yaml` and git tags.

Releases are no longer triggered by the creation of a tag. A release can now be triggered via the new 'Publish release' workflow. The new release workflow requires the ID of a workflow run that contains the artifacts to be released.

The Git SHA used to build the otelcol-sumo binary is now included in the version of the `--version` flag.


.github/workflows/dev_builds.yml
```diff
@@ -451,57 +451,6 @@ jobs:
             REPO_URL=sumologic/sumologic-otel-collector-dev \
             BUILD_TYPE_SUFFIX="-ubi"
 
-  package-msi:
```
chan-tim-sumo, line 454: what does removing MSI and wixext do? just curious

echlebek (changes requested):
Please update the release instructions: https://github.com/SumoLogic/sumologic-otel-collector/blob/main/docs/release.md

.github/workflows/dev_builds.yml
amdprophet: This is done in the packaging repository. We'll no longer be building or testing MSI or MSI code in this repository.

echlebek (approved):

PR #1705, opened 2024-11-15
fix(opamp): multiple fixes for opamp

amdprophet:
* Uses new sumologicextension from our fork of opentelemetry-collector-contrib. This is necessary for now as the changes are not available on upstream and upstream is using a newer version of the collector.

* Updated the OpAmp agent to output the filename (includes the source template ID) when processing source templates.

* Added the `file_storage` and `health_check` extensions to the generated `sumologic-remote.yaml` config when using otelcol-config.

### DIFF
--- a/otelcolbuilder/.otelcol-builder.yaml
+++ b/otelcolbuilder/.otelcol-builder.yaml
@@ -227,7 +227,6 @@ providers:
   - gomod: go.opentelemetry.io/collector/confmap/provider/httpsprovider v0.108.0
   - gomod: go.opentelemetry.io/collector/confmap/provider/yamlprovider v0.108.0
 
-
 excludes:
   - github.com/knadh/koanf v1.5.0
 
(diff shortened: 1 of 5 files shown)


otelcolbuilder/.otelcol-builder.yaml
```diff
@@ -243,5 +242,6 @@ replaces:
   # version which gets then translated into a replace in go.mod file.
   # This does not replace the version that sumologicexporter depends on.
   - github.com/SumoLogic/sumologic-otel-collector/pkg/extension/opampextension => ../../pkg/extension/opampextension
+  - github.com/open-telemetry/opentelemetry-collector-contrib/extension/sumologicextension v0.108.0 => github.com/SumoLogic/opentelemetry-collector-contrib/extension/sumologicextension bf012a09b9651cddda4a907ad4ff17834a0b81ef
```
chan-tim-sumo, line 245: does this mean the hash would need to be changed for every new version release? or is it here just for this PR 🤔

amdprophet: We're using a fork of upstream at v0.108.0. Upstream is already on 0.113.0. We'll need to get the changes on our fork into upstream. We can stop using our fork once we update to the next version they release with our changes.

fguimond (approved):

PR #1709, opened 2024-11-22
fix(opamp): update default opamp endpoint

fguimond:
The OpAmp endpoint with redirection has been moved to a different URL. Updating it.

### DIFF
--- a/pkg/tools/otelcol-config/remote_control.go
+++ b/pkg/tools/otelcol-config/remote_control.go
@@ -8,7 +8,7 @@ import (
 )
 
 const (
-	DefaultSumoLogicOpampEndpoint       = "wss://opamp-events.sumologic.com/v1/opamp"
+	DefaultSumoLogicOpampEndpoint       = "wss://opamp-collectors.sumologic.com/v1/opamp"
 	DefaultConfigurationDirectory       = "/etc/otelcol-sumo"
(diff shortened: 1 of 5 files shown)


chan-tim-sumo:
should it be changed in the opamp_agent.go file? found the endpoint in that file also. 

https://github.com/SumoLogic/sumologic-otel-collector/blob/74a9060c41224a13f8018f8ab511a167d8589bdf/pkg/extension/opampextension/opamp_agent.go#L47

pkg/tools/otelcol-config/remote_control.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 const (
-	DefaultSumoLogicOpampEndpoint       = "wss://opamp-events.sumologic.com/v1/opamp"
```
chan-tim-sumo, line 11: should it be changed in the opamp_agent.go file? found the endpoint in that file also. 

https://github.com/SumoLogic/sumologic-otel-collector/blob/74a9060c41224a13f8018f8ab511a167d8589bdf/pkg/extension/opampextension/opamp_agent.go#L47

fguimond: Good catch, updated.

chan-tim-sumo (approved):

PR #1728, opened 2025-01-27
Fix env variable parsing

gmrukwa:
The installation token wouldn't be picked up from env variable according to OpenTelemetry Collector config parser.

### DIFF
--- a/examples/sumologic.yaml
+++ b/examples/sumologic.yaml
@@ -5,7 +5,7 @@ extensions:
   ## Manages registration, heartbeats and authentication to Sumo Logic
   ## ref: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/sumologicextension
   sumologic:
-    installation_token: ${SUMOLOGIC_INSTALLATION_TOKEN}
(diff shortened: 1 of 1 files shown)


amdprophet:
@gmrukwa Both are supported (See: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/env-vars.md). Are you running into an issue?

gmrukwa:
> @gmrukwa Both are supported (See: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/env-vars.md). Are you running into an issue?

Yes, the _braces syntax_ didn't work for me. The value of the env variable wasn't picked up by the collector. I needed to switch to the _env provider syntax_. And it wasn't an issue with env variable name, as the name was fairly simple: `SUMOLOGIC_INSTALLATION_TOKEN`.

echlebek (approved):
The syntax for environment variables has changed, and the doc that @amdprophet linked is now out of date. This change make sense to me.

Astitva-sumo:
Closing stale PR

PR #1746, opened 2025-03-20
fix(ci): store config management tarballs as dev build artifacts

amdprophet:
Configuration Management tarballs are uploaded as a GitHub Action artifact in the release workflow, but are no longer being stored as artifacts in our GitHub Releases. This was overlooked after we switched to our new release workflow.

To resolve this I've added the code to build & upload the CM tarballs as GitHub Action artifacts to the dev builds workflow. All dev build artifacts are automatically included in GitHub Releases.

I've also removed the install-script and config-management-assets jobs from the release workflow as they no longer serve a purpose.


.github/workflows/releases.yml
```diff
@@ -282,78 +282,6 @@ jobs:
             REPO_URL=sumologic/sumologic-otel-collector \
             BUILD_TYPE_SUFFIX="-ubi"
 
-  # Store the install script from the packaging repository as a release artifact.
```
rnishtala-sumo, line 285: So to confirm, all the install scripts only live in the packaging repository as of the latest release workflow?

.github/workflows/dev_builds.yml
```diff
@@ -460,6 +473,42 @@ jobs:
           path: ./install-script/install.ps1
           if-no-files-found: error
 
+  config-management-assets:
```
rnishtala-sumo, line 476: To confirm, with this change these assets will now be under - [https://github.com/SumoLogic/sumologic-otel-collector/releases/latest/download/ansible-playbook](https://github.com/SumoLogic/sumologic-otel-collector/releases/latest/download/ansible-playbook.tar.gz)

.github/workflows/releases.yml
amdprophet: They still exist within this repository but they do not get included with the release. The Sumo website is also using the install scripts in the packaging repo since the release workflow changes went into effect.

.github/workflows/dev_builds.yml
amdprophet: Correct.

rnishtala-sumo (approved):

PR #1748, opened 2025-03-20
chore(ci): create pkg tags after release publish

amdprophet:
As a part of the release process, we used to create tags for each of the exportable packages under the `pkgs` directory. This was lost when we switched to the new release workflow. This PR adds a new `post-release` workflow to reimplement the creation of these tags. It will create and push tags automatically when a release is published. (e.g. draft -> published)

**NOTE** I've done some rudimentary testing but I don't think we can know if this will work until the next release.

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -1,0 +1,33 @@
+name: "Post release tasks"
+
+run-name: >
+  ${{ format('Post Release Tasks for Release: {0}', github.ref_name) }}
+
+on:
+  release:
+    types: [released]
+
+permissions:
+  contents: write
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  create-pkg-tags:
+    name: Create package tags
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: "${{github.ref_name}}"
+
+      - name: Create & push package tags
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/post-release.yml
```diff
@@ -0,0 +1,33 @@
...
+  contents: write
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  create-pkg-tags:
+    name: Create package tags
+    runs-on: ubuntu-latest
```
rnishtala-sumo, line 20: should this be ubuntu-24.04?

amdprophet: It should -- I was going to fix it after the other PR is merged and I can rebase this branch.

amdprophet: Done!

rnishtala-sumo (approved):

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: actions/checkout@v4
+        with:
+          ref: "${{github.ref_name}}"
+
+      - name: Create & push package tags
+        run: |
+          for dir in $(find ./pkg -type f -name "go.mod" ! -path "*pkg/test/*" -exec dirname {} \; | sort); do
+            tag_name="${dir:2}/${{github.ref_name}}";
+            echo "Creating tag for: ${tag_name}";
+            tag -a "${tag_name}" -s -m "${tag_name}" "${{github.ref_name}}";
```
echlebek, line 31: Is this supposed to say "git tag"?

PR #1770, opened 2025-06-04
otelcol-config: add timezone support

Gourav2906:
Add support to timezone so that we can use it in add collector workflow

### DIFF
--- a/pkg/tools/otelcol-config/timezone.go
+++ b/pkg/tools/otelcol-config/timezone.go
@@ -48,17 +48,17 @@ func SetTimezoneAction(ctx *actionContext) error {
 			// shouldn't ever happen
 			panic(err)
 		}
-		doc = buf.Bytes()
+		config = buf.Bytes()
 	} else {
 		expression := fmt.Sprintf(".extensions.sumologic.time_zone = %q", ctx.Flags.SetTimezone)
-		result, err := eval.EvaluateAll(expression, string(doc), encoder, decoder)
+		result, err := eval.EvaluateAll(expression, string(config), encoder, decoder)
 		if err != nil {
 			return fmt.Errorf("couldn't write timezone: %s", err)
 		}
-		doc = []byte(result)
+		config = []byte(result)
 	}
 
-	_, err = writer(doc)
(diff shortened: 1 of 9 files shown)


pkg/tools/otelcol-config/timezone.go
```diff
@@ -0,0 +1,67 @@
...
+	default:
+		writer = ctx.WriteConfD
+		config = conf.ConfD[ConfDSettings]
+	}
+
+	encoder := yqlib.YamlFormat.EncoderFactory()
+	decoder := yqlib.YamlFormat.DecoderFactory()
+	eval := yqlib.NewStringEvaluator()
+
+	if len(config) == 0 {
```
rnishtala-sumo, line 35: It would be nice to add a test to cover the expected behavior when this config file isn't present, example - https://github.com/SumoLogic/sumologic-otel-collector/blob/c96bd9fc8eaf3a215d3f2cf50d5d05ee88c36a42/pkg/tools/otelcol-config/remote_control_test.go#L75

This would hopefully test the default timezone scenario.

Also, show we include the timezone in the initial config here? - https://github.com/SumoLogic/sumologic-otel-collector/blob/c96bd9fc8eaf3a215d3f2cf50d5d05ee88c36a42/pkg/tools/otelcol-config/remote_control.go#L69

cc: @echlebek

Gourav2906: it is okay not to include default timezone since API also have default "UTC"

Gourav2906: Added more test

rnishtala-sumo: IMO it makes sense for the collector to have a default timezone written to its config when none is set and this should be documented in the readme. The sumo extension doesn't seem to set a default timezone either.

Gourav2906: Added default also

echlebek (approved):
Approved once tests pass

Gourav2906:
All tests are passing , so merging this

PR #1780, opened 2025-06-23
chore: removing sumologicextension

raunakmandaokar:
https://sumologic.atlassian.net/browse/OSC-1040
removing the sumologic extension. Already using from upstream.


echlebek (changes requested):
This needs to be mentioned in the changelog, it is a breaking change for any users that may be importing the module.

echlebek (approved):

PR #1793, opened 2025-07-16
bumping upstream version to 0.130.0 

pankaj101A:
1. Upgrade upstream dependency to version v0.130.0.
2. Auth field is now of type configoptional.Optional[configauth.Config] (was previously configauth.Config).
3. Updated opamp_agent and related test code to reflect this change in Auth field usage.


echlebek:
No longer necessary as we have done this in another branch.

PR #1794, opened 2025-07-16
chore(deps): remove SumoLogic/telegraf dep

echlebek:
Removes the dependency on github.com/SumoLogic/telegraf, which we forked several years ago.


echlebek:
Rebased on v0.130.0

PR #1796, opened 2025-07-21
chore: adding first batch of exporters to opamp extension

Astitva-sumo:
Adding another batch of components to opamp extension. This is being done to enable these components to be added in remotely managed collectors.


.changelog/1796.added.txt
```diff
@@ -0,0 +1 @@
+chore: adding first batch of exporters to opamp extension
```
pankaj101A, line 1: we should list all exporters added

Gourav2906: +1

Gourav2906 (approved):

pkg/extension/opampextension/testdata/opamp.d/opamp-debug-exporter-config.yaml
```diff
@@ -0,0 +1,18 @@
+receivers:
+  otlp:
+    protocols:
+      grpc:
+      http:
+exporters:
+  debug:
```
chan-tim-sumo, line 7: isn't debug already built in? or is this a different debug exporter

chan-tim-sumo (approved):

PR #1809, opened 2025-08-01
Move Dockerfiles, Refactor CI & Makefiles

amdprophet:
The Dockerfiles have been moved to https://github.com/sumologic/sumologic-otel-collector-containers to reduce the size of our GitHub Actions caches & to separate concerns. Similar to how our packaging workflow works, the `build-and-push.yml` workflow in the containers repository are triggered after the binaries in the `dev_builds.yml` workflow in this repository have successfully been built.

I've done some work to try and improve CI performance & code readability & reusability. There's still a lot that can be done but we're limited by the 10 GB size limit for GitHub Actions cache. We can potentially get around this by using Azure Blob Storage or using third-party runners.

I've also refactored our Makefiles to be more readable and fixed various issues I ran into.


amdprophet:
The failing test is unrelated to these changes. It seems that a SourceForge URL is returning a 403 when markdownlint checks the URL for validity.

.github/workflows/workflow-trigger-remote-workflow.yml
```diff
@@ -0,0 +1,69 @@
+#################################################################################
```
rnishtala-sumo, line 1: Is this just renaming the file?

.github/workflows/close-jira-task.yaml
```diff
@@ -12,10 +12,14 @@ jobs:
...
         run: |-
+          issue="${{ github.event.issue.number }}"
+          query="summary~\"sumologic-otel-collector%20and%20%23${issue}\""
+          url="${JIRA_BASE_URL}/rest/api/latest/search?jql=${query}"
+
           JIRA_ID=$(curl \
             --http1.1 \
             -u "${JIRA_USER_EMAIL}:${JIRA_API_TOKEN}" \
-            ${JIRA_BASE_URL}'/rest/api/latest/search?jql=summary~"sumologic-otel-collector%20and%20%23${{ github.event.issue.number }}"' \
+            "${url}" \
```
rnishtala-sumo, line 22: When do we close the Jira issue?

.github/actions/determine-binary-name/action.yml
```diff
@@ -0,0 +1,56 @@
+name: Determine the name of a binary
```
rnishtala-sumo, line 1: If I understand we've pulled out some of the steps from `workflow-build` to be their own actions?

.github/actions/determine-binary-name/action.yml
```diff
@@ -0,0 +1,56 @@
...
+    required: true
+
+outputs:
+  name:
+    value: ${{ steps.binary.outputs.name }}
+  name-with-platform:
+    value: ${{ steps.binary.outputs.name-with-platform }}
+
+runs:
+  using: "composite"
```
rnishtala-sumo, line 28: Composites do seem useful for better readability :+1:

rnishtala-sumo (commented):
@amdprophet  The changes make sense. I wonder if we could split the workflow build steps into actions (using composites) incrementally, just to reduce the number of changes in this PR.

rnishtala-sumo (commented):
@amdprophet  The changes make sense. I wonder if we could split the workflow build steps into actions (using composites) incrementally, just to reduce the number of changes in this PR.

.github/workflows/close-jira-task.yaml
amdprophet: I'm not sure about this; this change was just to reduce the line to <= 80 characters for linting.

.github/actions/determine-binary-name/action.yml
amdprophet: Correct as these steps are used in multiple places.

rnishtala-sumo (approved):

PR #1822, opened 2025-09-02
SUMO-269287 Add support for remote management in ansible

Gourav2906:
Add support for remote management in ansible


rahul-babarwal (commented):
Are there any ITs which can be included to test this out

Gourav2906:
No , we do not support ITs for chef / ansible / puppet, QE will work on this frame work for this in Q4

PR #1847, opened 2025-09-17
chore: adding next set of components to opampextension

Astitva-sumo:


Astitva-sumo:
> Looks good. Do you know if upstream is addressing the podman CVE?

Not sure. I checked the open issues, but could find any mention regarding the CVE. Interestingly, although snyk is marking this as critical, but CVE that its referencing is actually empty. https://www.cve.org/CVERecord?id=CVE-2025-47913

PR #1848, opened 2025-09-18
Puppet remote support for windows

himanshu130700:
Screenshot 2025-09-25 at 12 45 29 PM
Screenshot 2025-09-25 at 12 44 28 PM remote support for window


examples/puppet/modules/install_otel_collector/manifests/init.pp
```diff
@@ -30,6 +30,7 @@
   Optional[String] $version = undef,
   String $src_config_path = 'puppet:///modules/install_otel_collector/conf.d',
   Optional[String] $opamp_api_url = undef,
+  Boolean $remotely_managed = true,
```
Gourav2906, line 33: default should be false

PR #1849, opened 2025-09-18
chore:skip signgin mac/windows binaries for dependabot

pankaj101A:
Dependabot PRs trigger our GitHub Actions workflows, including macOS code-signing. This step relies on secrets (APPLE_DEVELOPER_CERTIFICATE_P12_BASE64, APPLE_DEVELOPER_CERTIFICATE_PASSWORD). [link](https://docs.github.com/en/code-security/dependabot/troubleshooting-dependabot/troubleshooting-dependabot-on-github-actions#troubleshooting-failures-when-dependabot-triggers-existing-workflows)

Problem
Secrets are not available to workflows triggered by Dependabot PRs (GitHub security restriction). As a result, the step apple-actions/import-codesign-certs@v5 fails with:

Error: At least one of p12-filepath or p12-file-base64 must be provided
Solution
We’ll skip code-signing when the PR author is dependabot and only run signing on trusted branches (main/release). Example fix:

- uses: apple-actions/import-codesign-certs@v5
  if: runner.os == 'macOS' && inputs.signing-enabled && github.actor != 'dependabot'
  with:
    p12-file-base64: ${{ secrets.APPLE_DEVELOPER_CERTIFICATE_P12_BASE64 }}
    p12-password: ${{ secrets.APPLE_DEVELOPER_CERTIFICATE_PASSWORD }}
This keeps Dependabot PRs green while maintaining secure signing for our release builds.

same issue goes for windows failed step

Dependabot updgrades only github action , go modules, docker version so it should not cause any problem


amdprophet:
I'd like to see the workflow approval process implemented instead of adding conditional logic: https://docs.github.com/en/actions/how-tos/manage-workflow-runs/approve-runs-from-forks. This will allow us to test binaries when necessary.

pankaj101A:
> Dependabot could open PRs to bump the version of the signing actions but they wouldn't be tested in this case. Manually approving CI runs or giving dependabot access to the secrets will ensure that all version bumps are properly tested.

In current PRs check all checks are ran and tested the binary, only signing part is left which is will tested when dev approve PR and gets merged. 

its not recommended to give secrets to dependabot.

tried: https://docs.github.com/en/actions/how-tos/manage-workflow-runs/approve-runs-from-forks but its only applicable for forked prs, dependabot prs are not forked

PR #1865, opened 2025-10-07
chore:set upstream to 0.137.0

pankaj101A:
set upstream to 0.137.0

### DIFF
--- a/.changelog/1845.changed.1.txt
+++ b/.changelog/1845.changed.1.txt
@@ -1,0 +1,1 @@
+set upstream to 0.137.0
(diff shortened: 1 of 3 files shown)


Gourav2906 (approved):

rnishtala-sumo (changes requested):
Recommend we run upgrade tests in the E2E before making a release with [this change](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/42511). The collectors shouldn't re-register on upgrade.

Unless I'm missing something, I don't currently see E2E tests for upgrade.

PR #2022, opened 2026-04-07
feat(OSC-1274): Remove Batch Processor references

ravishankar-sumo:


otelcolbuilder/cmd/testdata/filelog_reallife_complicated_sumologicexporter.yaml
```diff
@@ -628,7 +628,7 @@ exporters:
       enabled: false
     sending_queue:
       enabled: true
-      queue_size: 50000
```
jagan2221, line 631: In this config, removed batch processor has custom settings like timeout 5s and send_batch_size=512 , our default settings has 1s timeout etc. so we should override here right?

Also why is queue_size removed?

otelcolbuilder/cmd/testdata/filelog_reallife_complicated_sumologicexporter.yaml
```diff
@@ -628,7 +628,11 @@ exporters:
...
+      queue_size: 2560000
```
jagan2221: 512x50000 = 25600000 
I think 1 trailing 0 is missing, pls ensure.

jagan2221 (approved):

PR #2053, opened 2026-04-16
Add memorylimiter extension support

poblahblahblah:
Fixes #2051

PR #2060, opened 2026-04-22
add signaltometricsconnector to sumologic-otel-collector

poblahblahblah:
This adds the `signaltometricsconnector` connector to the sumologic-otel-collector. We are wanting to get some rough counts of logs based on a few attributes into our metrics system. Through experimentation outside of the sumologic-otel-collector this was determined to get us what we needed.


jagan2221:
https://github.com/SumoLogic/sumologic-otel-collector/pull/2053/changes#diff-033152de82721aa392c05b9098d2f4ecf72fede80d86e0e3589c39d25f33de21
Pls add the component in opampextension too as done above.

poblahblahblah:
I should have asked, but it doesn't appear any connectors are configured that way!

I based this change off of the connectors added in https://github.com/SumoLogic/sumologic-otel-collector/pull/1102 and https://github.com/SumoLogic/sumologic-otel-collector/pull/1891

The failures of the MR checks look to be transient?

jagan2221:
We should add connectors too. We can skip for this PR, we'll check and add missing connectors here.

jagan2221 (approved):

PR #2076, opened 2026-05-05
fix(opamp): remove expandconverter from OpAMP config provider to fix $1 pattern crash

raunakmandaokar:
### Problem
When using a remote Source Template (ST) with a hashing regex processing rule, the agent crashes during startup with: failed to setup configuration components: environment variable "1" has invalid name: must match regex ...
This only happens with remotely managed configs (OpAMP). Locally managed configs work fine.

### Root Cause
The processing pipeline for config values is:

1. confmap's escapeDollarSigns — leaves $1 untouched (digits are not valid env var name starts)
2. expandconverter — runs os.Expand("$1", ...) which treats $1 as a shell variable, calls the mapping function with "1", which fails env var name validation → crash

`NewOpAmpConfigProviderSettings` included `expandconverter.NewFactory()` in its `ConverterFactories`, but the local config path did not. This discrepancy was introduced when the local path was migrated to the otelcol builder (which removed `expandconverter`) but `NewOpAmpConfigProviderSettings `was missed.

Note: `expandconverter` has been officially deprecated since [otelcol v0.107.0](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.107.0) precisely because it causes double-escaping issues when used alongside confmap's resolver.

###Fix
Removed `expandconverter.NewFactory()` from `NewOpAmpConfigProviderSettings` in `otelcolbuilder/cmd/configprovider.go`, making the OpAMP config provider consistent with the local config path.

###Testing
Added regression tests in `otelcolbuilder/cmd/configprovider_test.go`:

- TestNewOpAmpConfigProviderSettingsHasNoExpandConverter — asserts ConverterFactories is empty
- TestOpAmpConfigWithDollarPatternsResolvesWithoutError — end-to-end resolution of a config with $1 capture group patterns using OpAMP provider settings

Also, tested end 2 end and verfied the hash is working fine.

### DIFF
--- a/.changelog/2076.fixed.txt
+++ b/.changelog/2076.fixed.txt
@@ -1,0 +1,1 @@
+fix(opamp): remove expandconverter from OpAMP config provider to fix $1 pattern crash
--- a/otelcolbuilder/.gitignore
+++ b/otelcolbuilder/.gitignore
@@ -1,5 +1,6 @@
 cmd/*
 !cmd/collector_config_test.go
+!cmd/configprovider_test.go
 !cmd/collector_windows.go
 !cmd/testdata/
 !cmd/configprovider.go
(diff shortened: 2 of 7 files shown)


jagan2221:
Looks good, pls add changelog and check failing tests. @raunakmandaokar

Astitva-sumo (approved):

PR #2110, opened 2026-06-16
Fix: race condition in opamp startup

jagan2221:
Why? 
1. https://sumologic.atlassian.net/browse/SUMO-284871?focusedCommentId=890663
Opamp sends restart command to service before service even starts which results in desired behaviour.

2. https://sumologic.atlassian.net/browse/SUMO-286636?focusedCommentId=891776
Opamp accepts configs while service is shutting down causing config update logic to delete local configs and not recreate them.

Ideally, opamp should start accepting messages once the pipeline is ready and stop accepting messages when collector is shutting down. This fix addresses both cases.

Added more UT's to ensure the opamp's lifecycle channels are working as expected on startup and Reload cases.

**Local testing:**
1. Installed new otel collector binary with this change.
4. Ran collector restart command multiple times and ensured startup is fine for all attempts.
5. Pushed remote config and verified opamp service restart works as expected.


pkg/extension/opampextension/opamp_agent.go
```diff
@@ -646,6 +674,18 @@ func (o *opampAgent) onMessage(ctx context.Context, msg *types.MessageData) {
 			o.logger.Error("Failed to update OpAMP agent effective configuration", zap.Error(err))
 		}
 
+		// Wait until the collector's service is fully ready before sending SIGHUP.
+		// If we send SIGHUP before the collector registers its signal handler
+		// (which happens after setupConfigurationComponents completes), the default
+		// OS disposition terminates the process.
+		select {
```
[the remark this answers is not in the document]
jagan2221: Covered in E2E

pkg/extension/opampextension/opamp_agent.go
```diff
@@ -208,13 +217,31 @@ func (o *opampAgent) Shutdown(ctx context.Context) error {
...
 
 	if err != nil {
 		return err
 	}
 
+	// Reinitialize readyCh and readyOnce for the new lifecycle after reload
+	o.readyCh = make(chan struct{})
+	o.readyOnce = sync.Once{}
+
 	return o.Start(ctx, o.host)
```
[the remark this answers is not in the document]
jagan2221: Done

pkg/extension/opampextension/opamp_agent.go
```diff
@@ -646,6 +677,18 @@ func (o *opampAgent) onMessage(ctx context.Context, msg *types.MessageData) {
 			o.logger.Error("Failed to update OpAMP agent effective configuration", zap.Error(err))
 		}
 
+		// Wait until the collector's service is fully ready before sending SIGHUP.
+		// If we send SIGHUP before the collector registers its signal handler
+		// (which happens after setupConfigurationComponents completes), the default
+		// OS disposition terminates the process.
+		select {
```
ravishankar-sumo: nit: It is better to have this before updating the effective config for readability sequence.

pkg/extension/opampextension/opamp_agent.go
```diff
@@ -208,13 +217,34 @@ func (o *opampAgent) Shutdown(ctx context.Context) error {
...
 func (o *opampAgent) Reload(ctx context.Context) error {
 	err := o.Shutdown(ctx)
 
 	if err != nil {
 		return err
 	}
 
+	// Reinitialize lifecycle-scoped fields for the new lifecycle after reload.
+	o.readyCh = make(chan struct{})
+	o.readyOnce = sync.Once{}
```
ravishankar-sumo: I am not entirely sure if we need this readyOnce. If the readyCh is closed the onMessage would proceed and all subsequent calls would also consider readyCh as closed.

pkg/extension/opampextension/opamp_agent.go
jagan2221: We are ready with the new effective config, but just ensuring service is started before issuing a restart command to the service.

pkg/extension/opampextension/opamp_agent.go
jagan2221: This is nothing but a guard against Ready being called twice and we end up trying to close readyCh twice which results in panic. As per current service flow, it calls Ready only once, but this is just a defensive guard to be cautious. What do you think? @ravishankar-sumo Also now i've removed resetting readyCh and readyOnce since resetting them is not required on reload.

pkg/extension/opampextension/opamp_agent.go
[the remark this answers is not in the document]
jagan2221: Added Unit tests

pkg/extension/opampextension/opamp_agent.go
```diff
@@ -646,6 +673,18 @@ func (o *opampAgent) onMessage(ctx context.Context, msg *types.MessageData) {
...
+		// (which happens after setupConfigurationComponents completes), the default
+		// OS disposition terminates the process.
+		select {
+		case <-o.readyCh:
+			// Service is ready, safe to reload
+		case <-o.lifetimeCtx.Done():
+			o.logger.Warn("OpAMP agent shutting down, skipping config reload")
+			return
+		}
+
```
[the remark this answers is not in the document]
ravishankar-sumo: This seems to be a valid case. Any reason for resolving @jagan2221 ?

After the first run, `readyCh` will be closed, so on subsequent calls the `readyCh` case will always be selected immediately. As a result, the `select` might skip the check of `lifetimeCtx`. Though the case statements are not selected in sequence there could be a race condition.

Having this check after select would resolve this race condition.
```
if o.lifetimeCtx.Err() != nil {
o.logger.Warn("OpAMP agent shutting down, skipping config reload")
return
}
```

pkg/extension/opampextension/opamp_agent.go
ravishankar-sumo: I checked the implementation of the pipeline notifier. As you mentioned there is no race condition that the Ready could be called twice. The pipeline notifier is called after services startAll call. IMO We can remove this defence for now. The upstream opamp agent also has a similar readyCh implementation and they rely directly on the Ready call which closes the channel [Ref](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/opampextension/opamp_agent.go#L246)

pkg/extension/opampextension/opamp_agent.go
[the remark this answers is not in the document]
jagan2221: Yes, this is a valid case, resolved since it's an ACK and locally i've adopted this change a little differently and investigating more related to another race condition we are seeing [here](https://sumologic.atlassian.net/browse/SUMO-286636) before pushing it.

pkg/extension/opampextension/opamp_agent.go
jagan2221: Moved the logic to beginning of onMessage

pkg/extension/opampextension/opamp_agent.go
```diff
@@ -208,13 +216,30 @@ func (o *opampAgent) Shutdown(ctx context.Context) error {
...
+func (o *opampAgent) Ready() error {
+	close(o.readyCh)
+	return nil
+}
+
+// NotReady implements extensioncapabilities.PipelineWatcher.
+// It is called when the collector is about to stop receiving data.
+func (o *opampAgent) NotReady() error {
+	return nil
+}
```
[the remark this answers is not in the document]
jagan2221: It is fine to rely on extension's Shutdown without over complicating by introducing another channel to the lifecycle. Component's Shutdown call and NotReady in real time will occur very closely so existing component shutdown check model would be sufficient.

pkg/extension/opampextension/opamp_agent.go
```diff
@@ -208,13 +216,30 @@ func (o *opampAgent) Shutdown(ctx context.Context) error {
 	return nil
 }
 
+// Ready implements extensioncapabilities.PipelineWatcher.
+// It is called when all pipelines are started and the collector is ready to receive data.
+// This unblocks any pending SIGHUP signals for config reload.
+func (o *opampAgent) Ready() error {
+	close(o.readyCh)
+	return nil
+}
```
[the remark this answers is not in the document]
jagan2221: makes sense.

pkg/extension/opampextension/opamp_agent.go
```diff
@@ -604,6 +633,24 @@ func (o *opampAgent) applyRemoteConfig(config *protobufs.AgentRemoteConfig) (con
 }
 
 func (o *opampAgent) onMessage(ctx context.Context, msg *types.MessageData) {
+	// drop messages once shutdown has begun
+	select {
```
ravishankar-sumo: nit: We could have the if condition check seemed simpler and also have this below the select of readyCh

ravishankar-sumo (approved):

pkg/extension/opampextension/opamp_agent.go
jagan2221: We can't have this in below select of readyCh because readyCh will always be selected irrespective of lifetimeCtx once pipelines are started, So we will miss it even when it's true. Modified with if check

